Issue #16, opened 2019-07-01
Automatically lock old/stale issues and PRs


josephperrott:
- [x] Framework
- [x] CLI
- [x] Components

PR #28, opened 2019-08-27
Initial lock issue Github action

clydin:


github-actions/lock-closed/src/main.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+async function run(): Promise<void> {
+  try {
+    // NOTE: `days` and `message` must not be changed without dev-rel and dev-infra concurrence
+
+    // Fixed amount of days a closed issue must be inactive before being locked
+    const days = 30;
+    // Standardized Angular Team message for locking issues
+    const message =
```
josephperrott: Would it be better to use [`commonTags.stripIndent`](https://github.com/declandewet/common-tags#stripindent) and move this to a template string?

This would also allow us to include the link as a variable a little easier.

```ts
// NOTE: `policyUrl`, `days` and `message` must not be changed without dev-rel and dev-infra concurrence

const policyUrl = `https://github.com/angular/angular/{sha}/docs/GITHUB_PROCESS.md#automatic-conversation-locking`;
// Fixed amount of days a closed issue must be inactive before being locked
const days = 30;
// Standardized Angular Team message for locking issues
const message = stripIndent`
  This issue has been automatically locked due to inactivity.
  Please file a new issue if you are encountering a similar or related problem.

  Read more about our [automatic conversation locking policy](${policyUrl})

  _This action has been performed automatically by a bot._`;
```

github-actions/lock-closed/src/main.ts
```diff
@@ -0,0 +1,78 @@
...
+      'This issue has been automatically locked due to inactivity.\n' +
+      'Please file a new issue if you are encountering a similar or related problem.\n' +
+      '\n' +
+      'Automatically locking inactive issues helps keep GitHub notifications relevant for project maintainers.\n' +
```
josephperrott: Rather than this line, I think we should link to the policy doc being created in https://github.com/angular/angular/pull/32405

```
Read more about our [automatic conversation locking policy](https://github.com/angular/angular/{sha}/docs/GITHUB_PROCESS.md#automatic-conversation-locking)
```

clydin: Done.

github-actions/lock-closed/src/main.ts
clydin: Added the policyUrl line as a template string.

I usually don't add in `common-tags` unless its needed several times in the codebase.  If you prefer its use, I can add it in though.

josephperrott: That's fine with me

josephperrott:
This PR is the first for #16

devversion:
The PR to add locking got locked 💯

PR #29, opened 2019-09-03
feat(github-actions): allow configuration of max issue locks per execution

clydin:


github-actions/lock-closed/action.yml
```diff
@@ -5,6 +5,9 @@ inputs:
   github-token:
     description: 'Github access token'
     required: true
+  max-per-execution:
```
josephperrott: Can we change this to `locks-per-execution` and then enforce a max in the code?

github-actions/lock-closed/src/main.ts
```diff
@@ -34,6 +34,7 @@ async function run(): Promise<void> {
       `Read more about our [automatic conversation locking policy](${policyUrl}).\n\n` +
       '_This action has been performed automatically by a bot._';
 
+    const maxPerExecution = +core.getInput('max-per-execution') || 400;
```
josephperrott: If we do this as `locks-per-execution`:

```ts
const maxPerExecution = Math.min(+core.getInput('locks-per-execution'), 400);
```

clydin: Done.

github-actions/lock-closed/action.yml
clydin: Done.

PR #30, opened 2019-09-03
ci: initial closed issue locking workflow

clydin:


.github/workflows/lock-closed.yml
```diff
@@ -0,0 +1,14 @@
+name: Lock issues that are closed and inactive
+
+on:
+  schedule:
+  - cron: 25 * * * *
```
josephperrott: Should the cron parameter be quoted? `'25 * * * *'`

Add comment explaining that this will run on the 25th min of each hour?

PR #34, opened 2019-09-03
fix(github-actions): add Github request logging on error

clydin:


clydin:
Copied that in manually: https://github.com/angular/dev-infra/blob/d2299fe5df76b14243012e9c89a4b787b5ec13ce/github-actions/lock-closed/lib/bundle.js#L21905

josephperrott:
For some reason in my mind, me not seeing it in the diff meant that it wasn't there.  But in fact me not seeing it means it is there.

PR #41, opened 2019-09-03
setup precommit hook to rebuild lock-closed bundle and run prettier

josephperrott:


package.json
```diff
@@ -0,0 +1,12 @@
...
+    "husky": "^3.0.5",
+    "prettier": "^1.18.2",
+    "pretty-quick": "^1.11.1"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "./scripts/pre-commit.sh"
+    }
+  }
+}
```
gkalpak: You might want to add `private: true` to indicate that this is not intended to be published on npm.

PR #42, opened 2019-09-03
use @octokit/rest directly

josephperrott:
This is necessary as a workaround for https://github.com/actions/toolkit/issues/106

### DIFF
--- a/github-actions/lock-closed/lib/bundle.js
+++ b/github-actions/lock-closed/lib/bundle.js
@@ -21891,6 +21891,11 @@ const CORE_PLUGINS = [
 
 var rest = core$2.plugin(CORE_PLUGINS);
 
+var GitHub = /*#__PURE__*/Object.freeze({
+	'default': rest,
+	__moduleExports: rest
+});
+
 var context = createCommonjsModule(function (module, exports) {
 Object.defineProperty(exports, "__esModule", { value: true });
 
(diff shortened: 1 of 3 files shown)


github-actions/lock-closed/lib/bundle.js
```diff
@@ -25247,15 +25252,15 @@ async function run() {
         // Fixed amount of days a closed issue must be inactive before being locked
         const days = 30;
         // Standardized Angular Team message for locking issues
-        const policyUrl = 'https://github.com/angular/angular/blob/67d80f9ae8082d446e2d58227375f5a92eeae933/docs/GITHUB_PROCESS.md#conversation-locking';
+        const policyUrl = 'https://com/angular/angular/blob/67d80f9ae8082d446e2d58227375f5a92eeae933/docs/GITHUB_PROCESS.md#conversation-locking';
```
gkalpak: Was this intentional?

github-actions/lock-closed/src/main.ts
```diff
@@ -48,7 +49,8 @@ async function run(): Promise<void> {
     // Fixed amount of days a closed issue must be inactive before being locked
     const days = 30;
     // Standardized Angular Team message for locking issues
-    const policyUrl = 'https://github.com/angular/angular/blob/67d80f9ae8082d446e2d58227375f5a92eeae933/docs/GITHUB_PROCESS.md#conversation-locking';
+    const policyUrl =
+      'https://com/angular/angular/blob/67d80f9ae8082d446e2d58227375f5a92eeae933/docs/GITHUB_PROCESS.md#conversation-locking';
```
gkalpak: Was this intentional?

PR #47, opened 2019-09-05
set max locks per run to 100 and adds timeout to prevent errant abuse detection 

josephperrott:


github-actions/lock-closed/src/main.ts
```diff
@@ -22,6 +22,11 @@ async function lockIssue(client: Octokit, issue: number, message?: string): Prom
   });
 }
 
+/** Creates a promise which resolves after a set period of time. */
+async function timeout(ms: number) {
```
gkalpak: Nit: No need for `async` :grin:

PR #53, opened 2019-09-11
Skip already locked items, add additional commenting

josephperrott:


github-actions/lock-closed/src/main.ts
```diff
@@ -104,24 +101,30 @@ async function run(): Promise<void> {
...
       let itemType: string | undefined;
       try {
-        itemType = issue.pull_request ? 'pull request' : 'issue';
-        console.info(`Locking ${itemType} #${issue.number}`);
-        await lockIssue(client, issue.number, message);
+        if (item.locked) {
+          console.info(`Skipping ${itemType} #${item.number}, already locked`);
+          continue;
+        }
+        itemType = item.pull_request ? 'pull request' : 'issue';
```
clydin: This needs to be before the if block.

Issue #56, opened 2019-10-02
Github Action to automatically rebase PR if possible


josephperrott:

josephperrott:
Closing as this has been determined to be infeasible as it would overwhelm CI

PR #58, opened 2019-10-07
create rebase-pr action

josephperrott:


josephperrott:
For #56

.github/workflows/rebase-pr.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  rebase_pr:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout repo
+        uses: actions/checkout@master
+      - name: Rebase PR
+        if: |
+          github.event_name == 'pull_request' &&
+          github.event.action == 'labeled' &&
```
clydin: Shouldn't need these first two since it will only be triggered on a labeled pull request from above.

.github/workflows/rebase-pr.yml
```diff
@@ -0,0 +1,20 @@
...
+          github.event.label.name == 'rebase-pr'
```
devversion: Just out of curiosity since I'm not too familiar with Github actions (yet :grin:). Do we need to check `event_name` and `event.action` again? (_aren't these covered by the `on:` field?

github-actions/rebase-pr/entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# The repository name and branch name for the PRs head and base
+HEAD_REPOSITORY=$(jq -r ".pull_request.head.repo.full_name" "$GITHUB_EVENT_PATH")
+HEAD_BRANCH=$(jq -r ".pull_request.head.ref" "$GITHUB_EVENT_PATH")
+BASE_REPOSITORY=$(jq -r ".pull_request.base.repo.full_name" "$GITHUB_EVENT_PATH")
+BASE_BRANCH=$(jq -r ".pull_request.base.ref" "$GITHUB_EVENT_PATH")
+
+# Change to the github workspace and initialize the repo if needed.
+cd "$GITHUB_WORKSPACE"
+git init
```
devversion: In what cases would the folder not be a Git repo?

github-actions/rebase-pr/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM alpine:3.10
+
+RUN apk --no-cache add jq bash git
+
+ADD entrypoint.sh /entrypoint.sh
```
clydin: nit: COPY is safer than ADD (COPY only supports files; ADD can access remote files via a URL)

github-actions/rebase-pr/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM alpine:3.10
+
+RUN apk --no-cache add jq bash git
```
clydin: It'd be nice if there was eventually a minimal base that had common utilities.

github-actions/rebase-pr/entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# If the rebase attempt fails, exit with a failing (1) exit code.
+if [ $? -ne 0 ]; then
+  # TODO(josephperrott): Push a comment to the PR asking the author to rebase.
+  exit 1;
+fi
+
+# Push the rebased code back to the remote, using force as this is required
+# for rebases, then exit with a success exit code.
+git push -f head "$HEAD_BRANCH";
```
devversion: Would it be worth using `force-with-lease`? It could avoid that the action overwrites new commits pushed to the PR while this executes I guess.

github-actions/rebase-pr/entrypoint.sh
josephperrott: The git repo is only already available because in instance we have the `actions/checkout` action called before our `angular/dev-infra/github-actions/rebase-pr` action.

If we were to not have a step using `actions/checkout` beforehand, the repo would not be initialized yet.

devversion: I see. would the `cd ""` then be incorrect if `GITHUB_WORKSPACE` is empty? Should we check for the `GITHUB_WORKSPACE` and otherwise create a tmp folder?

clydin: It's provided by the system: https://help.github.com/en/articles/virtual-environments-for-github-actions#environment-variables

josephperrott: The `GITHUB_WORKSPACE` directory is always created, but the `actions/checkout` action checks out the repo and places it there.

So if `GITHUB_WORKSPACE` is `/github/workspace/` then the repo would be there `/github/workspace/.git` being the location of the project git config (since that expresses where the repo is extracted at)

.github/workflows/rebase-pr.yml
```diff
@@ -0,0 +1,17 @@
...
+on:
+  pull_request:
+    types: [labeled]
+
+jobs:
+  rebase_pr:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout repo
+        uses: actions/checkout@master
```
gkalpak: Does this mean that it rebases PRs on master? Could we make ut so that it rebases against the base branch (ir is it too complicated)? If not, tgen this shoyld be part of the name and/or the label to make it clear to users.

github-actions/rebase-pr/entrypoint.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+set -e
+
+# Set email and name config
+git config --global user.email "devops+rebase@angular.io"
+git config --global user.name "Angular Rebase Bot"
+
+# Exit for code a successful run
```
gkalpak: `a` --> `on` (?)

github-actions/rebase-pr/entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+BASE_REPOSITORY=$(jq -r ".pull_request.base.repo.full_name" "$GITHUB_EVENT_PATH")
+BASE_BRANCH=$(jq -r ".pull_request.base.ref" "$GITHUB_EVENT_PATH")
+
+# Change to the github workspace and initialize the repo if needed.
+cd "$GITHUB_WORKSPACE"
+git init
+
+# Add head and base repos as remotes, then fetch the remote branches
+git remote add head https://x-access-token:$TOKEN@github.com/$HEAD_REPOSITORY
+git remote add base https://x-access-token:$TOKEN@github.com/$BASE_REPOSITORY
```
gkalpak: Is this the recommended way to use the token? Is there any chance this could result in the token's showing up in some logs or something?

github-actions/rebase-pr/entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+HEAD_BRANCH=$(jq -r ".pull_request.head.ref" "$GITHUB_EVENT_PATH")
+BASE_REPOSITORY=$(jq -r ".pull_request.base.repo.full_name" "$GITHUB_EVENT_PATH")
+BASE_BRANCH=$(jq -r ".pull_request.base.ref" "$GITHUB_EVENT_PATH")
+
+# Change to the github workspace and initialize the repo if needed.
+cd "$GITHUB_WORKSPACE"
+git init
+
+# Add head and base repos as remotes, then fetch the remote branches
+git remote add head https://x-access-token:$TOKEN@github.com/$HEAD_REPOSITORY
```
gkalpak: `HEAD` already has a meaning in git (and in case-insensitive systems, such as Windows, it is the same as `head`). Consider using a different name for the remote.

github-actions/rebase-pr/entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+git fetch base "$BASE_BRANCH";
+
+# Checkout the PR change branch from its remote
+git checkout --track head/"$HEAD_BRANCH"
+
+# Perform rebase of PR
+git rebase base/"$BASE_BRANCH";
+
+# If the rebase attempt fails, exit with a failing (1) exit code.
+if [ $? -ne 0 ]; then
```
gkalpak: Isn't this redundant due to the `set -e` (i.e. we can never get here)? I would just leave the TODO comment and remove the `if` block.

github-actions/rebase-pr/entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+  # TODO(josephperrott): Push a comment to the PR asking the author to rebase.
+  exit 1;
+fi
+
+# Push the rebased code back to the remote, using force as this is required
+# for rebases, then exit with a success exit code.
+git push --force-with-lease head "$HEAD_BRANCH";
+exit $SUCCESS_EXIT_CODE;
```
gkalpak: This feels refundant as well :thinking_face:

github-actions/rebase-pr/entrypoint.sh
josephperrott: Github prevents the printing of the token in the logs, so it shouldn't be a problem.

I agree that I don't love it as the method for providing the token, I was unfortunately not able to find another way that worked in this context to provide the token to git.

github-actions/rebase-pr/entrypoint.sh
josephperrott: The reason I have it there is because while `0` is the current success code.  I was unsure if this will always be the case for Github Actions as they have mentioned in the Actions issues repo that they may have other result codes as well.

.github/workflows/rebase-pr.yml
gkalpak: Yeah, sorry, I figured it out at some point during the review, but forgot to remove this comment 😇

github-actions/rebase-pr/entrypoint.sh
gkalpak: You can probably write something to a file and set that file as the credential helper store. Something similar to what we do on [CircleCI][1] (without the decryption step). Plus you might also need some `git config credential.helper "store --file=/path/to/credentials/file"`.

But that is not bullet-proof either. Not worth the extra work if GitHub prevents printing the token in the logs.

[1]: https://github.com/angular/angular/blob/64823f561c2c25f05e9bce04de68374ce997548a/.circleci/config.yml#L580

github-actions/rebase-pr/entrypoint.sh
```diff
@@ -0,0 +1,39 @@
...
+git init;
+
+# Add pr and target repos as remotes, then fetch the remote branches
+git remote add pr https://x-access-token:$TOKEN@github.com/$PR_REPOSITORY;
+git remote add target https://x-access-token:$TOKEN@github.com/$TARGET_REPOSITORY;
+git fetch pr "$PR_BRANCH";
+git fetch target "$TARGET_BRANCH";
+
+# Checkout the PR change branch from its remote
+git checkout --track pr/"$PR_BRANCH"
```
devversion: Optional nit: Ideally, there would be a comment explaining why we use `--track` here, and below why we use `--force-with-lease` (just for later reference), but that can be a follow-up :smile:

PR #64, opened 2019-10-11
check-cherry-pick: create an action to check if a PR can be cherry picked

josephperrott:


.github/workflows/check-cherry-pick.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  rebase-pr:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v1
+      - name: Check Cherry Pick
+        if: >
+          contains(github.pull_request.labels.*.name, 'PR target: master & patch') ||
```
devversion: What about minor labels?

.github/workflows/check-cherry-pick.yml
```diff
@@ -0,0 +1,19 @@
...
+          contains(github.pull_request.labels.*.name, 'PR target: patch')
+        uses: ./github-actions/check-cherry-pick
+        env:
+          TARGET: master
```
devversion: Should this be named `PATCH_BRANCH_NAME` or something?

github-actions/check-cherry-pick/entrypoint.sh
```diff
@@ -0,0 +1,33 @@
...
+PR_NUMBER=$(jq -r ".pull_request.number" "$GITHUB_EVENT_PATH")
+# The patch branch for the repository, using the provided TARGET environment variable if defined
+if [[ -z $TARGET ]]; then
+  TARGET_BRANCH=$(git ls-remote --heads https://github.com/$REPO_NAME.git | grep -E 'refs\/heads\/[0-9]+\.[0-9]+\.x' | cut -d '/' -f3 | sort -r | head -n1)
+else
+  TARGET_BRANCH=$TARGET
+fi
+
+# Retrieve the patch file for the pr
+curl -L -s https://github.com/$REPO_NAME/pull/$PR_NUMBER.diff > /tmp/pull.diff;
```
devversion: Should we add `--show-error` here? we still want to print error messages I guess?

github-actions/check-cherry-pick/entrypoint.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Add target repo as remote, then fetch the target branch from the remote
+git remote add target https://github.com/$REPO_NAME
+git fetch target $TARGET_BRANCH
+
+# Checkout the target branch from its remote
+git checkout --track target/$TARGET_BRANCH
```
devversion: Do we need `--track` here? I thought it's only needed if we push back to the remote.

.github/workflows/check-cherry-pick.yml
gkalpak: Can't we just take the branch from the GitHub event payload (at `$GITHUB_EVENT_PATH`) from `.pull_request.base.ref`?

.github/workflows/check-cherry-pick.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  rebase-pr:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v1
+      - name: Check Cherry Pick
+        if: >
+          contains(github.pull_request.labels.*.name, 'PR target: master & patch') ||
+          contains(github.pull_request.labels.*.name, 'PR target: patch')
```
gkalpak: I don't know about other repos, but there is no such label on angular/angular :grin: (For `PR target:` we have `master & patch`, `master-only`, `patch-only`, `LTS-only` and `TBD`.) 

Also, why limit this to PRs with specific `PR target` labels? Is it to save resources?

clydin: If it wasn't designed to be cherry-picked onto a different branch, there is a very high chance it will fail.  At least this is the case for the CLI.  With that in mind, testing something that you know will fail would both waste resources and give a false impression as to the viability of the PR (someone would then need to explain that failing here doesn't actually matter for this case).

Separately, there's also the consideration that there can be multiple LTS versions active at any point in time.  Currently both 6.x and 7.x are LTS (https://angular.io/guide/releases#support-policy-and-schedule).

gkalpak: Ooops, I somehow thought this was about rebasing on the target branch 😇
For checking "cherry-pick-ability" having `PR target: master & patch` and `PR target: patch-only` makes sense. (Notice that it is `patch-only`, not `patch`, though.)

.github/workflows/check-cherry-pick.yml
gkalpak: Nevermind, for some reason I though this was about rebasing on the target branch 🤷‍♂

.github/workflows/check-cherry-pick.yml
josephperrott: So this yaml file is actually just the definition for the workflow in this repo, not in the repos where we actually use the action such as angular/angular, angular/components, angular/angular-cli, etc.

Its only really here as an illustration.

.github/workflows/check-cherry-pick.yml
devversion: I see. I was confused because it lives in the `.github/` folder.

PR #83, opened 2021-02-09
Implement a GitHub bot for handling feature requests

mgechev:
The GitHub bot in this PR introduces a voting process for feature requests inspired by the way VSCode [manage their repository](https://www.youtube.com/watch?v=-Olo7N9xwV8).

**tl;dr**
The bot:
- Initiates voting process for each incoming feature request
- Shares a reminder when the end of the voting process approaches
- Closes the request if the issue has not collected enough 👍 
- Labels the issue for consideration if it has the necessary number of 👍


github-actions/feature-request/src/octokit.ts
```diff
@@ -0,0 +1,260 @@
...
+    log(`Deleting comment ${id} on ${this.issueData.number}`)
+    if (!this.options.readonly)
+      await this.octokit.issues.deleteComment({
+        owner: this.params.owner,
+        repo: this.params.repo,
+        comment_id: id,
+      })
+  }
+
+  async *getComments(last?: boolean): AsyncIterableIterator<Comment[]> {
```
josephperrott, line 191: Can we just yield the individual comments?  I suppose this is a comment on the general pattern as well now that I am looking in more detail.

It seems to me that since often when we request things, because have to iterate through them to make the localized version (i.e. `Octokit.IssuesListCommentsResponseItem` is converted into `Comment`), that we should just yield out the Comment directly. I feel like this is much easier to reason about in its usages because then. `getComents()` moves from returning what amounts to `Comment[][]` to `Comment[]`.

I am not sure if I am missing something that makes the current method more valuable.

josephperrott (changes requested):

github-actions/feature-request/src/action.ts
```diff
@@ -0,0 +1,206 @@
...
+  // until we post a warning and close it in `closeAfterWarnDaysDuration`.
+  oldIssueWarnDaysDuration: number;
+  // This flag indicates for how long after the initial comment
+  // that kicks off the voting process we should post a warning comment.
+  warnDaysDuration: number;
+  // Indicates how may days after the warning comment we should close
+  // the issue and post an explanation.
+  closeAfterWarnDaysDuration: number;
+
+  closeComment: string;
```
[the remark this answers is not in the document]
mgechev: I was thinking of keeping the implementation as generic as possible to make it easier to share with other teams at Google.

github-actions/feature-request/src/octokit.ts
```diff
@@ -0,0 +1,260 @@
...
+      for (const user of page.data as Octokit.OrgsListMembersResponse) {
+        this._orgMembers.add(user.login);
+        // Don't return early to make sure we
+        // populate the user set.
+        if (user.login === name) {
+          found = true;
+        }
+      }
+    }
+    return found;
```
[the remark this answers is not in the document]
mgechev: Kept the early return but removed the `found` flag. Let me know what you thikn.

mgechev:
@josephperrott thanks for the review! I addressed all your comments. Let me know what you think.

github-actions/feature-request/src/action.ts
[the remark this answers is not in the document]
josephperrott: Can we set defaults? I just would love to be able to manage the message for our team in a single location.

github-actions/feature-request/src/action.ts
```diff
@@ -0,0 +1,210 @@
+import { GitHubAPI, GitHubIssueAPI, MarkedComment, Issue } from './api';
+import { log } from './log';
+
+export enum CommentMarkers {
```
AndrewKushnir, line 4: I'd propose to add a quick comment on how these markers are used (we add them into comment body and classify comments later based on these comments).

github-actions/feature-request/src/action.ts
```diff
@@ -0,0 +1,210 @@
...
+ * @param githubAPI Object used for querying the GitHub API
+ * @param githubIssue Object used for communication with the GitHub API for a particular issue
+ * @param config Configuration of the GitHub bot
+ */
+const processIssue = async (githubAPI: GitHubAPI, githubIssue: GitHubIssueAPI, config: Config) => {
+  const issue = await githubIssue.get();
+
+  // Issues opened by team members bypass the process.
+  if (await githubAPI.isOrgMember(issue.author.name, config.organization)) {
+    log('The creator of this issue is a member of the organization.');
```
AndrewKushnir, line 72:
```suggestion
    log('The creator of issue #${issue.number} is a member of the organization.');
```

I was also thinking if we could change the `log` function to accept an issue #, action and the message, i.e.:
```
log(issue.number, 'skip', 'Issue author is a member of the organization');
```

So the output generated by the `log` fn might be something like:

```
#35671 [skip]: Issue author is a member of the organization
#41223 [comment]: Voting started comment posted
#12321 [close]: Issue is closed without sufficient votes count after 60 days
```

This is very minor though, feel free to ignore :)

github-actions/feature-request/src/action.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  if (timestamps.warn !== null && daysSince(timestamps.warn) >= config.closeAfterWarnDaysDuration) {
+    // In the future consider closing associated PRs if we have high
+    // level of confidence they are scoped to the feature request.
+    log(`Closing feature request #${issue.number}`);
+    return await Promise.all([
+      githubIssue.postComment(comment(CommentMarkers.Close, config.closeComment)),
+      config.closeWhenNoSufficientVotes
+        ? githubIssue.close()
+        : githubIssue.addLabel(config.insufficientVotesLabel),
```
AndrewKushnir, line 132: Just as idea: may be we can add labels in all cases, even if we are about to close the issue? That might help us with introspection later: we should be able to filter all issues/PRs closed with that particular label and:
- do post-process verification of selected issues
- provide various stats, such as avg number of issues close per week, etc

github-actions/feature-request/src/octokit.ts
```diff
@@ -0,0 +1,259 @@
...
+  constructor(
+    protected token: string,
+    protected params: { repo: string; owner: string },
+    protected options: { readonly: boolean } = { readonly: false },
+  ) {
+    this._octokit = new GitHub(token);
+  }
+
+  async *query(query: Query): AsyncIterableIterator<GitHubIssueAPI> {
+    const q = query.q + ` repo:${this.params.owner}/${this.params.repo}`;
```
AndrewKushnir, line 37:
```suggestion
    const q = `${query.q} repo:${this.params.owner}/${this.params.repo}`;
```

github-actions/feature-request/src/octokit.ts
```diff
@@ -0,0 +1,259 @@
...
+      number: issue.number,
+      title: issue.title,
+      labels: (issue.labels as Octokit.IssuesGetLabelResponse[]).map((label) => label.name),
+      open: issue.state === 'open',
+      locked: (issue as any).locked,
+      numComments: issue.comments,
+      reactions: (issue as any).reactions,
+      assignee: issue.assignee?.login ?? (issue as any).assignees?.[0]?.login,
+      createdAt: +new Date(issue.created_at),
+      updatedAt: +new Date(issue.updated_at),
```
AndrewKushnir, line 111: Is it possible that the issue was never updated (should we have a check similar to `closed_at` one below)?

AndrewKushnir (commented):
Looks great, thanks @mgechev 👍

github-actions/feature-request/src/octokit.ts
mgechev: That's a good point! Based on octokit's typings, `updated_at` is always a string, so we shouldn't ever get a `null` value.

github-actions/feature-request/src/action.ts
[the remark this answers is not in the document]
mgechev: Done

github-actions/feature-request/src/octokit.ts
```diff
@@ -0,0 +1,259 @@
...
+      log(`Page ${++pageNum}: ${page.map(({ number }) => number).join(' ')}`);
+      for (const issue of page) {
+        yield new OctoKitIssue(
+          this.token,
+          this.params,
+          this.octokitIssueToIssue(issue),
+          this.options,
+        );
+      }
+    }
```
josephperrott, line 72: What do you think of doing this instead and eliminating the iterator entirely?

This just has github paginate and then merge the pages together, allowing you to perform the transformation just before merging into a single list.  The return type of this is `Promise<OctoKitIssue[]>`, which I think ends up being easier to work with.  Additionally, it removes the need for the backoff timeouts as Github handles it for us.


```ts
    return await this.octokit.paginate(
      options,
      (page: Octokit.Response<Octokit.SearchIssuesAndPullRequestsResponseItemsItem[]>) => {
        return page.data.map(issue => new OctoKitIssue(
          this.token,
          this.params,
          this.octokitIssueToIssue(issue),
          this.options,
          ));
        });
      }
```

github-actions/feature-request/action.yml
```diff
@@ -0,0 +1,71 @@
...
+    default: 20
+  feature-request-label:
+    description: 'Label used to distinguish feature request from other issues.'
+    default: 'feature'
+  in-backlog-label:
+    description: 'Label used to distinguish feature requests, which are already part of the backlog.'
+    default: 'in-backlog'
+  requires-votes-label:
+    description: 'Label used to distinguish requests in voting phase from other issues.'
+    default: 'votes-required'
```
AndrewKushnir, line 22: We don't have a consistent way of naming labels in FW repo, but labels created recently don't have hyphens between words. I'd propose removing hyphens from new labels as well :) (so this label would be `votes required`).

AndrewKushnir (commented):
As we discussed with @mgechev, I'm adding a couple comments related to labels. Thank you.

github-actions/feature-request/action.yml
```diff
@@ -0,0 +1,71 @@
...
+    default: 10
+  minimum-votes-for-consideration:
+    description: 'Number of votes required to move a feature request to our consideration list.'
+    default: 20
+  feature-request-label:
+    description: 'Label used to distinguish feature request from other issues.'
+    default: 'feature'
+  in-backlog-label:
+    description: 'Label used to distinguish feature requests, which are already part of the backlog.'
+    default: 'in-backlog'
```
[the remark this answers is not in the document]
mgechev: >We can probably rename this config option to in-backlog-filter and specify the entire filter segment (e.g. milestone:Backlog) as a value.

I like this suggestion! I noticed that currently we have all the issues in the backlog. Is this expected?

github-actions/feature-request/src/octokit.ts
mgechev: I see we've already used this in the issue lock bot, so seems like `paginate` will simplify the implementation!

github-actions/feature-request/action.yml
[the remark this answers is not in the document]
AndrewKushnir: Yeah, that's expected (see 2nd part of my comment above). The idea of the current `Backlog` milestone is to say that these tickets are triaged (so no immediate processing is needed), all tickets that are not in that milestone should be looked at and either closed or added to the `Backlog` (by adding extra labels to an issue). With feature requests, the only label that is needed to consider it triaged is "feature" (i.e. once that label is applied, the ticket is added to the `Backlog` milestone by NgBot). 

I believe we should have a new label to tell the bot that the ticket was already processed or should be ignored.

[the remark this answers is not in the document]
mgechev: I think adding extra label will simplify things and will not require any changes to the current process (i.e., `Backlog` will still be only for triaged issues).

What about if we rename the `in backlog` label to `roadmap`? I'm also leaning towards continuing to use labels in the both rather than milestones for consistency and simplicity.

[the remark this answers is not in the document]
AndrewKushnir: > What about if we rename the in backlog label to roadmap?

I think that might send wrong message to the issue author :) Having the `roadmap` label might be perceived as "hey, your feature request is now a part of our roadmap (i.e. it's accepted and will be implemented)". I was thinking may be we can call it `status: triaged`?

github-actions/feature-request/src/octokit.ts
josephperrott: I haven't seen us running into out of memory issues, though I didn't think about that being an issue.  I am fine with either way you want to go about it.

github-actions/feature-request/src/action.ts
```diff
@@ -0,0 +1,218 @@
...
+ * Processes a particular issue from the repository. At the beginning of the invocation
+ * the function makes sure the issue is in the target group of feature requests we want to process.
+ * Depending on the current state of the issue, the function will either label it, close it,
+ * add a comment, or perform a noop.
+ *
+ * @param githubAPI Object used for querying the GitHub API
+ * @param githubIssue Object used for communication with the GitHub API for a particular issue
+ * @param config Configuration of the GitHub bot
+ */
+const processIssue = async (githubAPI: GitHubAPI, githubIssue: GitHubIssueAPI, config: Config) => {
```
josephperrott, line 73: Is there a reason to have this as a `const` rather than a declared local function?

github-actions/feature-request/src/main.ts
```diff
@@ -0,0 +1,48 @@
+import * as core from '@actions/core';
+import { context } from '@actions/github';
+import { run } from './action';
+import { OctoKit } from './octokit';
+
+// Gets a specific value from the YAML configuration.
+// The value could be either a number or a string.
+const v = <T extends number|string|boolean>(name: string): T => {
```
josephperrott, line 8: I think it would be better to have this be named something more intuitive.

Maybe `getInputValue`?

github-actions/feature-request/src/octokit.ts
josephperrott: The value is always presented because on creation it becomes updated_at the same time as creation.

josephperrott (approved):
LGTM overall, just a couple little comments.

github-actions/feature-request/src/action.ts
mgechev: A practice I use to reduce the risk of overriding a function if I have a second declaration with the same name.

josephperrott (approved):
LGTM

AndrewKushnir (approved):
LGTM, thanks @mgechev 👍

PR #86, opened 2021-05-05
Improve comment content and comment mechanism

mgechev:
- Add more context in the heads up comment for old issues
- Make sure we don't comment twice on issues with insufficient number of votes
- Set a limit configuration property to allow us go through only a limited number of issues

### DIFF
--- a/github-actions/feature-request/action.yml
+++ b/github-actions/feature-request/action.yml
@@ -44,14 +44,16 @@ inputs:
       Thank you for submitting your feature request! Looks like during the polling
       process it didn't collect a sufficient number of votes to move to the next stage.
 
+
       We want to keep Angular rich and ergonomic and at the same time be mindful about
       its scope and learning journey. If you think your request could live outside
       Angular's scope, we'd encourage you to collaborate with the
       [community](https://angular.io/resources?category=community) on publishing it
-      as an open source [package](https://angular.io/guide/creating-libraries)!
+      as an open source [package](https://angular.io/guide/creating-libraries).
+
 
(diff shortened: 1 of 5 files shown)


github-actions/feature-request/action.yml
```diff
@@ -38,32 +38,42 @@ inputs:
...
       its scope and learning journey. If you think your request could live outside
       Angular's scope, we'd encourage you to collaborate with the
       [community](https://angular.io/resources?category=community) on publishing it
-      as an open source [package](https://angular.io/guide/creating-libraries)!
+      as an open source [package](https://angular.io/guide/creating-libraries).
+
 
       You can find more details about the feature request process in our
-      [documentation](https://github.com/angular/angular/blob/automation-docs/docs/GITHUB_PROCESS.md#feature-request-process).
+      [documentation](https://github.com/angular/angular/blob/master/docs/GITHUB_PROCESS.md#feature-request-process).
```
josephperrott, line 59: We should point to a specific sha so that if the file ever moves or changes, the link we put in a comment still works.

josephperrott (changes requested):
LGTM

github-actions/feature-request/action.yml
mgechev: Opening a PR to framework right now. Once we merge it, I suppose we can get the change in together with the workflow configuration.

josephperrott (approved):
LGTM

Once we have the framework sha to rely on.

josephperrott (approved):
LGTM

PR #89, opened 2021-05-10
Make changes to infrastructure for better stability consistency

josephperrott:
See individual commits.


github-actions/feature-request/BUILD.bazel
```diff
@@ -0,0 +1,35 @@
+load("//tools:esbuild.bzl", "esbuild")
```
devversion, line 1: What do you think of moving all code into a folder like `src/`?

devversion (approved):

github-actions/feature-request/BUILD.bazel
josephperrott: My thinking was to have each of the "projects" within dev-infra in their own base directories.

- `/github-actions/`
- `/ng-dev/`

devversion: How about `src/github-actions` or `src/ng-dev`? not feeling strongly about it. I'm just liking this in case we add tooling around this repository that lives at the top-level too. e.g. `scripts/`, `tools/`, `test/` etc.

PR #96, opened 2021-06-02
fix: proper parsing of input parameters

mgechev:
The input parser was not processing properly negative numbers, which was
causing the bot to exit early.

### DIFF
--- a/github-actions/feature-request/src/get-input.ts
+++ b/github-actions/feature-request/src/get-input.ts
@@ -1,0 +1,28 @@
+import * as core from '@actions/core';
+
+interface InputProvider {
+  getInput(name: string): string;
+}
+
+// Gets a specific value from the YAML configuration.
+// The value could be either a number or a string.
+export const getInputValue = <T extends number | string | boolean>(
+  name: string,
+  _core: InputProvider = core,
+): T => {
+  const result = _core.getInput(name);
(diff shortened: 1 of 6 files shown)


github-actions/feature-request/src/get-input.ts
```diff
@@ -0,0 +1,28 @@
...
+    return (result === 'true' ? true : false) as T;
+  }
+  if (!/^-?\d+(\.?\d*)$/.test(result)) {
+    return result as T;
+  }
+  const num = parseFloat(result);
+  if (isNaN(num)) {
+    throw new Error(`Can't parse ${name} as a numeric value.`);
+  }
+  return num as T;
```
josephperrott, line 27:
```suggestion
  if (!isNaN(result)) {
    return parseFloat(result) as T;
  }
  return result as T;
```

josephperrott (approved):
Just one suggestion as an alternative way to parse.

github-actions/feature-request/src/get-input.ts
mgechev: Interesting! Did not know `isNaN` casts to number if we pass a string.

mgechev: TypeScript is unhappy though because `isNaN` expects a number.

PR #98, opened 2021-07-23
refactor: update version of @octokit/rest and accompanying packages

josephperrott:


github-actions/feature-request/src/octokit.ts
```diff
@@ -30,36 +32,36 @@ export class OctoKit implements GitHubAPI {
...
-    });
-
     let pageNum = 0;
 
     const timeout = async () => {
       if (pageNum < 2) {
         /* pass */
       } else if (pageNum < 4) {
-        await new Promise((resolve) => setTimeout(resolve, 10000));
+        await setTimeout.__promisify__(10000);
```
devversion, line 45: Nice. I did not know about that 😄

github-actions/feature-request/src/octokit.ts
```diff
@@ -77,39 +79,37 @@ export class OctoKit implements GitHubAPI {
...
+    const response = this.octokit.paginate.iterator(this.octokit.orgs.listMembers, {
+      org,
+      per_page: 100,
+    });
 
     for await (const page of response) {
-      for (const user of page.data as Octokit.OrgsListMembersResponse) {
-        this._orgMembers.add(user.login);
+      for (const user of page.data) {
+        this._orgMembers.add(user!.login);
```
devversion, line 89: In what case is `user` null/undefined?

devversion (approved):

github-actions/feature-request/src/octokit.ts
josephperrott: In the typings evidently, but from the results that are actually made its not.

devversion: okay, definitely surprising because then a page response would have an array with a `undefined` or `null` value inside. so yeah, probably an issue with types.

PR #108, opened 2021-07-26
refactor: use ts_library for building github actions

josephperrott:


github-actions/feature-request/BUILD.bazel
```diff
@@ -1,36 +1,47 @@
...
+        exclude = ["src/*.spec.ts"],
+    ),
     deps = [
         "@npm//@actions/core",
         "@npm//@actions/github",
         "@npm//@octokit/rest",
         "@npm//@types/jasmine",
+        "@npm//@types/node",
         "@npm//github-app-installation-token",
         "@npm//jasmine",
```
clydin, line 19: is jasmine needed outside of testing?

tools/defaults.bzl
```diff
@@ -1,9 +1,18 @@
 load("@npm//@bazel/jasmine:index.bzl", _jasmine_node_test = "jasmine_node_test")
-load("@npm//@bazel/typescript:index.bzl", _ts_project = "ts_project")
+load("@npm//@bazel/typescript:index.bzl", _ts_library = "ts_library", _ts_project = "ts_project")
+
+def jasmine_node_test(name, specs = [], **kwargs):
+    native.filegroup(
+        name = "%s_js_files",
```
devversion, line 6: Why is the filegroup needed? The Jasmine rule would automatically extract specs from the deps. Is there a specific use-case?

devversion (approved):
LGTM. One minor comment

tools/defaults.bzl
josephperrott: Because we have set `use_direct_specs` to be `True`, we have to provide the specific spec javascript files as srcs in `jasmine_node_test`

PR #109, opened 2021-07-26
Set up build rules and tooling for newly imported areas

josephperrott:


tslint-rules/BUILD.bazel
```diff
@@ -1,13 +1,20 @@
...
     deps = [
-        "@npm//tslib",
         "@npm//tslint",
         "@npm//typescript",
     ],
 )
+
+pkg_npm(
+    name = "npm_package",
+    build_package_json_from_template = True,
```
devversion, line 16: Why do the individual sub-folders need a `package.json` at all? Wouldn't we just use the project root one this time? with proper separation of `devDependencies` and `dependencies`.

josephperrott: The idea was that we actually are creating 3 distinct packages, so the deps that are listed for each should be managed independently of eachother.

devversion: how would that reflect in NPM? since we publish the dev-infra package to NPM as a single package (is that still the plan?), the nested `package.json` dependencies don't really have an effect.

I assume it would be more convenient (and the standard way) to have a single root package.json (for intelli-sense) that we need to maintain? WDYT

josephperrott: I was under the impression/understanding we were going to be splitting into three packages: `ng-dev`, `tslint-rules` and `bazel`

bazel/BUILD.bazel
```diff
@@ -1,13 +1,30 @@
...
+        "//:__subpackages__",
+    ],
     srcs = [
         "BUILD.bazel",
         "expand_template.bzl",
         "extract_js_module_output.bzl",
-        "//dev-infra/bazel/api-golden:files",
-        "//dev-infra/bazel/browsers:files",
-        "//dev-infra/bazel/remote-execution:files",
+        "index.bzl",
```
devversion, line 24: I think the `index.bzl` file needs to go into the project root as otherwise `rules_nodejs` will not consider it as a package that can be loaded from within Bazel.

bazel/api-golden/BUILD.bazel
```diff
@@ -15,9 +15,6 @@ ts_library(
         "index_npm_packages.ts",
         "test_api_report.ts",
     ],
-    # A tsconfig needs to be specified as otherwise `ts_library` will look for the config
-    # in `//:package.json` and this breaks when the BUILD file is copied to `@npm//`.
-    tsconfig = "//dev-infra:tsconfig.json",
```
devversion, line 20: Removing this will break consumption of the dev-infra package within Bazel in general.

bazel/package.json.tmpl
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@angular/dev-infra-private",
+  "version": "0.0.0",
+  "description": "INTERNAL USE ONLY - Angular internal DevInfra tooling/scripts - INTERNAL USE ONLY",
+  "license": "MIT",
+  "private": true
+}
```
devversion, line 7: This file  can be removed now right?

ng-dev/pr/merge/BUILD.bazel
```diff
@@ -34,27 +34,18 @@ ts_library(
...
-    # Disable the Bazel patched module resolution. It always loads ".mjs" files first. This
-    # breaks NodeJS execution for "node-fetch" as it uses experimental modules which are not
-    # enabled in NodeJS. TODO: Remove this with rules_nodejs 3.x where patching is optional.
-    # https://github.com/bazelbuild/rules_nodejs/commit/7d070ffadf9c3b41711382a4737b995f987c14fa.
-    args = ["--nobazel_patch_module_resolver"],
-    deps = [
+    specs = [
         ":test_lib",
-        "@npm//node-fetch",
-        "@npm//semver",
```
devversion, line 58: Out of curiosity, do you know why these deps can be removed now? do the tests still pass?

ng-dev/release/publish/test/common.spec.ts
```diff
@@ -35,7 +35,7 @@ describe('common release action logic', () => {
     latest: new ReleaseTrain('10.0.x', parse('10.0.1')),
   };
 
-  describe('version computation', async () => {
+  describe('version computation', () => {
```
devversion, line 38: 👍

ng-dev/utils/BUILD.bazel
```diff
@@ -9,8 +9,8 @@ ts_library(
 
     # prodmode target must be set to es5 as the Octokit class, which is extended is not a proper
     # prototyped object.
-    prodmode_target = "es5",
-    visibility = ["//dev-infra:__subpackages__"],
+    #prodmode_target = "es5",
```
devversion, line 12: Did you test something out here? should this be re-added or what's the plan?

ng-dev/utils/config.ts
```diff
@@ -114,7 +115,7 @@ function readConfigFile(configPath: string, returnEmptyObjectOnError = false): o
     // dev-infra tool runs in NodeJS which does not support ES modules by default.
     // Additionally, set the `dir` option to the directory that contains the configuration
     // file. This allows for custom compiler options (such as `--strict`).
-    require('ts-node').register({
+    register({
```
devversion, line 118: We intentionally load this dynamically as otherwise we'd always require `ts-node`. I'm not sure where we stand here. I guess all configs use TypeScript but seems like this would be better in a separate change cleaning this up completely if we go that way.

tools/BUILD.bazel
```diff
@@ -1,8 +1,14 @@
...
 
 ts_config(
     name = "tsconfig",
     src = "//:tsconfig.json",
 )
+
+nodejs_binary(
+    name = "inline-package-json-deps",
+    entry_point = "//tools:inline-package-json-deps.js",
+)
```
devversion, line 14: This can be removed now right?

devversion: or do you want to keep it for now if we go consider the individual NPM packages in the future?

tools/defaults.bzl
```diff
@@ -30,7 +31,7 @@ def ts_library(name, testonly = False, deps = [], **kwargs):
         devmode_target = "es2020",
         prodmode_module = "commonjs",
         prodmode_target = "es2020",
-        tsconfig = "//tools:tsconfig",
+        tsconfig = kwargs.pop("tsconfig", "//tools:tsconfig"),
```
devversion, line 34: Nit: Can you add a TODO to clean this up? I think we'd not want to use the `tools/tsconfig` ever for code not being in `tools`?

devversion (approved):
🎉

ng-dev/pr/merge/BUILD.bazel
josephperrott: Fetch and semver aren't actually used in these tests, so they were being provided needlessly

devversion: 👍

ng-dev/utils/BUILD.bazel
josephperrott: We have since moved away from the problematic Octokit class with our upgrade of `@octokit/*`

devversion: Ah nice! we should remove the comment then. 👍

devversion (approved):

PR #111, opened 2021-07-27
Setup ng-dev locally and use ng-dev for formatting

josephperrott:


.ng-dev/config.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const github: GithubConfig = {
+  owner: 'angular',
+  name: 'dev-infra',
+};
+
+module.exports = {
+  format,
+  github,
+};
```
devversion, line 17: I see that you have `format` and `github` set here, but also have individual files. I think we should just re-export the exports from `format.ts` and `github.ts` right?

devversion (approved):

.ng-dev/config.ts
josephperrott: Whoops. Fixed.

PR #112, opened 2021-07-28
ci: setup circleci for linting and testing

josephperrott:


.circleci/config.yml
```diff
@@ -0,0 +1,117 @@
...
+cache_key: &cache_key v7-{{ checksum ".bazelversion" }}-{{ checksum "WORKSPACE" }}-{{ checksum "yarn.lock" }}
+# We want to invalidate the cache if the postinstall patches change. In order to apply new
+# patches, a clean version of the node modules is needed. Additionally, we invalidate the cache
+# if the Bazel version changes. We do this because otherwise the `bazelisk` cache folder will
+# contain all previously used versions and ultimately cause the cache restoring to be slower.
+cache_fallback_key: &cache_fallback_key v7-{{ checksum ".bazelversion" }}-
+
+gcp_decrypt_token: &gcp_decrypt_token "angular"
+
+restore_cache: &restore_cache
```
devversion, line 17: FWIW: I prefer the way do this in the comp repo where we name it `var_1: &<name>`. This avoids the ambiguity where you could think that `restore_cache` is an actual config setting from CircleCI here.

.circleci/config.yml
```diff
@@ -0,0 +1,117 @@
...
+  test:
+    executor: default-executor
+    steps:
+      - checkout_and_rebase
+      - *restore_cache
+      - setup_bazel_config
+      - run:
+          name: yarn install
+          command: yarn install --frozen-lockfile --non-interactive
+          no_output_timeout: 45m
```
devversion, line 92: Out of curiosity: why such a large timeout? I think if it takes over ~2min then something is wrong! 😄

devversion (approved):

.circleci/config.yml
josephperrott: I copied this over from our previous configurations, I believe its a hold over from when we were running into issues with yarn in the past.  Updated.

devversion (approved):

PR #118, opened 2021-07-30
fix(dev-infra): browser archive rule should handle `.dmg` files (#42992)

devversion:
Backport of https://github.com/angular/angular/commit/bfd48391d7a91ff42739a2c4007743a81748342f.

We recently reworked our browser archive extraction to happen
at analysis time for better caching. This resulted in us breaking
the extraction of macOS dmg files so that Firefox is currently
not usable for local testing on macOS. We implement a similar special
logic for `.dmg` files to what has been done within the Bazel
webtesting rules.

PR Close #42992

### DIFF
--- a/bazel/browsers/browser_archive_repo.bzl
+++ b/bazel/browsers/browser_archive_repo.bzl
@@ -1,11 +1,27 @@
 """Implementation of the `browser_archive` rule."""
 
 def _browser_archive_impl(ctx):
-    ctx.report_progress("Downloading browser archive from: %s" % ctx.attr.url)
-    ctx.download_and_extract(
-        url = ctx.attr.url,
-        sha256 = ctx.attr.sha256,
-    )
+    url = ctx.attr.url
+    sha256 = ctx.attr.sha256
+
+    # If the URL resolves to a `.dmg` file, then we need to convert the file
+    # to a zip so that we can extract the actual binaries. We use the `convert_dmg`
+    # script provided by the webtesting Bazel rules.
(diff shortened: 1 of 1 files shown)


bazel/browsers/browser_archive_repo.bzl
```diff
@@ -1,11 +1,27 @@
...
+
+    # If the URL resolves to a `.dmg` file, then we need to convert the file
+    # to a zip so that we can extract the actual binaries. We use the `convert_dmg`
+    # script provided by the webtesting Bazel rules.
+    if url.endswith(".dmg"):
+        download_file_name = "_download_file_%s.dmg" % ctx.attr.name
+        result_zip_name = "_converted_file_%s.zip" % ctx.attr.name
+
+        ctx.download(url, download_file_name, sha256)
+        ctx.execute([ctx.path(Label("@io_bazel_rules_webtesting//web/internal:convert_dmg.sh")), download_file_name, result_zip_name])
```
clydin, line 15: is this script location and function fairly stable?
I suppose it would fail in an obvious way if it changed.

clydin (approved):

bazel/browsers/browser_archive_repo.bzl
devversion: yeah, we rely on some rules webtesting internals in other places as well. And yes, Bazel will report an error if something does not work, or in the last place if the extraction fails

devversion:
Closed by commit aaf68585ed51e34e61bcb5b61c21be2be2e8882c

PR #119, opened 2021-08-03
Support different main branches

devversion:
See individual commits


devversion:
This came up because running `yarn ng-dev format changed` currently does not work in this repo! 😄

devversion:
@josephperrott Looks like esbuild is failing. Windows and Linux seem to be different here 😢 I think we should address this in a separate pull request.

ng-dev/pr/merge/defaults/labels.ts
```diff
@@ -57,8 +64,8 @@ export async function getDefaultTargetLabelConfiguration(
     {
       pattern: 'target: minor',
       // Changes labeled with `target: minor` are merged most commonly into the next branch
-      // (i.e. `master`). In rare cases of an exceptional minor version while being already
-      // on a major release train, this would need to be overridden manually.
+      // (i.e. `main`/`master`). In rare cases of an exceptional minor version while being
```
josephperrott, line 67:
```suggestion
      // (i.e. `main`). In rare cases of an exceptional minor version while being
```

Since we are just providing an example, lets use the preferred branch name as the example

ng-dev/pr/merge/strategies/autosquash-merge.ts
```diff
@@ -93,11 +93,11 @@ export class AutosquashMergeStrategy extends MergeStrategy {
 
     this.pushTargetBranchesUpstream(targetBranches);
 
-    // For PRs which do not target the `master` branch on Github, Github does not automatically
-    // close the PR when its commit is pushed into the repository.  To ensure these PRs are
+    // For PRs which do not target the main branch on Github, Github does not automatically
```
josephperrott, line 96:
```suggestion
    // For PRs which do not target the `main` branch on Github, Github does not automatically
```

tools/defaults.bzl
```diff
@@ -38,31 +38,40 @@ def ts_library(name, testonly = False, deps = [], srcs = [], **kwargs):
...
 
-def esbuild(
-        name,
-        entry_point,
-        minify = True,
-        sourcemap = "inline",
-        save_to_repo = False,
-        external = [],
-        deps = []):
+def esbuild(platform = "node", target = "node12", **kwargs):
```
josephperrott, line 41: Is there a case where we don't want to  always target the same version?

ng-dev/utils/config.ts
```diff
@@ -23,6 +23,8 @@ export interface GitClientConfig {
   useSsh?: boolean;
   /** Whether the specified repository is private. */
   private?: boolean;
+  /** Main branch name for the repository. */
+  mainBranchName?: string;
```
josephperrott, line 27:
```suggestion
  mainBranchName: string;
```


I think we should just begin requiring this as a value in the config, though I could be convinced to not require it and have defaults elsewhere.  My concern is about having to constantly manage the default value at the usage site.

ng-dev/pr/merge/strategies/autosquash-merge.ts
```diff
@@ -93,11 +93,11 @@ export class AutosquashMergeStrategy extends MergeStrategy {
...
+    // close the PR when its commit is pushed into the repository. To ensure these PRs are
     // correctly marked as closed, we must detect this situation and close the PR via the API after
     // the upstream pushes are completed.
-    if (githubTargetBranch !== 'master') {
+    if (githubTargetBranch !== this.git.mainBranchName) {
```
josephperrott, line 100:
```suggestion
    if (githubTargetBranch !== getNextBranchName(this.git.mainBranchName)) {
```

ng-dev/release/versioning/active-release-trains.ts
```diff
@@ -26,20 +26,18 @@ export interface ActiveReleaseTrains {
...
 ): Promise<ActiveReleaseTrains> {
+  const nextBranchName = repo.nextBranchName;
   const nextVersion = await getVersionOfBranch(repo, nextBranchName);
   const next = new ReleaseTrain(nextBranchName, nextVersion);
   const majorVersionsToConsider: number[] = [];
   let expectedReleaseCandidateMajor: number;
 
-  // If the `next` branch (i.e. `master` branch) is for an upcoming major version, we know
-  // that there is no patch branch or feature-freeze/release-candidate branch for this major
+  // If the `next` branch (i.e. `main` or `master` branch) is for an upcoming major version, we
```
josephperrott, line 39:
```suggestion
  // If the `next` branch (i.e. `main`) is for an upcoming major version, we
```

ng-dev/utils/git/git-client.ts
```diff
@@ -44,6 +44,9 @@ export class GitClient {
   /** Octokit request parameters object for targeting the configured remote. */
   readonly remoteParams = {owner: this.remoteConfig.owner, repo: this.remoteConfig.name};
 
+  /** Name of the primary branch of the upstream remote. */
+  readonly mainBranchName = this.config.github.mainBranchName ?? 'master';
```
josephperrott, line 48:
```suggestion
  readonly mainBranchName = getNextBranchName(this.config.github.mainBranchName);
```

josephperrott (changes requested):
Additionally, we likely can remove `ng-dev/ng-dev.js` and `ng-dev/build-worker.js` as we don't rely on them locally.  Our `ng-dev` script just uses `ts-node` as for dev-infra we are willing to accept the runtime cost of ts-node.

tools/defaults.bzl
devversion: great point. let's always use this for now. I was thinking there might be cases in the future where we expose browser-specific bundles but that's something we can change once it's the case.

ng-dev/pr/merge/strategies/autosquash-merge.ts
devversion: As discussed offline. This should be different to the next branch, but rather about the main branch set for the repository. Could be different to the `next` release-train branch

josephperrott (approved):
LGTM

josephperrott:
Closed by commit c106aa50e8a6ddd4f8d5e391496c49d1de420b31

PR #120, opened 2021-08-05
ci: publish a snapshot build for each commit which creates a changed npm package output

josephperrott:


.circleci/config.yml
```diff
@@ -109,9 +109,31 @@ jobs:
...
+  publish_snapshot_build:
+    executor: default-executor
+    steps:
+      - checkout_and_rebase
+      - *restore_cache
+      - setup_bazel_config
+      - run:
+          name: yarn install
+          command: yarn install --frozen-lockfile --non-interactive
+          no_output_timeout: 45m
```
devversion, line 121:
```suggestion
          no_output_timeout: 2m
```

.circleci/publish_to_github.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+
+# TODO(josephperrott): Remove this script in favor of common orb to perform the task.
```
devversion, line 3: I assume something like this could also be a nice script as part of the shared dev-infra package (written in JS)?

.circleci/publish_to_github.sh
```diff
@@ -0,0 +1,87 @@
...
+  exit 1
+fi
+
+# Function to publish artifacts of a package to Github.
+publishPackage() {
+  packageName="dev-infra-private"
+  packageRepo="dev-infra-private-builds"
+  buildDir="$(pwd)/dist/bin/npm_package"
+  buildVersion=$(node -pe "require('./package.json').version")
+  branchName="master"
```
devversion, line 26: You might want to keep the `CIRCLE_BRANCH` name like in the COMP repo (in case we ever want to deploy snapshots for e.g. a development branch; we can then control the deployment in the Circle config)

.circleci/publish_to_github.sh
```diff
@@ -0,0 +1,87 @@
...
+    echo "Cloning default branch and creating branch '${branchName}' on top of it."
+
+    git clone ${repoUrl} ${repoDir} --depth 1
+    echo "Cloned repository and switched into directory. Creating new branch now.."
+
+    git checkout -b ${branchName}
+  fi
+
+  # Copy the build files to the repository
+  cd ${repoDir}
```
devversion, line 63: If I look at the script in the COMP repo, the `cd` was done in the if-else clause above. If I don't miss anything, I think currently `git checkout -b` above will fail as we are not in the repo? should we keep it like in the COMP repo?

.circleci/publish_to_github.sh
```diff
@@ -0,0 +1,87 @@
...
+  echo "https://${DEV_INFRA_BUILDS_TOKEN}:@github.com" > .git/credentials
+
+  echo "Git configuration has been updated to match the last commit author. Publishing now.."
+
+  git add -A
+
+  if git diff-index --quiet HEAD --; then
+    echo "Skipping push as no changes occured between this push and the previous commit."
+  else
+    git commit --allow-empty -m "${buildCommitMessage}"
```
devversion, line 81: Nit: This cannot be empty right due to the check you added?

```suggestion
    git commit  -m "${buildCommitMessage}"
```

.circleci/publish_to_github.sh
```diff
@@ -0,0 +1,87 @@
...
+    git push origin ${branchName} --force
+    echo "Published package artifacts for ${packageName}#${buildVersionName} into ${branchName}"
+  fi
+}
+
+publishPackage "dev-infra-private" "dev-infra-private-builds"
```
devversion, line 87: You hard-coded the package and repo name so I think this is not needed, or you re-add the `$1` and `$2` substitutions (in order to keep this script more flexible).

devversion (approved):
LGTM. A few minor comments

.circleci/publish_to_github.sh
josephperrott: Yeah, thats the general idea, I am not sure of the exact implementation, but the overall idea sounds correct to me.

devversion (approved):

PR #123, opened 2021-08-05
ci: use authenticated url for origin to authenticate git push for snapshots

josephperrott:


devversion (approved):
That seems more clean. Although not sure how it shows up in the console. i.e if it will leak the token?

josephperrott:
@devversion I confirmed it does not try to print to the stdout or stderr.  Additionally, for environment variables, when they are printed to the output CircleCI `[redacts]` them automatically.

PR #124, opened 2021-08-05
build: include generated javascript files for ng-dev tooling to allow for importing

josephperrott:
Including the generated javascript files for the `ng-dev` toolset allows for downstream repos to import the exported items from the code.  When only including the bundled executables, we don't provide proper entry points for accessing the utils/config typings.

Additionally, we update the `bin` to note we would like to run the `-bundle` file rather than the unbundled version.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -7,11 +7,6 @@ ts_config(
     src = "tsconfig.json",
     visibility = ["//visibility:public"],
 )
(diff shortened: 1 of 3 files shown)


devversion (approved):

devversion:
Closed by commit 8bd39aa690bfd5e2a711b3779fb4ae35e2baf3c1

PR #131, opened 2021-08-11
fix(ng-dev): remove python3 check for publish command

devversion:
In the past, our repositories install Skydoc which relied
on python. We removed skydoc in the Framework repository,
and other major repositories of the Angular org are not
using it. We can remove this requirement. The instructions
did not work for the caretaker in the components repo on
a Mac (not clear why specifically), but we do not need this
check anyway and it seems rather wrong to have here anyway.


josephperrott (approved):
LGTM

PR #132, opened 2021-08-12
refactor(ng-dev): move breaking change and deprecations to beginning of release notes

devversion:
Various improvements for release notes. See individual commits.

### DIFF
--- a/ng-dev/release/publish/test/release-notes/generation.spec.ts
+++ b/ng-dev/release/publish/test/release-notes/generation.spec.ts
@@ -58,6 +58,7 @@ describe('release notes generation', () => {
         ### cdk/a11y
         | Commit | Description |
         | -- | -- |
+        | <..> | refactor(cdk/a11y): with breaking change |
         | <..> | fix(cdk/a11y): not yet released #1 |
         ## Special Thanks:
       `);
@@ -82,6 +83,7 @@ describe('release notes generation', () => {
         ### cdk/a11y
         | Commit | Description |
         | -- | -- |
+        | <..> | refactor(cdk/a11y): with deprecation |
         | <..> | fix(cdk/a11y): not yet released #1 |
         ## Special Thanks:
       `);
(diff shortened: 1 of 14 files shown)


ng-dev/release/publish/test/release-notes/generation.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+describe('release notes generation', () => {
+  let releaseConfig: ReleaseConfig;
+  let githubConfig: GithubConfig;
+  let client: SandboxGitClient;
+
+  beforeEach(() => {
+    // Clear the temporary directory used by the sandbox git client. We do not want
+    // the repo state to persist between tests.
+    rmdirSync(testTmpDir, {recursive: true});
```
josephperrott, line 27: Should we perform this clean of the temp directory as part of the `getMockGitClient` function, if I remember correctly this is duplicated in another place where `getMockClient` is used.

josephperrott (approved):
LGTM

ng-dev/release/publish/test/release-notes/generation.spec.ts
devversion: I have looked into this as well; but couldn't come up with a good way to de-duplicate this without "hiding stuff". Doing it inside `getMockGitClient` seems unexpected because I could totally call this method twice but then I'd not expect it to clear the whole temporary directory. Also it's not specified that `getMockGitClient` necessarily runs at the beginning of a test right? That's why it currently exists here in `beforeEach` and for the release action tests within the `setup<..>Action` test function.

josephperrott: Hmm okay, lets leave it as is then and we can keep it in mind for if we find a way in the future.

devversion: yeah, happy to change if we can come up with something better. I think at this point it's duplicated, but more clear than doing the cleanup somewhere inside. (note I have also considered doing it inside `SandboxGitClient.createInstance`)

PR #136, opened 2021-08-12
feat(ng-dev): remove redundant commit message scope

alan-agius4:
With this change we remove the commit message scope from the list of commits. This is currently redundant since the scope is listed as header.

**Changelog**
**Before**
```md
### @schematics/angular
| Commit | Description |
| -- | -- |
| 268a03b63 | feat(@schematics/angular): add `noImplicitOverride` and `noPropertyAccessFromIndexSignature` to workspace tsconfig |
| 5986befcd | feat(@schematics/angular): update ngsw-config resources extensions |
```

**After**
```md
### @schematics/angular
| Commit | Description |
| -- | -- |
| 3ba13f467| feat: add `noImplicitOverride` and `noPropertyAccessFromIndexSignature` to workspace tsconfig |
| a7b2e6f51| feat: update ngsw-config resources extensions |
```

**Github release**
For GitHub releases we now also add badges. We do this only for GitHub releases since the changelog.md can be quite long and each badge will result in a seperate HTTP call which can end up in hundreds of these.

Screenshot 2021-08-12 at 14 30 52

### DIFF
--- a/ng-dev/release/notes/context.ts
+++ b/ng-dev/release/notes/context.ts
@@ -175,6 +175,27 @@ export class RenderContext {
       .filter((a) => !botsAuthorNames.includes(a))
       .sort();
   }
+
+  /**
+   * Convert a commit object to a Markdown linked badged.
+   */
+  commitToBadge(commit: CommitFromGitLog): string {
+    let color = 'yellow';
+    switch (commit.type) {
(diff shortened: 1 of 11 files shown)


josephperrott (approved):
LGTM

PR #137, opened 2021-08-12
feat(ng-dev): support using no targeting labels for merge

josephperrott:
Allow the configuration of no target labels instead always merging into the main branch provided in the `GithubConfig`.

### DIFF
--- a/ng-dev/pr/merge/config.ts
+++ b/ng-dev/pr/merge/config.ts
@@ -77,20 +77,10 @@ export interface MergeConfig {
   targetLabelExemptScopes?: string[];
 }
 
-/**
- * Configuration of the merge script in the dev-infra configuration. Note that the
- * merge configuration is retrieved lazily as usually these configurations rely
- * on branch name computations. We don't want to run these immediately whenever
- * the dev-infra configuration is loaded as that could slow-down other commands.
- */
-export type DevInfraMergeConfig = {
-  merge: MergeConfig;
-};
-
 /** Loads and validates the merge configuration. */
 export function assertValidMergeConfig<T>(
-  config: T & Partial<DevInfraMergeConfig>,
-): asserts config is T & DevInfraMergeConfig {
+  config: T & Partial<{merge: MergeConfig}>,
+): asserts config is T & {merge: MergeConfig} {
   const errors: string[] = [];
   if (config.merge === undefined) {
     throw new ConfigValidationError('No merge configuration found. Set the `merge` configuration.');
--- a/ng-dev/pr/merge/config.ts
+++ b/ng-dev/pr/merge/config.ts
@@ -52,7 +52,7 @@ export interface MergeConfig {
    */
   remote?: GithubConfig;
   /** List of target labels. */
-  labels: TargetLabel[];
+  noTargetLabeling?: boolean;
   /** Required base commits for given branches. */
   requiredBaseCommits?: {[branchName: string]: string};
   /** Pattern that matches labels which imply a signed CLA. */
@@ -96,11 +96,6 @@ export function assertValidMergeConfig<T>(
     throw new ConfigValidationError('No merge configuration found. Set the `merge` configuration.');
   }
 
-  if (!config.merge.labels) {
-    errors.push('No label configuration.');
-  } else if (!Array.isArray(config.merge.labels)) {
-    errors.push('Label configuration needs to be an array.');
-  }
   if (!config.merge.claSignedLabel) {
     errors.push('No CLA signed label configured.');
   }
--- a/ng-dev/pr/merge/config.ts
+++ b/ng-dev/pr/merge/config.ts
@@ -6,8 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {GithubConfig} from '../../utils/config';
-import {GithubClient} from '../../utils/git/github';
+import {ConfigValidationError, GithubConfig} from '../../utils/config';
 
 import {GithubApiMergeStrategyConfig} from './strategies/api-merge';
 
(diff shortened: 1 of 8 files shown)


ng-dev/pr/merge/pull-request.ts
```diff
@@ -85,15 +85,17 @@ export async function loadAndValidatePullRequest(
...
+    const commitsInPr = prData.commits.nodes.map((n) => parseCommitMessage(n.commit.message));
 
-  try {
-    assertPendingState(prData);
-    assertChangesAllowForTargetLabel(commitsInPr, targetLabel, config);
-    assertCorrectBreakingChangeLabeling(commitsInPr, labels, config);
-  } catch (error) {
-    return error;
+    try {
+      assertPendingState(prData);
```
devversion, line 93: I feel like the pending state and correct breaking change stuff is unrelated to the labeö
config and should still run?

ng-dev/pr/merge/target-label.ts
```diff
@@ -30,6 +31,12 @@ export function getTargetLabelFromPullRequest(
   config: Pick<MergeConfig, 'labels'>,
   labels: string[],
 ): TargetLabel {
+  if (config.labels === false) {
```
devversion, line 34: Nit: I think it would be cleaner to create the labels config when this method is invoked instead. This method shouldn't really fetch the ng-dev config or deal with the defaults IMO.

devversion: Also, the git client with the main branch name would exist there already

ng-dev/pr/merge/pull-request.ts
josephperrott: Good call.

ng-dev/pr/merge/target-label.ts
josephperrott: Would it make sense instead to change the config to drop `labels` and add `noBranchTargeting?: boolean`

Then branch targeting can turned off entirely, so the options become either use the common branch targeting or just perform merges to wherever the PR targets in Github?

devversion: great idea 👍

josephperrott:
@devversion I kind of redid everything to move to having a property `noTargetLabeling` instead.

This is now built on top of #170

ng-dev/pr/merge/defaults/labels.ts
```diff
@@ -32,15 +33,25 @@ import {assertActiveLtsBranch} from './lts-branch';
...
  */
-export async function getDefaultTargetLabelConfiguration(
-  api: GithubClient,
-  githubConfig: GitClientConfig,
-  releaseConfig: ReleaseConfig,
+export async function getTargetLabels(): Promise<TargetLabel[]>;
+export async function getTargetLabels(
+  _api?: GithubClient,
+  _config?: Partial<{github: GitClientConfig; release: ReleaseConfig}>,
+): Promise<TargetLabel[]>;
```
devversion, line 40: Are these overloads really needed here? Seems like the implementation signature already allows for these? We should remove them for readability.

ng-dev/pr/merge/defaults/labels.ts
```diff
@@ -32,15 +33,25 @@ import {assertActiveLtsBranch} from './lts-branch';
...
+export async function getTargetLabels(
+  _api?: GithubClient,
+  _config?: Partial<{github: GitClientConfig; release: ReleaseConfig}>,
 ): Promise<TargetLabel[]> {
-  const nextBranchName = getNextBranchName(githubConfig);
+  const api = _api || GitClient.get().github;
+
+  const config = _config || (getConfig() as any);
```
devversion, line 47: We should use `??` instead of `||` for falling back (for `_api` as well). You could also inline the `GitClient.get().github` as default parameter IIUC. Maybe we should have a small local function here that does the same for the config (with the both assertions; so we do not need the weird `as any` cast here)

ng-dev/pr/merge/pull-request.ts
```diff
@@ -73,22 +73,37 @@ export async function loadAndValidatePullRequest(
...
+
+  const githubTargetBranch = prData.baseRefName;
   let targetLabel: TargetLabel;
-  try {
-    targetLabel = getTargetLabelFromPullRequest(config, labels);
-  } catch (error) {
-    if (error instanceof InvalidTargetLabelError) {
-      return new PullRequestFailure(error.failureMessage);
+  let targetBranches: string[];
+  if (!config.noTargetLabeling) {
```
devversion, line 82: I'd probably swap the clauses here so that we do not double-negate.

ng-dev/pr/merge/defaults/labels.ts
devversion: While reading below where `getTargetLabels` is called, I think we should just always require people to pass a `GithubClient` and the `config`. That is less confusing and more readable. Thoughts?

ng-dev/pr/merge/pull-request.ts
devversion: Also, how about extracting this into a dedicated function? That would also allow us to use that for testing, compared to manually having to handle the `noTargetLabeling` case in tests (this would give us more coverage).

devversion (approved):
LGTM. A few minor suggestions

ng-dev/pr/merge/defaults/labels.ts
josephperrott: The overloads are just in place so that we can provide the values in the `integration.spec.ts` tests.

ng-dev/pr/merge/defaults/labels.ts
josephperrott: I initially had them as default parameters, but when it was a default parameter I ran into issues in the tests, but now they aren't happening so it must have been something else.

devversion: I see, yeah I think just making them required is probably the best IMO.

josephperrott: I am going to hold off on making them required as we are just moving the problem up a level in terms of the call stack by making it required here.  I think there is a larger refactoring that can address it.

devversion: I think it would be expected to pass the config around, compared to always retrieving it through the single-ton, but it's true that this is probably just up to personal preference. 

Yeah, let's keep it as is and we can see how it evolves over time.

ng-dev/pr/merge/config.ts
```diff
@@ -85,49 +84,29 @@ export interface MergeConfig {
  * the dev-infra configuration is loaded as that could slow-down other commands.
```
devversion, line 84: @josephperrott can we update that comment? it might be worth preserving the contents in some way, but saying that we do not support custom configuration, not sure. can be a follow-up too.

ng-dev/pr/merge/defaults/integration.spec.ts
```diff
@@ -30,6 +31,7 @@ describe('default target labels', () => {
     api = new GithubClient();
     githubConfig = {owner: 'angular', name: 'dev-infra-test', mainBranchName: 'master'};
     releaseConfig = {
+      releaseNotes: {},
```
devversion, line 34: Nit: this is no longer needed. `releaseNotes` is an option config.

```suggestion
```

ng-dev/pr/merge/pull-request.ts
```diff
@@ -290,3 +273,32 @@ function assertPendingState(pr: RawPullRequest) {
       throw PullRequestFailure.isMerged();
   }
 }
+
+/** Get the branches the pull request will be merged into.  */
+async function getTargetBranches(
+  config: {merge: MergeConfig; github: GithubConfig},
+  labels: string[],
+  targetBranch: string,
```
devversion, line 281: We should keep the `githubTargetBranch` terminology to avoid ambiguity with actual target branches being determined. it's obviously reasonable from the context, but doesn't hurt staying more explicit.

```suggestion
  githubTargetBranch: string,
```

ng-dev/pr/merge/defaults/integration.spec.ts
```diff
@@ -104,15 +106,17 @@ describe('default target labels', () => {
...
+    const labels = await computeTargetLabels();
+    spyOn(labelDefaults, 'getTargetLabels').and.resolveTo(labels);
     let label: TargetLabel;
     try {
-      label = getTargetLabelFromPullRequest({labels}, [name]);
+      label = await getTargetLabelFromPullRequest({}, [name]);
     } catch (error) {
+      console.log(error);
+      if (error.message === 'This repository does not use target labels') {
+        return [githubTargetBranch];
```
devversion, line 118: This should be removed now. I think it's a dead path. We should fully rely on the actual logic from the merge script.

ng-dev/pr/merge/defaults/labels.ts
devversion: I don't follow how these all overloads combined, are different from just having a single signature like:

```ts
export async function getTargetLabels(
  api = GitClient.get().github,
  config = getConfig() as Partial<{github: GithubConfig; release: ReleaseConfig}>,
): Promise<TargetLabel[]> {}

// both still work; regardless of overloads
getTargetLabels();
getTargetLabels(githubApi, config);
```

devversion (approved):
LGTM. A few minor comments, but looks great overall.

ng-dev/pr/merge/defaults/labels.ts
josephperrott: I though I had removed the overloads when I was able to use the default parameters.  Fixed.

ng-dev/pr/merge/config.ts
josephperrott: I just removed it entirely as we no longer even need that type.

PR #140, opened 2021-08-12
test(ng-dev): fix usage of TEST_TMPDIR

josephperrott:
Bazel expects tests "not attempt to remove, chmod, or otherwise alter [TEST_TMPDIR]," so instead a subdirectory
should be used to be created/destroyed throughout tests.  If the TEST_TMPDIR is directly used the first usage
works as expected but removing the directory is not successful causing failure on the next creation.  This
failure does not occur on windows due to Bazel's lack of sandboxing and therefore lack of control over the
interactions with the TEST_TMPDIR.


devversion (approved):
LGTM. I wonder why this didn't fail in CI but makes sense 👍

PR #142, opened 2021-08-13
refactor(ng-dev): remove minimist usage

alan-agius4:
Use `yargs` since it's the preferred parser.


ng-dev/utils/testing/virtual-git-client.ts
```diff
@@ -99,8 +99,14 @@ export class VirtualGitClient extends AuthenticatedGitClient {
 
   /** Handler for the `git push` command. */
   private _push(args: string[]) {
-    const [repoUrl, refspec] = parseArgs(args, {boolean: ['q']})._;
-    const ref = this._unwrapRefspec(refspec);
+    const [repoUrl, refspec] = yargs(args)
```
devversion, line 102: Is there a way to manage `repoUrl` and `refspec` as "positional" here? that would save us from adding the type casts.

devversion (approved):
👍 Minimist is less verbose for the tests.. but Yargs is definitely the better choice.

ng-dev/utils/testing/virtual-git-client.ts
alan-agius4: You'd need to define commands. 
Example: https://github.com/yargs/yargs/blob/master/docs/advanced.md#advanced-topics

alan-agius4:
> Minimist is less verbose for the tests

Agreed, but since yargs is already a dependency might as well use that.

ng-dev/utils/testing/virtual-git-client.ts
devversion: which I guess could make sense as it's basically mocking a command, but I'm fine either way.

alan-agius4: I can get back to replacing this next week.

josephperrott (approved):

ng-dev/utils/testing/virtual-git-client.ts
```diff
@@ -112,15 +117,17 @@ export class VirtualGitClient extends AuthenticatedGitClient {
     if (existingPush !== undefined) {
       existingPush.head = pushedHead;
     } else {
-      this.pushed.push({remote: {repoUrl, name}, head: pushedHead});
+      this.pushed.push({remote: {repoUrl: repoUrl, name}, head: pushedHead});
```
devversion, line 120:
```suggestion
      this.pushed.push({remote: {repoUrl, name}, head: pushedHead});
```

ng-dev/utils/testing/virtual-git-client.ts
```diff
@@ -154,10 +166,19 @@ export class VirtualGitClient extends AuthenticatedGitClient {
...
+      target,
+    } = yargs(rawArgs)
+      .command('$0 <target>', false)
+      .positional('target', {demandOption: true, type: 'string'})
+      .option('detach', {boolean: true})
+      .option('q', {boolean: true})
+      .option('B', {boolean: true})
+      .parseSync();
+
+    //const target = targets[0];
```
devversion, line 181:
```suggestion
```

devversion (approved):

PR #143, opened 2021-08-13
ci: enable renovate bot

alan-agius4:
With this change dependencies are now managed via Renovate bot.

- Major updates are scheduled after 10pm and before 4am every weekday and unscheduled during weekends
- lock file maintenance, every Monday
- minor and patch updates are grouped and scheduled after 1am every Thursday

### DIFF
--- a/renovte.json
+++ b/renovte.json
@@ -1,0 +1,40 @@
+{
+  "pinVersions": false,
+  "semanticCommits": true,
+  "semanticPrefix": "build",
+  "separateMajorMinor": false,
+  "prHourlyLimit": 2,
+  "labels": ["comp: build", "dependencies"],
+  "timezone": "America/Tijuana",
+  "lockFileMaintenance": {
+    "enabled": true
+  },
+  "schedule": ["after 10pm every weekday", "before 4am every weekday", "every weekend"],
+  "baseBranches": ["main"],
+  "ignoreDeps": ["@types/node", "@types/selenium-webdriver", "selenium-webdriver"],
+  "packageFiles": ["WORKSPACE", "package.json"],
+  "packageRules": [
+    {
+      "packagePatterns": ["^@bazel/.*", "^build_bazel.*"],
+      "groupName": "bazel",
+      "pinVersions": false
+    },
+    {
+      "packageNames": ["typescript"],
+      "updateTypes": ["major"],
+      "enabled": false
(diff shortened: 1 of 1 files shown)


renovte.json
```diff
@@ -0,0 +1,40 @@
...
+  "separateMajorMinor": false,
+  "prHourlyLimit": 2,
+  "labels": ["comp: build", "dependencies"],
+  "timezone": "America/Tijuana",
+  "lockFileMaintenance": {
+    "enabled": true
+  },
+  "schedule": ["after 10pm every weekday", "before 4am every weekday", "every weekend"],
+  "baseBranches": ["main"],
+  "ignoreDeps": ["@types/node", "@types/selenium-webdriver", "selenium-webdriver"],
```
devversion, line 14: Is there context on these ignore packages?

devversion (approved):
LGTM. I personally never favorized automatic version updates as I believe updating dependencies manually encourages more thorough investigation of how the update would affect the code (especially for the Bazel stuff this is needed often).

renovte.json
alan-agius4: -`selenium-webdriver` version 4 beta which is current marked as latest.
- `@types/node` seems to be pinned to the min supported version fo the Angular framework and hence the update needs to happen manually.

devversion: 👍 thanks. good to capture this here.

alan-agius4:
> I personally never favorized automatic version updates as I believe updating dependencies manually encourages more thorough investigation of how the update would affect the code (especially for the Bazel stuff this is needed often).

I think in most cases this is only true for major versions. I personally feel that updating dependencies manually is too much of a burden. (Side note: `rules_nodejs` repo is also managed by Renovate).

In the CLI we have been using Renovate for a couple of years now and truly cannot imagine these updates to be done manually nowadays. That's just my personal opinion of course. If code coverage is low, I'd agree that it could be problematic.

devversion:
yeah, it's definitely up to preference. We've been doing very fine without Renovate/Dependabot for years in the COMP repo (now it's enabled; but not a big fan still 😄; especially if you need to push to a Renovate PR to fix it up)

josephperrott (approved):

alan-agius4:
@josephperrott note: you need to enable renovate once this is merged.

PR #147, opened 2021-08-14
feat(github-actions): make breaking changes configurable

alan-agius4:
Different repos use different labels, and hence one label doesn't fit all.

- CLI: `breaking change` (Will eventually make a PR to align with FW)
- FW:  `flag: breaking change`
- COMP: `breaking changes`


devversion:
I was thinking about this as well, but wanted to just make the label not configurable. what do you think?

that would save us from having to sync the configurations in the action and .ng-dev.

devversion (approved):
I don't feel strongly though. The label doesn't change very often, so LGTM.

alan-agius4:
I don’t feel strongly either, I made this configurable because the merge tools always this to be configured.  I am all in into having the same label across all repos.

The FW recently changed the label to `flag: breaking change` so I guess this would be the “correct” value.

josephperrott:
I think we should make the label non-configurable. 

We can go ahead and just choose what the label should be, I can see how `flag: breaking change` is nice because we could have all of the flags end up showing other if we make them similarly named.

alan-agius4:
Will open a new PR to update the label as per above.

alan-agius4:
T

PR #153, opened 2021-08-16
fix(ng-dev): update default breaking change label to `flag: breaking change`

alan-agius4:
This is to align all repos to use the same label


josephperrott (approved):
LGTM

PR #158, opened 2021-08-16
feat(ng-dev): remove breaking change label configuration from merge command

devversion:
Removes the breaking change label configuration from the
merge command for `ng-dev`. Also uses the constant in the
Github action so everything is in sync.


josephperrott (approved):
LGTM

PR #159, opened 2021-08-16
Create slash commands action

josephperrott:
See individual commits


.github/workflows/dev-infra.yml
```diff
@@ -0,0 +1,15 @@
+name: DevInfra
```
devversion, line 1: should we name this slash commands instead? I imagine if we add more actions part of dev-infra, all could fall into this workflow file, but that is not always possible/desired (if different events are used)

github-actions/slash-commands/commands/rerun-circleci.ts
```diff
@@ -0,0 +1,110 @@
...
+  /** The pull request for the action currently running. */
+  const {data: pullRequest} = await github.pulls.get({
+    ...context.repo,
+    pull_number: context.payload.issue!.number,
+  });
+
+  try {
+    const workflowId = await getCircleCiWorkflowIdForPullRequest(pullRequest, github);
+    const rerunUrl = `https://circleci.com/api/v2/workflow/${workflowId}/rerun?circle-token=${core.getInput(
+      'circleci-token',
```
devversion, line 21: Nit: you could extract the token into a variable for readability.

github-actions/slash-commands/commands/rerun-circleci.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    assertNoErrorsInCircleCiResponse(response);
+  } catch (err) {
+    // If the rerun attempt failed comment on the issue informing the requestor.
+    await github.issues.createComment({
+      ...context.repo,
+      issue_number: pullRequest.number,
+      body:
+        `@${context.actor} the CircleCI rerun you requested failed.  See details below:\n\n` +
+        `\`\`\`\n${err.message}\n\`\`\``,
```
devversion, line 42: could we leak the CircleCI token by accident here? we might want to sanitize the output?

github-actions/slash-commands/commands/rerun-circleci.ts
```diff
@@ -0,0 +1,110 @@
...
+  /** The list of statuses for the latest ref of the PR. */
+  const {statuses} = (
+    await github.repos.getCombinedStatusForRef({
+      owner,
+      repo,
+      ref: pullRequest.head.ref,
+    })
+  ).data;
+  /** The target url of the discovered CircleCI status. */
+  let targetUrl = statuses.find((status) => status.context.startsWith('ci/circleci:'))?.target_url;
```
devversion, line 65: what happens if there are multiple workflows running on PRs?

github-actions/slash-commands/commands/rerun-circleci.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  if (jobIdMatcher === null) {
+    throw Error('Unable to determine the job ID for the CircleCI job creating the status');
+  }
+
+  /** The job ID. */
+  const job = jobIdMatcher[0];
+  /** The full url of the API request to CircleCI. */
+  const url = `https://circleci.com/api/v2/project/gh/${owner}/${repo}/job/${job}?circle-token=${core.getInput(
+    'circleci-token',
```
devversion, line 82: In combination with the Nit above; we could just pass the token as a parameter here then.

github-actions/slash-commands/commands/rerun-circleci.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Checks the provided response from CircleCI's API to determine if it is an error message.
+ *
+ * Properly handled failures in the CircleCI requests are returned with an HTTP response code of 200
+ * and json response with a `:message` key mapping to the failure message.  If `:message` is not
+ * defined, the API request was successful.
+ */
+function assertNoErrorsInCircleCiResponse(response: any) {
+  if (response[':message'] !== undefined) {
```
devversion, line 107: I might miss something, but the CircleCI API spec seems to just have `message` without the colon, right? Also message is always set also for confirmed re-runs. Should we just deal with the status code?

devversion: https://circleci.com/docs/api/v2/#operation/rerunWorkflow

github-actions/slash-commands/main.ts
```diff
@@ -0,0 +1,90 @@
...
+    .split(' ')
+    .filter((word: string) => !!word)
+    .map((word: string) => word.trim());
+
+  if (command[0] !== commandMarker) {
+    throw Error();
+  }
+  parsedCommand = command.slice(1);
+
+  return parsedCommand;
```
devversion, line 27: Why not use a multi-line regex like `^/ng-bot (.*)$` for all of this? Also any strong reason to cache the command here? I think it would be cleaner to just pass it around.

ng-dev/pr/rebase/index.ts
```diff
@@ -42,16 +41,12 @@ const PR_SCHEMA = {
  * Rebase the provided PR onto its merge target branch, and push up the resulting
  * commit to the PRs repository.
  */
-export async function rebasePr(
-  prNumber: number,
-  githubToken: string,
-  config: Pick<NgDevConfig, 'github'> = getConfig(),
-) {
+export async function rebasePr(prNumber: number, githubToken: string) {
```
devversion, line 44:
```suggestion
* @returns a status code indicating whether the rebase was successful or not.
*/
export async function rebasePr(prNumber: number, githubToken: string): Promise<number> {
```

github-actions/slash-commands/main.ts
```diff
@@ -0,0 +1,90 @@
...
+  if (!['created', 'edited'].includes(context.payload.action!)) {
+    core.info('Skipping as this action was not a created or edited issue_comment trigger');
+    return false;
+  }
+  if (!context.payload.comment!.body!.startsWith(commandMarker)) {
+    core.info(`Skipping as the comment does not begin with ${commandMarker}`);
+    return false;
+  }
+  if (getParsedCommand().length === 0) {
+    core.info(`Skipping as only the commandMarket (${commandMarker}) is provided as a command`);
```
devversion, line 56:
```suggestion
    core.info(`Skipping as only the commandMarker (${commandMarker}) is provided as a command`);
```

github-actions/slash-commands/main.ts
```diff
@@ -0,0 +1,90 @@
...
+    return false;
+  }
+  return true;
+}
+
+/** The main function for the slash-command action. */
+async function run(): Promise<void> {
+  if (!shouldPerformAction()) {
```
devversion, line 64: Should we also limit the action to the PR author, and people with write access? That's what I did in my SSR github action: https://github.com/angular/components/pull/23032/files.

Related to that, I was also wondering in terms of security: consider a case where one command will deal with the checked-out PR. This is usually prone to leaking the key. See: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/. Maybe this is already a potential issue for the rebase action right now? (e.g. if I set up some Git hooks; which we might be able to prevent though as a side-note)

There is an action that deals with slash commands in some way already: https://github.com/peter-evans/slash-command-dispatch.

josephperrott: Updated to include a permissions check.

For security, as discussed offline, https://github.com/angular/dev-infra/pull/162 should help us to create a good platform to build upon


The reason I didn't want to use the slash command actions already created is that they are managed and configured entirely in the `.github/<workflow>.yml` file which means that it is extremely difficult for us to centrally manage in any way since its done in the downstream repo.

github-actions/slash-commands/commands/rerun-circleci.ts
josephperrott: I don't believe so, both as the auth headers are not included in outputs

github-actions/slash-commands/commands/rerun-circleci.ts
josephperrott: Added TODO

github-actions/slash-commands/commands/rerun-circleci.ts
josephperrott: While their documentation notes it this way, in my testing/experience it seems that its `:message`

ng-dev/utils/git/authenticated-git-client.ts
```diff
@@ -123,6 +123,25 @@ export class AuthenticatedGitClient extends GitClient {
...
     }
+    if (process.env['GITHUB_ACTIONS']) {
+      throw Error(
+        'Cannot use `configure` static method to create AuthenticatedGitClient in a Github Action.',
+      );
+    }
     AuthenticatedGitClient._authenticatedInstance = new AuthenticatedGitClient(token);
   }
+
+  /** Configures an authenticated git client. */
```
devversion, line 134:
```suggestion
  /** Configures an authenticated git client in the context of a Github action. */
```

ng-dev/utils/git/authenticated-git-client.ts
```diff
@@ -123,6 +123,25 @@ export class AuthenticatedGitClient extends GitClient {
         'Unable to configure `AuthenticatedGitClient` as it has been configured already.',
       );
     }
+    if (process.env['GITHUB_ACTIONS']) {
```
devversion, line 126: I like the `configureForGithubActions` call for now, but I think long-term (or when you feel like doing it right now), I think we should create a wrapper in the shared github action utils that just creates an instance of the `AuthenticatedGitClient`. It feels wrong integrating this special Github action logic in the `ng-dev` command tool.. especially beause right now I would not be able to run a `yarn ng-dev` command in an action right?

devversion (approved):

ng-dev/utils/git/authenticated-git-client.ts
josephperrott: Agreed, we should do something more of a wrapper that properly handles it, I wanted to put this catch in so if we accidentally try to rely on `configure` during a Github action it would tell us it is wrong.

devversion: yeah, that definitely make sense. Might be a good thing to enforce through a linter or through Bazel target visibility

PR #160, opened 2021-08-17
fix(ng-dev): update github release entry when re-tagging major as latest

devversion:
Update the Github release entry `isPrerelease` state when re-tagging
a major as `latest` from `next`.


josephperrott (approved):
LGTM

PR #161, opened 2021-08-17
fix(ng-dev): release fails if release notes are too long

devversion:
While manually testing the release tool a couple of times
for the COMP and FW repos, the script sometimes failed
based on the release notes comparison range. i.e. if
there were too many changes in the release notes, the
Github API errored out when we send the `POST` http
request for creating the Github release entry.

Github does not seem to give official details about
the limits on the release body property. The limit
seems to be between 25k and 125k. This is surprising
but based on actual testing. Initially I was seeing failures
with 25k characters limit, but today (8/17/2021) the limit
seems to be (for all my forks; also organization repos) set
to 125k. I think it's best to stay with the lowest we have
seen. If we exceed, we just link to the changelog, assuming it
still exists in the next branch.

image

### DIFF
--- a/ng-dev/release/publish/actions.ts
+++ b/ng-dev/release/publish/actions.ts
@@ -13,7 +13,11 @@ import * as semver from 'semver';
 
 import {debug, error, green, info, promptConfirm, red, warn, yellow} from '../../utils/console';
 import {AuthenticatedGitClient} from '../../utils/git/authenticated-git-client';
-import {getListCommitsInBranchUrl, getRepositoryGitUrl} from '../../utils/git/github-urls';
+import {
+  getFileContentsUrl,
+  getListCommitsInBranchUrl,
+  getRepositoryGitUrl,
+} from '../../utils/git/github-urls';
 import {createExperimentalSemver} from '../../utils/semver';
 import {BuiltPackage, ReleaseConfig} from '../config/index';
 import {ReleaseNotes} from '../notes/release-notes';
@@ -23,7 +27,12 @@ import {runNpmPublish} from '../versioning/npm-publish';
 
 import {FatalReleaseActionError, UserAbortedReleaseActionError} from './actions-error';
 import {getCommitMessageForRelease, getReleaseNoteCherryPickCommitMessage} from './commit-message';
-import {changelogPath, packageJsonPath, waitForPullRequestInterval} from './constants';
+import {
+  changelogPath,
+  githubReleaseBodyLimit,
+  packageJsonPath,
+  waitForPullRequestInterval,
+} from './constants';
 import {invokeReleaseBuildCommand, invokeYarnInstallCommand} from './external-commands';
 import {findOwnedForksOfRepoQuery} from './graphql-queries';
 import {getPullRequestState} from './pull-request-state';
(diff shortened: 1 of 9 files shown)


ng-dev/release/publish/actions.ts
```diff
@@ -511,16 +523,33 @@ export abstract class ReleaseAction {
...
     info(green(`  ✓   Tagged v${releaseNotes.version} release upstream.`));
 
+    let releaseBody = await releaseNotes.getGithubReleaseEntry();
+
+    // If the release body exceeds the Github body limit, we just provide
+    // a link to the changelog entry in the Github release entry.
+    if (releaseBody.length > githubReleaseBodyLimit) {
+      releaseBody =
+        `Release notes are too large to be captured here. ` +
+        `[View all changes here](${await this._getChangelogUrlForNextBranch(releaseNotes)}).`;
```
devversion, line 533: This assumes that the changelog is always in the next branch. I don't like how this is a requirement for invoking this method (`_createGithubReleaseForVersion`), but I don't see a better way. @josephperrott any thoughts?

josephperrott: Shouldn't we be able to link to the tree with the branch name, since we know the exact tag being used?

`https://github.com/${org}/${repo}/tree/${tag}/CHANGELOG.md` -> `https://github.com/angular/angular/blob/12.2.1/CHANGELOG.md`

ng-dev/release/publish/constants.ts
```diff
@@ -14,3 +14,12 @@ export const changelogPath = 'CHANGELOG.md';
...
 export const waitForPullRequestInterval = 10000;
+
+/**
+ * Maximum number of characters a Github release entry can use for its body.
+ *
+ * Note: Github does not specify an official limit for this, but based on local testing,
+ * Github limits switch between 25000 and 125000 characters. We use the lowest limit we have
+ * seen so far, as otherwise the limit can potentially be wrong and result in errors.
+ */
+export const githubReleaseBodyLimit = 25000;
```
josephperrott, line 25: I will reach out to Github and see if we can find out the official number because I remember personally putting in a request that was way more than 125000 characters successfully in the past during my testing.

ng-dev/release/publish/actions.ts
devversion: I was thinking about it, but was unsure because sometimes the changelog is edited afterwards. Although I'd agree that this is cleaner.

ng-dev/release/publish/constants.ts
clydin: For some additional metrics, the tooling major releases seem to have a count of around 50,000.  But those have been entered manually.  Maybe a fallback to manual entry could be an option if the limit is problematic?

devversion: sounds good. Should we proceed with the 25k for now though? you can see in the screenshot above that it mentions 25k. Today I was seeing 125k limit. I tried multiple scenarios

devversion: @clydin Yeah, that seems possible. thx for additional info. I think the caretaker always would have the option to go back and edit the release manually.

devversion:
@josephperrott Updated. Should be ready for another pass!

ng-dev/release/publish/constants.ts
josephperrott: I tried a bunch of local testing and was unable to get any error message other than 125k characters.  Hopefully Github will get back to me on my request.

josephperrott (approved):
LGTM

ng-dev/release/publish/constants.ts
devversion: yeah, same for me today. Although the screenshot shows that it was 25k a week ago (this showed up when I tested the release tool for the COMP repo)

PR #162, opened 2021-08-17
Create `post` commands to ensure installation tokens are revoked after usage.

josephperrott:
See individual commits.


devversion (approved):
Nice cleanup as well

PR #165, opened 2021-08-18
feat(ng-dev): provide a --depth for fetches during rebase to handle rebases in shallow cloned repos

josephperrott:
If a repo is shallow when the rebase command is used, it is not possible to determine the common merge-base, to address this we fetch the last 500 commits for each ref to increase the likelihood of finding a common ancestor. Going even further would likely be impossible to cleanly rebase anyway.


ng-dev/pr/rebase/index.ts
```diff
@@ -85,13 +85,16 @@ export async function rebasePr(prNumber: number, githubToken: string) {
   }
 
   try {
+    // Fetches are done with --depth=500 increase the likelihood of finding a common ancestor when
```
devversion, line 88: Does using `--unshallow` work as well?

devversion (approved):

ng-dev/pr/rebase/index.ts
josephperrott: That was what I thought too, but I tried it and got: `fatal: --unshallow on a complete repository does not make sense`

devversion: Oh weird. Especially since the `git man` also talks about the complete repository case. good to have this captured here in the PR though. We thought about it 😄

PR #166, opened 2021-08-18
feat(ng-dev): prompt selection of major or minor during branching off in release

### LINKED ISSUE #154: Prompt selection of major or minor during branching off
During any usage of actions based off of `BranchOffNextBranchBaseAction` we should prompt the user to select whether the newly created branch should be a major or minor.

josephperrott:
Fixes #154

### DIFF
--- a/ng-dev/release/publish/actions/branch-off-next-branch.ts
+++ b/ng-dev/release/publish/actions/branch-off-next-branch.ts
@@ -7,6 +7,7 @@
  */
 
 import * as semver from 'semver';
+import {prompt} from 'inquirer';
 
 import {green, info, yellow} from '../../../utils/console';
 import {semverInc} from '../../../utils/semver';
@@ -50,6 +51,7 @@ export abstract class BranchOffNextBranchBaseAction extends ReleaseAction {
     );
     const newVersion = await this._computeNewVersion();
     const newBranch = `${newVersion.major}.${newVersion.minor}.x`;
+    const newNextVersion = await this._promptForNewNextVersion();
 
     // Branch-off the next branch into a new version branch.
     await this._createNewVersionBranchFromNext(newBranch);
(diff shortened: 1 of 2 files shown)


ng-dev/release/publish/actions/branch-off-next-branch.ts
```diff
@@ -134,4 +134,26 @@ export abstract class BranchOffNextBranchBaseAction extends ReleaseAction {
...
+        'branch to be updated',
+    );
+
+    return await prompt<semver.SemVer>({
+      name: 'newNextVersion',
+      message: 'Please select the new version for the next branch',
+      type: 'list',
+      choices: [
+        {name: `Minor (${newMinorVersion.format()})`, value: newMinorVersion},
+        {name: `Major (${newMajorVersion.format()})`, value: newMajorVersion},
```
devversion, line 155: Personally I'm not 100% convinced of prompting when branching-off because switching to major-mode is rather uncommon, compared to switching to minor (which is common), which is why I think it should be only done separate action (obviously it still exists; but I think it should be the only option).

This would match up with the goal of making it really hard to mess things up (asking every time will require the caretaker to think about this and know beforehand what `next` will be. this is what we initially wanted to avoid).

If you feel strongly though, then I think the prompt should rather be a two-step process like:

```
Should the `next` branch target the upcoming new major? Yes
Are you sure? Yes
```

ng-dev/release/publish/test/branch-off-next-branch-testing.ts
```diff
@@ -71,6 +72,11 @@ export async function expectBranchOffActionToRun(
   expectedNewBranch: string,
 ) {
   const action = setupReleaseActionForTesting(actionType, active, isNextPublishedToNpm);
+
+  spyOn(action.instance, '_promptForNewNextVersion' as any).and.resolveTo(
+    semverInc(active.next.version, 'preminor', 'next'),
+  );
```
devversion, line 78: Also, if we go with that prompt, then I think we should somehow manage to get tests for the version computation, and also some tests for the major variant. Right now we only test minor, and we replicate the `preminor` next bump manually here (which is not really good in terms of coverage I think).

PR #169, opened 2021-08-19
feat(ng-dev): support for custom grouping when generating release notes

devversion:
Supports for custom grouping to be configured when generating
release notes. This is useful as the commit scopes vary
based on projects. e.g. CLI uses the full scope for the
package name and expects commits to be grouped based on the
package. In the components repo though, we also include the
secondary entry-point. We expect commits to be still sorted
based on the package though (and not having a section for every
possible entry-point. this will result in large changelogs).


ng-dev/release/publish/test/release-notes/generation.spec.ts
```diff
@@ -38,6 +38,65 @@ describe('release notes generation', () => {
...
+          ### cdk
+          | Commit | Description |
+          | -- | -- |
+          | <..> | fix: commit #2 |
+          | <..> | fix: commit #1 |
+          ### material
+          | Commit | Description |
+          | -- | -- |
+          | <..> | fix: commit #4 |
+          | <..> | fix: commit #3 |
```
devversion, line 94: @josephperrott 2de52df6523fe49d593a7b27e3c0cbe042adbd0d removed the commit scope from commits. This now brings up another issue for COMP if we'd group by package. It would not be clear what scope the `fix` targets. 

Any idea on how we can make this more configurable?

ng-dev/release/notes/release-notes.ts
```diff
@@ -18,6 +18,9 @@ import changelogTemplate from './templates/changelog';
 import githubReleaseTemplate from './templates/github-release';
 import {getCommitsForRangeWithDeduping} from './commits/get-commits-in-range';
 
+/** Default function for retrieving the release note group of a commit. */
+export const defaultCommitToGroupFn = (commit: CommitFromGitLog) => commit.scope;
```
josephperrott, line 22: This appears to be unused correct?

devversion, line 22: Ah, this is a leftover from testing I need to remove

ng-dev/release/publish/test/release-notes/generation.spec.ts
josephperrott: I suppose we would want to have a way to say that if a custom grouping is used, then the scope should be included in the commit description?

```md
### material
| Commit | Description |
| -- | -- |
| <..> | fix(material/dialog): commit #4 |
```

We could make the template aware of if custom grouping is being used and then include the scope if appropriate.

Have a method in `RenderContextData` like this (with a better name):

```ts
getCommitTypeAndOrScope(commit: Commit) {
  if (this.commitToGroup === undefined) {
    return commit.type;
  }
  if (commit.scope === undefined) {
    return commit.type;
  }
  return `${commit.type}(${commit.scope})`
}
```


---

Though I imagine that we actually would want to be able to control what the scope looks like then so that we could have the common part of the group removed if appropriate.

```md
### material
| Commit | Description |
| -- | -- |
| <..> | fix(dialog): commit #4 |
```

---

I am struggling to think of a way to make this type of thing possible without a second property for defining something like this.

But we could extend that method above to use this if provided in the configuration:

```ts
getCommitTypeAndOrScope(commit: Commit) {
  if (this.commitToGroup === undefined) {
    return commit.type;
  }
  if (this.commitScopeTransformer !== undefined) {
    return this.commitScopeTransformer(commit);
  }
  if (commit.scope === undefined) {
    return commit.type;
  }
  return `${commit.type}(${commit.scope})`
}
```

ng-dev/release/notes/context.ts
```diff
@@ -49,6 +50,14 @@ export class RenderContext {
 
   constructor(private readonly data: RenderContextData) {}
 
+  /** Gets the group name for the given commit. */
+  private _commitToGroup(commit: CommitFromGitLog): string {
+    if (this.data.commitToGroup) {
+      return this.data.commitToGroup(commit);
+    }
+    return commit.scope;
+  }
```
josephperrott, line 59: Previously the only usage of `data` was during construction, setting up the values of the class members.  I don't know if it's worth keeping that pattern for this or not.

```suggestion
  private _commitToGroup = this.data.commitToGroup || (commit: CommitFromGitLog) => commit.scope;
```

josephperrott (commented):
**WARNING:** What I am about to layout here is probably not a good idea, but I felt compelled to type it out.  It probably goes to far creating something that isn't actually worth having in place.

---

Would it make more sense to move this `group` value into the commit parsing?

In the commit message we could have something like `groupingFn` which provides the the `type` and `scope` immediately after parsing, then it can return a group it belongs to?

`CommitMessageConfig` would gain 
```ts
{
  groupingFn: (type: string, scope?: string) => string;
}
```

`Commit` and `CommitFromGitLog` would gain
```ts
{
  group: string
}
```

Then in the release notes we could have a boolean option whether to group by `scope` or by `group`.  This could also allow for us long term to have custom commit message validation/requirements that are group specific.

ng-dev/release/notes/context.ts
devversion: yeah, I was thinking about something like this (that's why there is the `defaultCommitToGroupFn` leftover. I think it's more clear in a function though. We use `this.data` in some functions below as well.

ng-dev/release/publish/test/release-notes/generation.spec.ts
devversion: Implemented as discussed offline.

devversion:
@josephperrott Updated based on what we discussed offline. A couple of notes:

* I have considered your idea with grouping in the commit message parsing instead. I didn't feel like going that way because at this point the groups are super-specific to the release notes generation, and implementation-wise it seems bad that we would not be able to pass the whole `CommitFromGitLog` information to the grouping function then.
* I have also considered having two individual options (I did that initially). I decided to go with a combined approach where you can simply "categorize" a commit. This means we only need to do this _once_ per commit (and not multiple times when needed), and it allows me to also split the scope into package name + entry-point only once in the release config.

josephperrott (approved):
LGTM

PR #170, opened 2021-08-20
Refactor: overhaul how configurations are loaded and validate

josephperrott:
See individual commits.


ng-dev/release/publish/test/release-notes/generation.spec.ts
```diff
@@ -28,8 +28,8 @@ describe('release notes generation', () => {
     mkdirSync(testTmpDir);
 
     releaseConfig = {npmPackages: [], buildPackages: async () => []};
-    githubConfig = {owner: 'angular', name: 'dev-infra-test', mainBranchName: 'main'};
-    client = getMockGitClient(githubConfig, /* useSandboxGitClient */ true);
+    GitClientConfig = {owner: 'angular', name: 'dev-infra-test', mainBranchName: 'main'};
```
devversion, line 31: This should not be capital as it's easy to confuse with the type that way. I'm also thinking we should still keep all of this `GithubConfig` because the config is not necessarily only used for the git client + it's scoped under `.github` in the dev-infra config. I can send a follow-up for this so we can move forward with this one.

devversion (approved):
LGTM. Nice cleanup. that is definitely much better.

PR #172, opened 2021-08-20
refactor(ng-dev): rename GitClientConfig to GithubConfig

devversion:
See individual commits


josephperrott (approved):
LGTM

I almost did this in my PR but figured a follow up was a better option, so perfect!

PR #174, opened 2021-08-23
fix(ng-dev) update to the body character limit for github releases to 125000

josephperrott:


devversion:
did you get a response from the Github team, and is there an explanation on why I saw 25k as per the screenshot?

josephperrott:
The response i got 

Hi,

>We typically do not run into this limit, but discovered it during a release attempt. For the purpose of testing (used to create the above response), I created a body with 250000 random numbers, so the length of the body was 250000 characters. The error I receive contains the message: body is too long (maximum is 125000 characters), however another team member during testing saw a message of body is too long (maximum is 25000 characters)

We've found that the value was recently increased from 25,000 to 125,000.

Let us know if you have further questions.

Regards,
< Github Employee Name >

devversion:
haha, what are the odds that they increased the limit during that period. it was 1.5 weeks ago. Anyway, thanks for sharing!

devversion (approved):

PR #175, opened 2021-08-23
fix(ng-dev): remove colon from `Special Thanks` header

alan-agius4:
This is to align with other headers, which don't have a colon at the end.


josephperrott (approved):
LGTM

PR #183, opened 2021-08-26
Replace Ora with a local spinner implementation.

josephperrott:
See individual commits.

Ora has transitioned to using ES modules with their latest release.  For us to rely on it we would also have to switch to ES modules which we are not looking to do at this time.  Additionally, due to our minimal usage of spinners, the Spinner implementation we move to is less complex and smaller in the bundle.


gkalpak:
It would be useful to capture in the commit messages (or here now that it's too late 😛) why we made this change.
Is it to reduce bundle size (since we only needed a small subset of Ora features) or was there another reason?

josephperrott:
Updated the PR description to reflect the reasoning.

PR #188, opened 2021-08-30
build: fix changelog action constructing incorrect version

devversion:
We accidentally used `getDay` instead of `getDate` for
constructing the changelog version.


josephperrott (approved):
LGTM

PR #190, opened 2021-08-31
ci: update to use noreply.github.com email for commit email on change logs

josephperrott:
Update CI to use the correct email for github to recognize the authoring account.

Revert the changelog to allow it to be regenerated.


tools/local-actions/changelog/lib/main.ts
```diff
@@ -46,7 +46,7 @@ async function run(): Promise<void> {
   AuthenticatedGitClient.configure(await getAuthTokenFor(ANGULAR_ROBOT));
   /** The authenticed GitClient. */
   const git = AuthenticatedGitClient.get();
-  git.run(['config', 'user.email', 'angular-robot@google.com']);
+  git.run(['config', 'user.email', '56403804+angular-robot[bot]@users.noreply.github.com']);
```
devversion, line 49: should these details be exposed as a constant in the github action utils? I think it would be okay for the internal action to still rely on that.

devversion (approved):

tools/local-actions/changelog/lib/main.ts
josephperrott: That makes sense, let me do this in a follow up once we confirm we have it working, since testing all of this on Github actions has me unsure if this will work as desired.

devversion: sounds good 👍

PR #194, opened 2021-09-01
refactor: cleanup any casts for octokit API response errors

devversion:
With the new TypeScript 4.4.x update, we had to use `as any`
for a couple of errors. This was necessary because Octokit's
GraphQL implementation did not expose the error constructor
being used for endpoint errors, nor did the error class contain
any useful fields. This has been fixed upstream as the Angular team
has contributed to `@octokit/graphqljs`.

This commit updates to the latest version and cleans up any
of the `as any` usages.


josephperrott (approved):
LGTM

PR #195, opened 2021-09-01
refactor: remove `setPrototypeOf` workaround for custom errors

devversion:
When the shared dev-infra package was still built inside of the
framework repository, we used ES5 for devmode and shipping. This
meant that we needed to use a workaround for making `instanceof`
checks work with ES5-downleveled classes extending `Error`.

In the new repository, devmode and prodmode is ES2020, so we
can stop using the `setPrototypeOf` workaround.


ng-dev/utils/dry-run.ts
```diff
@@ -32,12 +32,4 @@ export function isDryRun(): boolean {
 }
 
 /** Error to be thrown when a function or method is called in dryRun mode and shouldn't be. */
-export class DryRunError extends Error {
-  constructor() {
-    super('Cannot call this function in dryRun mode.');
```
josephperrott, line 37: We need to preserve this super call correct?

josephperrott (approved):
LGTM

Just one comment to address first

PR #196, opened 2021-09-01
feat(bazel): support browsers for m1 platform

devversion:
Work in progress. Blocked on `rules_webtesting` not shipping darwin ARM64 binaries

### DIFF
--- a/bazel/BUILD.bazel
+++ b/bazel/BUILD.bazel
@@ -7,6 +7,7 @@ filegroup(
         "//bazel/api-golden:files",
         "//bazel/benchmark:files",
         "//bazel/browsers:files",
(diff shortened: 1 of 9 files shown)


devversion:
@josephperrott we could already land this in theory. It just updates Chromium, and the m1 binaries are just not usable due to the rules_webtesting issue I mentioned to you

josephperrott:
We can land it now if you would like, or we can wait until we are able to get the `rules_webtesting` changes sorted out first.

devversion:
I'm happy landing this now. Marking as ready for review.

josephperrott (approved):
LGTM

PR #199, opened 2021-09-01
ci: use authenticated git url for pushing updated changelog entries

josephperrott:
See individual commits


.github/workflows/update-changelog.yml
```diff
@@ -11,6 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
```
devversion, line 15: I saw your work on the `main-test` branch, but I think a lot of information/insight you have determined could be shared/incorporate here (here is a question we should answer with a comment IMO)

* Why do we want to _not_ persist credentials? is it because otherwise it might accidentally result in the Github action being used for pushes/etc. instead of our specified https url credentials for the Angular robot?

ng-dev/utils/git/github-urls.ts
```diff
@@ -20,7 +20,8 @@ export const GITHUB_TOKEN_GENERATE_URL = 'https://github.com/settings/tokens/new
 /** Adds the provided token to the given Github HTTPs remote url. */
 export function addTokenToGitHttpsUrl(githubHttpsUrl: string, token: string) {
   const url = new URL(githubHttpsUrl);
-  url.username = token;
+  url.password = token;
+  url.username = '_';
```
devversion, line 24: Out of curiosity: how did our merge tooling etc. work without this before? I assume (but you know better with your testing in `main-test`) because in those cases we deal with a repo that is actively used and where a user is already set in the repo? and for actions we now this need because we have set `persist-credentials: false`, so there is no actual user stored in credentials (like it would be local for the caretaker running yarn ng-dev pr merge)?

I think we should try to capture this here in the PR, or best in a comment..

tools/local-actions/changelog/lib/main.ts
```diff
@@ -46,9 +47,11 @@ async function run(): Promise<void> {
   AuthenticatedGitClient.configure(await getAuthTokenFor(ANGULAR_ROBOT));
   /** The authenticed GitClient. */
   const git = AuthenticatedGitClient.get();
-  git.run(['config', 'user.email', '56403804+angular-robot[bot]@users.noreply.github.com']);
+  git.run(['config', 'user.email', 'angular-robot@google.com']);
   git.run(['config', 'user.name', 'Angular Robot']);
 
+  /** The url of the git repository with the authenticated token set. */
+  const authenticatedGitUrl = addTokenToGitHttpsUrl(git.getRepoGitUrl(), git.githubToken);
```
devversion, line 54: The `git.getRepoGitUrl` method from the authenticated client already incorporates the token? is that really needed?

josephperrott: Correct, I don't know why when I was testing before that it didn't include it.

ng-dev/utils/git/github-urls.ts
josephperrott: Agreed we should document this somewhere.  I am not sure where the best place to include it is though because I think a comment is the best option, but I don't know where that comment would even go.

The answer to why it worked before is that everyone running the commands were always running things from an environment that already had authentication set up, so if the authentication didn't work via the `remote` url, it just used the git cached credentials.

devversion: I think this comment is also sufficient for future reasoning

josephperrott: Added that block into the commit message as well.

devversion (approved):
LGTM, only the changelog action main needs to be re-updated. 

```
//tools/local-actions/changelog:main                                     FAILED in 2.2s
```

PR #201, opened 2021-09-02
refactor(ng-dev): release config buildPackages should always stamp

devversion:
See individual commits.


josephperrott (approved):
LGTM

I think it makes sense to do this refactor and if/once we create common tooling for publishing our snapshots on CI we can refactor things based on what we need to do then.

devversion:
yeah, that sounds reasonable 👍

PR #202, opened 2021-09-02
refactor(ng-dev): create enum for target label names

devversion:
Instead of hard-coding the label names for the target labels
the whole Angular organization uses, we now have an enum for
these names, allowing for consistent use of the labels.

This is in preparation for adding the proper target labels
automatically to staging and cherry-pick release PRs. Previously
this was not possible because we still had a possible way to configure
custom target labels, and missed such an enum.

This new enum came together with some refactorings that cleaned up
some of the validation and target label extraction logic. Mostly code
has been split up into individual files, allowing for more readable code
and less convuluted files.

**Note**: This also fixes `yarn ng-dev pr check-target-branches` if target labels are disabled.

### DIFF
--- a/ng-dev/pr/merge/fetch-pull-request.ts
+++ b/ng-dev/pr/merge/fetch-pull-request.ts
@@ -1,0 +1,60 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {params, types as graphqlTypes} from 'typed-graphqlify';
+import {AuthenticatedGitClient} from '../../utils/git/authenticated-git-client';
+import {getPr} from '../../utils/github';
+
+/** Graphql schema for the response body the requested pull request. */
+const PR_SCHEMA = {
+  url: graphqlTypes.string,
+  isDraft: graphqlTypes.boolean,
+  state: graphqlTypes.oneOf(['OPEN', 'MERGED', 'CLOSED'] as const),
+  number: graphqlTypes.number,
+  // Only the last 100 commits from a pull request are obtained as we likely will never see a pull
+  // requests with more than 100 commits.
+  commits: params(
+    {last: 100},
+    {
+      totalCount: graphqlTypes.number,
+      nodes: [
+        {
+          commit: {
+            status: {
+              state: graphqlTypes.oneOf(['FAILURE', 'PENDING', 'SUCCESS'] as const),
+            },
+            message: graphqlTypes.string,
+          },
+        },
+      ],
+    },
+  ),
+  baseRefName: graphqlTypes.string,
+  title: graphqlTypes.string,
+  labels: params(
+    {first: 100},
+    {
+      nodes: [
+        {
+          name: graphqlTypes.string,
+        },
+      ],
+    },
+  ),
+};
+
+/** A pull request retrieved from github via the graphql API. */
+export type RawPullRequest = typeof PR_SCHEMA;
+
+/** Fetches a pull request from Github. Returns null if an error occurred. */
+export async function fetchPullRequestFromGithub(
+  git: AuthenticatedGitClient,
+  prNumber: number,
+): Promise<RawPullRequest | null> {
+  return await getPr(PR_SCHEMA, prNumber, git);
+}
--- a/ng-dev/pr/merge/target-label.ts
+++ b/ng-dev/pr/merge/target-label.ts
@@ -7,23 +7,34 @@
  */
 
 import {MergeConfig} from './config';
-import {matchesPattern} from './string-pattern';
 import {getTargetLabelsForActiveReleaseTrains} from './defaults';
 import {GithubConfig} from '../../utils/config';
 import {Commit} from '../../commit-message/parse';
 import {assertChangesAllowForTargetLabel} from './validations';
 import {PullRequestFailure} from './failures';
 
-/** Describes possible values that can be returned for `branches` of a target label. */
-export type TargetLabelBranchResult = string[] | Promise<string[]>;
+/**
+ * Enum capturing available target label names in the Angular organization. A target
+ * label is set on a pull request to specify where its changes should land.
+ *
+ * More details can be found here:
+ * https://docs.google.com/document/d/197kVillDwx-RZtSVOBtPb4BBIAw0E9RT3q3v6DZkykU#heading=h.lkuypj38h15d
+ */
+export enum TargetLabelName {
+  MAJOR = 'target: major',
+  MINOR = 'target: minor',
+  PATCH = 'target: patch',
+  RELEASE_CANDIDATE = 'target: rc',
+  LONG_TERM_SUPPORT = 'target: lts',
+}
 
 /**
- * Matcher that can resolve a Github label to a list of branches into which a pull
- * request should be merged into..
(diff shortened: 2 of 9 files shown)


devversion:
@josephperrott I'm on M1 right now, so cannot run the formatter, but can tomorrow. This is a good chance asking for your thoughts on another release of buildifier. Their master branch adds support for M1, but we miss a release 😄

ng-dev/pr/merge/fetch-pull-request.ts
```diff
@@ -0,0 +1,60 @@
...
+      ],
+    },
+  ),
+};
+
+/** A pull request retrieved from github via the graphql API. */
+export type RawPullRequest = typeof PR_SCHEMA;
+
+/** Fetches a pull request from Github. Returns null if an error occurred. */
+export async function fetchPullRequestFromGithub(
```
josephperrott, line 55: I know that this is just a refactoring of the function into a separate file, but looking at it now, I think we should have this file just export the schema for the GraphQL query and then we can just directly call `await getPr(PR_SCHEMA, prNumber, git)` in the `ng-dev/pr/merge/pull-request.ts` file.  `getPr` is already set up to use the type as the outputted type.

ng-dev/pr/merge/target-label.ts
```diff
@@ -6,9 +6,34 @@
...
+/** Describes possible values that can be returned for `branches` of a target label. */
+export type TargetLabelBranchResult = string[] | Promise<string[]>;
+
+/**
+ * Matcher that can resolve a Github label to a list of branches into which a pull
+ * request should be merged into..
+ */
+export interface TargetLabel {
+  /** Pattern that matches the given target label. */
+  pattern: RegExp | string;
```
josephperrott, line 26: This can no longer be a `RegExp` with our new setup, we know that it will be `string` correct?

ng-dev/pr/merge/target-label.ts
```diff
@@ -6,9 +6,34 @@
...
+import {Commit} from '../../commit-message/parse';
+import {assertChangesAllowForTargetLabel} from './validations';
+import {PullRequestFailure} from './failures';
+
+/** Describes possible values that can be returned for `branches` of a target label. */
+export type TargetLabelBranchResult = string[] | Promise<string[]>;
+
+/**
+ * Matcher that can resolve a Github label to a list of branches into which a pull
+ * request should be merged into..
```
josephperrott, line 22:
```suggestion
 * request should be merged into.
```

ng-dev/pr/merge/target-label.ts
```diff
@@ -6,9 +6,34 @@
...
+  pattern: RegExp | string;
+  /**
+   * List of branches a pull request with this target label should be merged into.
+   * Can also be wrapped in a function that accepts the target branch specified in the
+   * Github Web UI. This is useful for supporting labels like `target: development-branch`.
+   *
+   * @throws {InvalidTargetLabelError} Invalid label has been applied to pull request.
+   * @throws {InvalidTargetBranchError} Invalid Github target branch has been selected.
+   */
+  branches: TargetLabelBranchResult | ((githubTargetBranch: string) => TargetLabelBranchResult);
```
josephperrott, line 35: To make this simpler should we just change this to

```suggestion
  branches: (githubTargetBranch: string) => Promise<TargetLabelBranchResult>;
```

Since its no longer user configuration that defines `TargetLabel`. Currently only one of them is a non-function, and we always `await` the `branches` value anyway.  We could then drop the `getBranchesFromTargetLabel` function as well.

ng-dev/pr/merge/target-label.ts
```diff
@@ -27,19 +52,19 @@ export class InvalidTargetLabelError {
 }
 
 /** Gets the target label from the specified pull request labels. */
-export async function getTargetLabelFromPullRequest(
+export async function getMatchingTargetLabelForPullRequest(
   config: Pick<MergeConfig, 'noTargetLabeling'>,
-  labels: string[],
+  labelsOnPullRequest: string[],
+  allTargetLabels: TargetLabel[],
```
josephperrott, line 58: Why not just have the `getTargetLabelsForActiveReleaseTrains()` call occur within this function?  If we are worried about calling it multiple times we could just memoize the results in some way.

josephperrott (commented):
Overall I think its good, just a few ideas.

An idea after seeing how these pieces work together now, maybe in a different follow up PR. I think we should disentangle pull request merging from pull request validation.  I will write up an issue for it to have the discussion there.

ng-dev/pr/merge/target-label.ts
devversion: This is done to make the function more self-contained and less concerned with other specifics of the merge tooling. Also it allows us to conveniently use this for testing, where we previously retrieved all target labels for active release trains, and then patched the same method with the result

josephperrott: Okay, that works for me

ng-dev/pr/merge/fetch-pull-request.ts
devversion: Yeah, I think having the wrapper function allows us to nicely isolate the pull request fetching into this file, while otherwise we'd manually need to feed it the schema. I guess it doesn't matter too much.

devversion: I'd like to keep it just for the sake of isolating everything into the `fetch-pull-request.ts` file here. we can revisit if you feel strongly.

ng-dev/pr/merge/target-label.ts
devversion: Yeah, we should probably change this to a simple `string` and `name`.

ng-dev/pr/merge/fetch-pull-request.ts
josephperrott: We can keep it there, it's something we could also look at during the larger refactor outlined in #203 to see if it makes sense when all of that settles.

ng-dev/pr/merge/target-label.ts
devversion: 👍

devversion:
@josephperrott Addressed most of the feedback. thanks. Please have another look

ng-dev/pr/merge/fetch-pull-request.ts
devversion: Agreed 👍

ng-dev/pr/merge/target-label.ts
```diff
@@ -27,19 +63,19 @@ export class InvalidTargetLabelError {
...
 
-  const targetLabels = await getTargetLabels();
-  /** List of discovered target labels for the PR. */
-  const matches = [];
-  for (const label of labels) {
-    const match = targetLabels.find(({pattern}) => matchesPattern(label, pattern));
+  const matches: TargetLabel[] = [];
+
+  for (const label of labelsOnPullRequest) {
+    const match = allTargetLabels.find(({name}) => label === name);
```
josephperrott, line 78:
```suggestion
   if (allTargetLabels.includes(label)) {
```

josephperrott (approved):
LGTM

Just one little nit as a proposed change

ng-dev/pr/merge/target-label.ts
devversion: If I don't miss anything, I think we need a reference to the marched target label.

josephperrott: Whoops, your right.

PR #204, opened 2021-09-04
Run formatter for CHANGELOG.md when generated during release

### LINKED ISSUE #197: Code formatting should be run on CHANGELOG.md during release
During the release process, when the new Changelog entry is generated and added to the CHANGELOG.md file, we should leverage the `formatFiles` function from `ng-dev/format/format` to format the newly modified CHANGELOG.md file.

This notably presents as a failure on CI for CLI as markdown files are formatted using prettier.

josephperrott:
See individual commits as this change contains a few refactors to make this easier to reason about.

Closes #197


ng-dev/format/format.ts
```diff
@@ -19,7 +19,7 @@ export async function formatFiles(files: string[]) {
```
devversion, line 19: I can't comment on the `formatFiles` function signature, but we should add an explicit return type so we never can accidentally return without any value/or a different value. + a `@returns` JSDoc would be nice to have.

ng-dev/release/notes/cli.ts
```diff
@@ -48,33 +45,35 @@ function builder(argv: Argv): Argv<ReleaseNotesOptions> {
       choices: ['github-release', 'changelog'] as const,
       default: 'changelog' as const,
     })
-    .option('outFile', {
-      type: 'string',
-      description: 'File location to write the generated release notes to',
-      coerce: (filePath?: string) => (filePath ? join(process.cwd(), filePath) : undefined),
+    .option('updateChangelog', {
```
devversion, line 48: I think `prependToChangelog` would be more accurate and less ambiguous here.

ng-dev/release/notes/release-notes.ts
```diff
@@ -10,33 +10,43 @@ import * as semver from 'semver';
...
   private renderContext: RenderContext | undefined;
   /** The title to use for the release. */
   private title: string | false | undefined;
-  /** The configuration for release notes. */
-  private config: ReleaseNotesConfig = this.getReleaseConfig().releaseNotes ?? {};
+  /** The configuration ng-dev. */
+  private config: {release: ReleaseConfig} = getConfig([assertValidReleaseConfig]);
+  /** The configuration for the release notes. */
+  private get notesConfig() {
+    return this.config.release.releaseNotes || {};
```
devversion, line 42:
```suggestion
    return this.config.release.releaseNotes ?? {};
```

ng-dev/release/notes/release-notes.ts
```diff
@@ -50,6 +60,28 @@ export class ReleaseNotes {
     return render(changelogTemplate, await this.generateRenderContext(), {rmWhitespace: true});
   }
 
+  /** Prepends the generated release note to the CHANGELOG file. */
+  async prependEntryToChangelog() {
+    /** The fully path to the changelog file. */
+    const filePath = join(this.git.baseDir, 'CHANGELOG.md');
```
devversion, line 66: We should have a constant for the `CHANGELOG.md` file, like we do for the release action code. Ideally this constant should be in `release/versioning` as a common place to capture version-specific things (maybe there already is a constant for this; I don't recall)

ng-dev/release/notes/release-notes.ts
```diff
@@ -50,6 +60,28 @@ export class ReleaseNotes {
     return render(changelogTemplate, await this.generateRenderContext(), {rmWhitespace: true});
   }
 
+  /** Prepends the generated release note to the CHANGELOG file. */
```
devversion, line 63: Nit: For good code quality I'd  personally recommend we also say that the changelog file is updated in the repository directory as provided by the Git client (part of the JSDoc description).

It's not clear where the file is updated if I just look at `ReleaseNotes#prependEntryToChangelog` from outside.

ng-dev/release/notes/release-notes.ts
```diff
@@ -50,6 +60,28 @@ export class ReleaseNotes {
...
+  async prependEntryToChangelog() {
+    /** The fully path to the changelog file. */
```
devversion, line 65:
```suggestion
    /** The absolute path to the changelog file. */
```

ng-dev/release/notes/release-notes.ts
```diff
@@ -50,6 +60,28 @@ export class ReleaseNotes {
...
+      changelog = readFileSync(filePath, {encoding: 'utf8'});
+    }
+    /** The new changelog entry to add to the changelog. */
+    const entry = await this.getChangelogEntry();
+
+    writeFileSync(filePath, `${entry}\n\n${changelog}`);
+
+    try {
+      assertValidFormatConfig(this.config);
+      await formatFiles([filePath]);
```
devversion, line 79: In what cases do we need to run a formatter? Are we emitting potentially-broken markdown here? I'm asking because this would set a precedent for running the formatter in other places too. e.g. when updating the `package.json`.

it also seems like a bad place to do this inside the `ReleaseNotes` class? making the function perform more than it suggests (as per my interpretation). Stuff is getting more convoluted that way. Should formatter instead be run when we actually prepend with some comment on why we run the formatter afterwards (e.g. when we actually create the commit)

ng-dev/release/publish/test/release-notes/generation.spec.ts
```diff
@@ -28,12 +31,10 @@ describe('release notes generation', () => {
 
     releaseConfig = {npmPackages: [], buildPackages: async () => []};
     githubConfig = {owner: 'angular', name: 'dev-infra-test', mainBranchName: 'main'};
+    setConfig({github: githubConfig, release: releaseConfig});
```
devversion, line 34: `setConfig` persists and leaks between tests. I think a _spy_ is the best solution here as it gets cleared automatically on test completion.

ng-dev/release/publish/test/release-notes/generation.spec.ts
```diff
@@ -436,4 +437,37 @@ describe('release notes generation', () => {
...
+      const changelog = readFileSync(`${testTmpDir}/CHANGELOG.md`, 'utf8');
+
+      expect(changelog).toBe(
+        dedent`
+      <a name="13.0.0"></a>
+      # 13.0.0 (${buildDateStamp()})
+      ### ng-dev
+      | Commit | Type | Description |
+      | -- | -- | -- |
+      | [${shortSha}](https://github.com/angular/dev-infra-test/commit/${fullSha}) | fix | commit *1 |
```
devversion, line 464: Seems like this should be tested in a dedicated spec. This should just confirm the prepending.

We don't have such a test yet; so that is great! but we should have a dedicated test instead IMO.

ng-dev/release/publish/test/test-utils/action-mocks.ts
```diff
@@ -69,6 +65,9 @@ export function setupMocksForReleaseAction<T extends boolean>(
   // to persist between tests if the sandbox git client is used.
   prepareTempDirectory();
 
+  // Set the configuration to be used throughout the spec.
+  setConfig({github: githubConfig, release: releaseConfig});
```
devversion, line 69: Same comment on `setConfig` as above (where the Spy is actually better as it does not leak)

ng-dev/release/publish/test/test-utils/sandbox-testing.ts
```diff
@@ -78,14 +78,23 @@ export class SandboxGitRepo {
...
+    return this;
+  }
+
+  /** Retrieve the sha for the commit. */
+  getShaForCommitId(commitId: number, type: 'long' | 'short' = 'long'): string {
     const commitSha = this._commitShaById.get(commitId);
 
     if (commitSha === undefined) {
-      throw Error('Unable to cherry-pick. Unknown commit id.');
+      throw Error('Unable to get determine SHA due to an unknown commit id.');
```
devversion, line 91:
```suggestion
      throw Error('Unable to retrieve SHA due to an unknown commit id.');
```

ng-dev/utils/testing/bazel-env.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+import {join} from 'path';
+
+/**
+ * Temporary directory which will be used as project directory in tests. Note that this environment
+ * variable is automatically set by Bazel for tests. Bazel expects tests "not attempt to remove,
+ * chmod, or otherwise alter [TEST_TMPDIR]," so a subdirectory path is used to be created/destroyed.
+ */
+export const testTmpDir: string = join(process.env['TEST_TMPDIR']!, 'dev-infra');
```
devversion, line 16: Unfortunately this is prone to errors as the `dev-infra` folder inside the Bazel test tmpdir does not necessarily exist (resulting in the issues I had when I ran tests on my fresh Bazel output base).

We should expose a method for ensuring it exists, but I foresee cases where we accidentally use `testTmpDir` without ensuring it exists. Maybe it should be only possible to retrieve the test tmp dir through a function like `prepareTestTmpDir()` where it returns the path then.

ng-dev/utils/assertion-typings.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  intersection: infer I,
+) => void
+  ? I
+  : never;
+
+type AssertedType<A> = A extends AssertionFn<infer T> ? T : never;
+type AllAssertedTypes<A extends readonly AssertionFn<unknown>[]> = {
+  [K in keyof A]: AssertedType<A[K]>;
+};
+type TupleToUnion<T> = T[keyof T & number];
```
devversion, line 11: It does not seem like this deals with a tuple (in combination with `AllAssertedTypes`). I think we should name it: `ExtractValuesAsUnionFromObject`. what do you think? `AllAssertedTypes` returns an object type like this IIUC:

```ts
type = {
   0: {github: GithubConfig},
   1: {release: ReleaseConfig},
}
```

devversion (commented):
LGTM. Nice work on the assertion stuff. That is pretty cool!

A couple of comments on logic organization & the testing bits.

ng-dev/utils/assertion-typings.d.ts
josephperrott: Yeah, we can do that, and correct I believe that is what the type looks like from `AllAssertedTypes`

ng-dev/release/notes/release-notes.ts
josephperrott: I moved the `CHANGELOG.md` constant into `ng-dev/release/notes`

ng-dev/release/notes/release-notes.ts
josephperrott: Yeah, I can switch it to do that instead.

josephperrott: I think the piece I am unsure of if is whether formatting should be blocking, or if formatting fails we should just notify and continue on in the process.

devversion: I think the question would be: do you we actually need to run the formatter for the MD we generate at all? 

It's a tough decision whether this is blocking or not. it kind of depends on whether we enforce formatting or not right?

josephperrott: Because we don't have a standard answer at the current moment for how MD should be formatted, my thought is just to run the formatter on all of the changes to ensure it meets whatever the criteria is for the repository we are in.

Long term we should have the same formatting requirements, and then we can just have this formatting check be part of the testing of the generation of entries.

devversion: That sounds reasonable. I'd make it non-blocking given your points above. Can we add a TODO?

ng-dev/utils/testing/virtual-git-client.ts
```diff
@@ -74,6 +69,11 @@ export class VirtualGitClient extends AuthenticatedGitClient {
   /** List of pushed heads to a given remote ref. */
   pushed: {remote: RemoteRef; head: GitHead}[] = [];
 
+  constructor(githubToken: string, baseDir?: string, config?: {github: GithubConfig}) {
+    super(githubToken, baseDir, config);
+    console.log(baseDir);
+  }
```
devversion, line 75: Console leftover / can the constructor be removed completely?

tools/jasmine/bootstrap.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {readFileSync, writeFileSync} from 'fs';
+import {join} from 'path';
+import {testTmpDir} from '../../ng-dev/utils/testing/bazel-env';
+
+describe('bootstraping script', () => {
+  // This specific describe block is explicitly testing that
```
devversion, line 14: Is the comment unfinished?

tools/jasmine/bootstrap.ts
```diff
@@ -0,0 +1,25 @@
...
+ *  The `boot` utility function from `@bazel/jasmine` allows initialization before the runner, with
+ *  protections to prevent calling `boot` again.
+ */
+const {boot} = require('@bazel/jasmine');
+
+// Initialize jasmine, because beforeEach and other jasmine "globals" are not defined in the
+// environment until jasmine has been initialized.  Since our bootstrap explicitly is meant to use
+// these globals, we need to ensure that jasmine has already initialized before the major content
+// of our script executes.
+boot();
```
devversion, line 20: What happens if boot is called a second time? just checking to make sure we don't run tests twice.

devversion: Asking because I think, to really test that it only runs once, we'd need to move the testCount constant out of the closure to the file toplevel, right?

tools/jasmine/jasmine.bzl
```diff
@@ -0,0 +1,21 @@
...
+    ]
+
+    _jasmine_node_test(
+        name = name,
+        srcs = kwargs.pop("srcs", []) + specs,
+        use_direct_specs = True,
+        deps = kwargs.pop("deps", []) + [
+            "//tools/jasmine:bootstrap",
+            # Because we don't provided a bundled script, we must ensure that the dependencies for
+            # bootstrap are also included as dependencies.
```
devversion, line 16: Aren't the transitive deps of the bootstrap target automatically unwrapped?

you could simplify this by just depending on the TS target here while still having the other filegroup for the ES5 file above in the templated args (named bootstrap_init_file or something)

josephperrott: I switched to doing so, though I don't love that we have to still have both dependencies listed here.

tools/jasmine/bootstrap.ts
josephperrott: I can't think of how it would get called again, but I added in an `undefined` check to be sure.

devversion (approved):
Nice! this turned out pretty large but it's a great refactoring with the `beforeEach` jasmine logic.

PR #207, opened 2021-09-08
fix(ng-dev/release): fetch release notes compare tag and store it locally

devversion:
The release notes compare tag might not exist locally if the local
repository has not been synced with upstream. We fetch the release
notes compare tag in order to be able to build the changelog. Our
current fetch logic incorrectly stores the tag reference in
`FETCH_HEAD`, while we want it to be stored locally. This commit
fixes that.

### DIFF
--- a/ng-dev/release/publish/actions.ts
+++ b/ng-dev/release/publish/actions.ts
@@ -415,7 +415,14 @@ export abstract class ReleaseAction {
     const releaseNotesCompareTag = getReleaseTagForVersion(compareVersionForReleaseNotes);
 
     // Fetch the compare tag so that commits for the release notes can be determined.
-    this.git.run(['fetch', this.git.getRepoGitUrl(), `refs/tags/${releaseNotesCompareTag}`]);
+    // We forcibly override existing local tags that are named similar as we will fetch
+    // the correct tag for release notes comparison from the upstream remote.
+    this.git.run([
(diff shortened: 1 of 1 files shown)


ng-dev/release/publish/actions.ts
```diff
@@ -415,7 +415,10 @@ export abstract class ReleaseAction {
     const releaseNotesCompareTag = getReleaseTagForVersion(compareVersionForReleaseNotes);
 
     // Fetch the compare tag so that commits for the release notes can be determined.
-    this.git.run(['fetch', this.git.getRepoGitUrl(), `refs/tags/${releaseNotesCompareTag}`]);
+    // We forcibly override existing local tags that are named similar as we will fetch
+    // the correct tag for release notes comparison from the upstream remote.
+    this.git.run(['fetch', '--force', this.git.getRepoGitUrl(),
+        `refs/tags/${releaseNotesCompareTag}:refs/tags/${releaseNotesCompareTag}`]);
```
devversion, line 421: Note that `tag <name>` would also work as refspec but I think it's better being explicit here. Quote from Git docs:

> tag <tag> means the same as refs/tags/<tag>:refs/tags/<tag>; it requests fetching everything up to the given tag.

josephperrott (approved):
LGTM

PR #208, opened 2021-09-09
test: move test count assertion to file-level check

devversion:
Compared to keeping track of the amount of tests being
run in the describe block, we move the variable up to
the file top-level. This makes the check, ensuring tests
do not execute twice, more reliable as in theory the
`describe` closure can be called twice where the count
would be reset then (resulting in false-positive tests)


josephperrott (approved):
LGTM

PR #210, opened 2021-09-09
build: automatically set the action: merge label for renovate changes

josephperrott:
Since any change renovate makes is intended to be merged as soon as it passes, it
should have the `action: merge` label as the PR author is ready for its merge.


alan-agius4 (approved):

josephperrott:
@alan-agius4 Since we are still using our `ng-dev` merge tooling in this repo, we are not leveraging the automatic merging at this point.

PR #212, opened 2021-09-09
feat(ng-dev/release): add marker between generated changelog entries

### LINKED ISSUE #198: Add separators between generated entries to ease being manipulating changelog files
Add a separator between changelog entries in CHANGELOG.md files to allow us to easily do things like archiving or filtering when updating changelogs.

Proposal:
```md
<a name="1.2.3"></a>
# 1.2.3 (2000-01-01)
### ng-dev
| Commit | Type | Description |
| -- | -- | -- |
| [abc123](https://github.com/angular/dev-infra/commit/abc123) | feat | **release:** Add changlog separators |

<!-- CHANGELOG SPLIT MARKER -->

<a name="1.2.2"></a>
# 1.2.2 (2000-01-01)
```

josephperrott:
Adding a marker `<!-- CHANGELOG SPLIT MARKER -->` between changelog entries allows for easier
automated modification of the changelog file, for things like removing prerelease changelog entries
when publishing a new major, or automatically moving old changelog entries to an archive changelog.

Fixes #198


ng-dev/release/notes/changelog.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  describe('adds entries to the changelog', () => {
+    it('creating a new changelog file if one does not exist.', () => {
+      expect(existsSync(changelog.changelogPath)).toBe(false);
+
+      changelog.prependEntryToChangelog(createChangelogEntry('0.0.0'));
+      expect(existsSync(changelog.changelogPath)).toBe(true);
+    });
+
+    it('not including a split marker when only one changelog entry is in the changelog.', () => {
```
devversion, line 34: Just a small nit: I think jasmine tests should read as they appear in the code. e.g. `it should not include a split marker ..`. The current test description is not clear about the expected outcome.

ng-dev/release/notes/changelog.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  describe('adds entries to the changelog archive', () => {
+    it('only updating or creating the changelog archive if necessary', () => {
+      changelog.prependEntryToChangelog(createChangelogEntry('1.0.0'));
+      expect(existsSync(changelog.changelogArchivePath)).toBe(false);
+
+      changelog.moveEntriesPriorToVersionToArchive(new SemVer('1.0.0'));
+      expect(existsSync(changelog.changelogArchivePath)).toBe(false);
+
+      changelog.moveEntriesPriorToVersionToArchive(new SemVer('2.0.0'));
+      expect(existsSync(changelog.changelogArchivePath)).toBe(true);
```
devversion, line 58: Should we actually test how the split markers are inserted in the `describe` above, and check here that the archive is generated as expected?

IMO It's good to actually assert these things IMO to have full coverage (like we do with the release notes generation specs I have created)

devversion: (can be in a follow-up as well; but I think we should not forget it)

ng-dev/release/notes/release-notes.ts
```diff
@@ -70,23 +69,17 @@ export class ReleaseNotes {
...
+    changelog.prependEntryToChangelog(await this.getChangelogEntry());
 
     // TODO(josephperrott): Remove file formatting calls.
     //   Upon reaching a standardized formatting for markdown files, rather than calling a formatter
     //   for all creation of changelogs, we instead will confirm in our testing that the new changes
     //   created for changelogs meet on standardized markdown formats via unit testing.
     try {
       assertValidFormatConfig(this.config);
-      await formatFiles([this.changelogPath]);
+      await formatFiles([this.changelog.changelogPath]);
```
devversion, line 82: Optional idea: How about changing the constant to `this.changelog.filePath` and `this.changelog.archiveFilePath` 

(to avoid the repetition; the context of changelog is already clear within the class)

ng-dev/release/notes/release-notes.ts
```diff
@@ -70,23 +69,17 @@ export class ReleaseNotes {
...
    */
   async prependEntryToChangelog() {
-    /** The changelog contents in the current changelog. */
-    let changelog = '';
-    if (existsSync(this.changelogPath)) {
-      changelog = readFileSync(this.changelogPath, {encoding: 'utf8'});
-    }
-    /** The new changelog entry to add to the changelog. */
-    const entry = await this.getChangelogEntry();
+    const changelog = new Changelog(this.git);
```
devversion, line 72: Seems like this constant is not needed? there is already `this.changelog`

ng-dev/release/notes/changelog.ts
```diff
@@ -0,0 +1,127 @@
...
+export class Changelog {
+  /** The absolute path to the changelog file. */
+  readonly changelogPath = join(this.git.baseDir, changelogPath);
+  /** The absolute path to the changelog archive file. */
+  readonly changelogArchivePath = join(this.git.baseDir, changelogArchivePath);
+
+  /** The changelog entries in the CHANGELOG.md file. */
+  private entries = this.getEntriesFor(this.changelogPath);
+  /** The changelog entries in the CHANGELOG_ARCHIVE.md file. */
+  private archiveEntries = this.getEntriesFor(this.changelogArchivePath);
```
devversion, line 48: IIRC there were some problems with reading _such_ large files in a Node process. I think we should only read files when actually needed?

ng-dev/release/notes/changelog.ts
```diff
@@ -0,0 +1,127 @@
...
+  prependEntryToChangelog(entry: string) {
+    this.entries.unshift(parseChangelogEntry(entry));
+    this.writeToChangelogFiles();
+  }
+
+  /**
+   * Move all changelog entries from the CHANGELOG.md file for versions prior to the provided
+   * version to the changelog archive.
+   */
+  moveEntriesPriorToVersionToArchive(version: semver.SemVer) {
```
devversion, line 64: Can you elaborate on why this concept is used, compared to just limiting the entries to a certain amount (like you did with the changelog github action). That way we can limit entries in the changelog to a fixed number and not have to worry about version-specific details.

ng-dev/release/notes/changelog.ts
```diff
@@ -0,0 +1,127 @@
...
+ * <changelog entry content>
+ *
+ * <!-- CHANGELOG SPLIT MARKER -->
+ *
+ * <changelog entry content>
+ */
+const joinMarker = `\n\n${splitMarker}\n\n`;
+
+/** A RegExp matcher to extract the version of a changelog entry from the entry content. */
+const versionAnchorMatcher = new RegExp(`<a name="(.*)"></a>`);
```
devversion, line 31: I'm a bit unhappy about the reliance on some template-specifics here. It will become problematic if we make changes of the anchor. e.g. even when we switch from `<a name` to `<a id` (which is actually better for fragment resolution in w3c).

If you see my comment on `moveEntriesPriorToVersionToArchive`, an option is that we might not need the version at all, or alternatively we could incorporate the version in a specific comment that is generated here as part of the `Changelog` class (that keeps the logic more local and self-contained)

devversion (commented):
Nice! just a few comments/questions (especially because this logic will be core to all Angular repos)

ng-dev/release/notes/changelog.ts
josephperrott: I can put this in a comment, but it made the most sense to me for us to migrate all of the entries from versions after something like the LTS period ends, or upon its entry into the LTS version.

For dev-infra it made sense to do it based on the number of entries because we have roughly one entry per week.  The entries in our normal changelogs are representative of actual versioning, so the context for whether something matters in the changelog is actually the version rather than the date it was created.

ng-dev/release/notes/changelog.ts
josephperrott: Should we include a metadata comment object?  Something like.

```md
<!--ReleaseMetadata {version: '1.2.3', name: 'release name'} -->
```

Then we can retrieve it with a regex and parse the results

`.  We can then `JSON.parse` the matched content into something like `Partial<ReleaseMetadata>` where we have
```ts
interface ReleaseMetadata {
  version: string;
  name: string;
}


const metadataMatcher = RegExp(/<!--ReleaseMetadata.*(\{.*\})/);

for (const entry of entries) {
  const metadata = JSON.parse(entry.match(metadataMatcher)[1]) as Partial<ReleaseMetadata>;
}
```

There would need to be more validation/assertions I suppose, but this concept.

ng-dev/release/notes/changelog.ts
devversion: That's fair 👍 this comment on the PR is sufficient I assume

ng-dev/release/notes/changelog.ts
devversion: yeah, something around that is what I'd love to see. Could also be just something like:

```
<!-- ReleaseMetadata: Version|Name-->
```

Doesn't necessarily need to be JSON. I'm happy with either thing.

josephperrott: I think it makes sense to use JSON since it will make parsing easier.  It won't be as whitespace sensitive and is easy to roll and unroll.

devversion: 👍

josephperrott: Actually as I am looking at it now, I think it kind of ends up being a lot more change than we currently have to set this up, how do you feel about doing it in a follow up PR?  It just feels like it ends up nearly unrelated to the current change.

devversion: sure, works for me!

ng-dev/release/notes/changelog.ts
```diff
@@ -0,0 +1,141 @@
...
+   * version to the changelog archive.
+   *
+   * Versions should be used to determine which entries are moved to archive as versions are the
+   * most accurate piece of context found within a changelog entry to determine its relationship to
+   * other changelog entries.  This allows for example, moving all changelog entries out of the
+   * main changelog when a version moves out of support.
+   */
+  moveEntriesPriorToVersionToArchive(version: semver.SemVer) {
+    [...this.entries].reverse().forEach((entry: ChangelogEntry) => {
+      if (version.compare(entry.version) === 1) {
```
devversion, line 80: I think we should use: semver.lt or gt as that is more obvious, compared to the current function whic primarily exists for comparators in e.g a sort

ng-dev/release/notes/changelog.ts
josephperrott: Creating tracking issue at: https://github.com/angular/dev-infra/issues/219

ng-dev/release/notes/changelog.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  /** Update the changelog files with the known changelog entries. */
+  private writeToChangelogFiles(): void {
+    const changelog = this.entries.map((entry) => entry.content).join(joinMarker);
+    writeFileSync(this.filePath, changelog);
+
+    // Only update the changelog archive if it previously had content, or if there is now content.
+    if (this.alreadyHasChangelogArchive || this.archiveEntries.length) {
+      const changelogArchive = this.archiveEntries.map((entry) => entry.content).join(joinMarker);
+      writeFileSync(this.archiveFilePath, changelogArchive);
```
devversion, line 96: How about just having a separate method for writing the archive instead? Right now we'd parse the changelog archive file unnecessarily when just an entry has been prepended?

devversion (approved):

PR #215, opened 2021-09-13
feat(bazel): support for terser v5 and 2020 in rollup_bundle rule

devversion:
Supports for terser v5 in the `ng_rollup_bundle` rule. currently
terser fails because there is a conflicting `--beautify` CLI flag
set by `@bazel/terser` and the `output` option we manually control.

Instead of `output`, the new non-conflicting `format` name should be used:

```
ERROR: Please only specify either output or format option, preferrably
format.
```


josephperrott (approved):
LGTM

PR #216, opened 2021-09-13
fix(ng-dev/pr): Move the cleanup of the merge attempt to the finally block

josephperrott:
Move the cleanup of the merge attempt to the finally block to ensure that the cleanup
of the generated branches occurs regardless of the success of the merge attempt.

### DIFF
--- a/ng-dev/pr/merge/task.ts
+++ b/ng-dev/pr/merge/task.ts
@@ -130,13 +130,11 @@ export class PullRequestMergeTask {
 
     // Branch or revision that is currently checked out so that we can switch back to
     // it once the pull request has been merged.
-    let previousBranchOrRevision: null | string = null;
+    const previousBranchOrRevision = this.git.getCurrentBranchOrRevision();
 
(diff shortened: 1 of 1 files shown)


ng-dev/pr/merge/task.ts
```diff
@@ -162,11 +154,7 @@ export class PullRequestMergeTask {
       }
       throw e;
     } finally {
-      // Always try to restore the branch if possible. We don't want to leave
-      // the repository in a different state than before.
-      if (previousBranchOrRevision !== null) {
-        this.git.runGraceful(['checkout', '-f', previousBranchOrRevision]);
-      }
+      await strategy.cleanup(pullRequest, previousBranchOrRevision);
```
devversion, line 157: Everything looks good, except that I still believe the previous branch should be checked out here explicitly and not as part of the strategy cleanup. The strategy is only concerned with cleaning up its artifacts. Could you please change that?

devversion (approved):
LGTM. One small comment

PR #218, opened 2021-09-14
feat(ng-dev): add safety checks for shallow repositories

devversion:
Adds safety checks for shallow repositories when executing
the merge or release publish script. These two commands
have similar checks on the local repository and these tools
act on the local repository as part of a common development
workflow where an error beforehand would be more useful.


josephperrott (approved):
LGTM

PR #220, opened 2021-09-15
fix(bazel): no browsers matching for windows x86_64 cpu

devversion:
With the recent refactorings for supporting M1 browsers, we
added our own constraints/config settings. An incorrect assumption
was made that the x64 CPU constraint would also match the x32 constraint
given that Windows x64 can always run x32 programs. This seems to be
not the case (and makes sense in retrospective).

This commit fixes the browser configurations by always using x64
instead. The x32 constraint is removed because `rules_webtesting`
and other parts of the tooling seem to use x64 windows only, and
at this point supporting 32 bit windows seems not wort maintaining/
the additional maintenance.


josephperrott (approved):
LGTM

PR #221, opened 2021-09-15
ci: upload test result XML files to CircleCI after tests

josephperrott:
Uploading test results to CircleCI will provide both the test failure information in the CircleCI UI
as well as allowing CircleCI to provide insights information about test success rates on an individual
test level.

This is useful both in this repository, but will also act as a proving ground for making this kind of
information available in other repositories we manage to gather better metrics.

[Here](https://app.circleci.com/pipelines/github/angular/dev-infra/415/workflows/abe37406-8cb7-43db-8974-feda2aab5323/jobs/841/tests) is an example of what failing tests look like in the CircleCI UI.

### DIFF
--- a/.circleci/copy-junit-reports.js
+++ b/.circleci/copy-junit-reports.js
@@ -1,0 +1,53 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+const {execSync} = require('child_process');
+const {join} = require('path');
+const {mkdirSync, rmSync, statSync, readdirSync, copyFileSync} = require('fs');
+
+/**
+ * Discover all test results, which @bazel/jasmine stores as `test.xml` files, in the directory and
+ * return back the list of absolute file paths.
+ */
+const findTestResultsInDir = function (dirPath, files) {
+  for (const file of readdirSync(dirPath)) {
+    const filePath = join(dirPath, file);
+    if (statSync(filePath).isDirectory()) {
+      files = findTestResultsInDir(filePath, files);
+    } else {
+      if (file === 'test.xml') {
+        files.push(filePath);
+      }
+    }
+  }
+  return files;
+};
+
+/** StdOut result from the bazel info command. */
+const bazelInfoOutput = execSync('yarn bazel info', {stdio: 'pipe', encoding: 'utf8'});
+/** Absolute path to the bazel instance's testlog directory.  */
+const testLogPath = bazelInfoOutput.match(/bazel\-testlogs: (.*)/)[1];
+/** List of test result files. */
+const testResultPaths = findTestResultsInDir(testLogPath, []);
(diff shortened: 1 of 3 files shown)


.circleci/copy-junit-reports.js
```diff
@@ -0,0 +1,53 @@
...
+      if (file === 'test.xml') {
+        files.push(filePath);
+      }
+    }
+  }
+  return files;
+};
+
+/** StdOut result from the bazel info command. */
+const bazelInfoOutput = execSync('yarn bazel info', {stdio: 'pipe', encoding: 'utf8'});
```
devversion, line 31: bazel info bazel-testlogs
```suggestion
const bazelInfoOutput = execSync('yarn bazel info bazel-testlogs', {stdio: 'pipe', encoding: 'utf8'});
```

.circleci/copy-junit-reports.js
```diff
@@ -0,0 +1,53 @@
...
+ * Discover all test results, which @bazel/jasmine stores as `test.xml` files, in the directory and
+ * return back the list of absolute file paths.
+ */
+const findTestResultsInDir = function (dirPath, files) {
+  for (const file of readdirSync(dirPath)) {
+    const filePath = join(dirPath, file);
+    if (statSync(filePath).isDirectory()) {
+      files = findTestResultsInDir(filePath, files);
+    } else {
+      if (file === 'test.xml') {
```
devversion, line 22: It seems like Bazel does not guarantee this being named like that. Can we use all
XML files? Just an optional comment/question. It does not matter a lot

devversion (approved):

.circleci/copy-junit-reports.js
josephperrott: I googled this for like 10 minutes and could not find how to do this.  I feel silly now.

.circleci/copy-junit-reports.js
josephperrott: Sure we can just use all XML files, as it looks like they are the only files.

.circleci/copy-junit-reports.js
```diff
@@ -19,18 +19,20 @@ const findTestResultsInDir = function (dirPath, files) {
...
     }
   }
   return files;
 };
 
-/** StdOut result from the bazel info command. */
-const bazelInfoOutput = execSync('yarn bazel info', {stdio: 'pipe', encoding: 'utf8'});
 /** Absolute path to the bazel instance's testlog directory.  */
-const testLogPath = bazelInfoOutput.match(/bazel\-testlogs: (.*)/)[1];
+const testLogPath = execSync('yarn bazel info bazel-testlogs', {
```
devversion, line 32: Nit: Should we use --silent for Yarn or will that pollute the stdout?

devversion (approved):
LGTM, one minor nit

.circleci/copy-junit-reports.js
josephperrott: Good call, I have it set to always be silent in my env, so I forget about it sometimes.

PR #226, opened 2021-09-16
feat(bazel): switch away from deprecated rollup plugins

devversion:
Switches away from the deprecated rollup plugins and uses
their corresponding first-party `@rollup`-scoped plugins.


josephperrott (approved):
LGTM

Issue #203, opened 2021-09-02
[Proposal] Refactor ng-dev's pr directory to disentangle the utilities and common pieces


josephperrott:
Currently we have a structure like this.
```
ng-dev/pr                   
├── checkout                │ Checks out a PR locally to modify
├── check-target-branches   │ Prints the branches a pr would merge into
├── common                  │ 
│   └── checkout-pr.ts      │ Actual logic for checking out a PR used by checkout and rebase
├── discover-new-conflicts  │ Determines what other pending PRs can merge cleanly after a PR merges
├── merge                   │ Merges PRs (also houses the config.ts)
└── rebase                  │ Attempts to rebase a PR and push the rebased commits back to upstream
```

We have had a few people ask for something akin to `ng-dev pr check <pr-number>`, which would return the pr title, number, if it needs approvals, what branches it would land in etc.  I think we should transform the `check-target-branches` into this command.

I would propose that we then separate out a few pieces that can be reused both literally in other files/commands and conceptually.

```
ng-dev/pr                   
├── checkout                │ Stays the same
├── info                    │ Leverage the validation logic and PullRequest object to get all PR info
├── common                  │ 
│   ├── checkout-pr         │ Stays the same
│   ├── get-pr              │ Common util to get PR from github as a consistent PullRequest object
│   ├── targeting           │ Target labels and utils to get a target branch for a PR, etc
│   └── validation          │ Validate that a PR meets all of the expected criteria for the repo
├── discover-new-conflicts  │ Stays the same
├── merge                   │ Merges PR, using the new common utils
└── rebase                  │ Stays the same
```

Additionally, I think that `config.ts` should move up one level into `ng-dev/pr`

PR #227, opened 2021-09-16
refactor(ng-dev/pr): move pr configuration out of the merge directory

josephperrott:
As part of #203, refactoring the pr directory into more reusable pieces.

### DIFF
--- a/ng-dev/pr/config.ts
+++ b/ng-dev/pr/config.ts
@@ -6,15 +6,21 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {ConfigValidationError, GithubConfig} from '../../utils/config';
-
-import {GithubApiMergeStrategyConfig} from './strategies/api-merge';
+import {ConfigValidationError, GithubConfig} from '../utils/config';
 
 /**
  * Possible merge methods supported by the Github API.
  * https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button.
  */
 export type GithubApiMergeMethod = 'merge' | 'squash' | 'rebase';
+
+/** Configuration for the Github API merge strategy. */
+export interface GithubApiMergeStrategyConfig {
+  /** Default method used for merging pull requests */
+  default: GithubApiMergeMethod;
+  /** Labels which specify a different merge method than the default. */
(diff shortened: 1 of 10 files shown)


ng-dev/pr/config.ts
```diff
@@ -6,16 +6,22 @@
...
+export interface GithubApiMergeStrategyConfig {
+  /** Default method used for merging pull requests */
+  default: GithubApiMergeMethod;
+  /** Labels which specify a different merge method than the default. */
+  labels?: {pattern: string; method: GithubApiMergeMethod}[];
+}
+
 /**
  * Configuration for the merge script with all remote options specified. The
  * default `MergeConfig` has does not require any of these options as defaults
```
devversion, line 27: This is something I'm not 100% sure about. Does it make sense to capture the merge-specific configuration in something as generic as `ng-dev/pr`? are there any benefits you foresee here?

josephperrott: I actually think we should go a step further and rename this config to PullRequestConfig, while the primary usage is for merging, we rely on it for other actions in the `pr` directory like the `determine-target-branches` (soon to be `info`).  While much of it is about merging PRs, it really it describing how the tooling interacts with PRs.

Additionally, this aligns with how the rest of the ng-dev tooling has a config per primary subcommand (i.e. `ReleaseConfig`, `CaretakerConfig`, `FormatConfig`)

devversion: That's fair. I wish some of the interfaces, like for the API merge strategy to stay local to their logic, but I don't feel strongly

devversion (approved):
Lgtm we should rename the config

ng-dev/pr/config.ts
josephperrott: We can look to move the merge strategy back, I wasn't going to move it initially, but it caused a circular dependency iat the build level.  I will move all the symbols used for the ApiMergeStrategy into the api-merge.ts file

josephperrott: Actually I missed a piece before that comment, we can't avoid the circular dependency

josephperrott:
This PR was merged into the repository by commit cf18f08e546a481b3c0fe5506df587a652946681.

PR #229, opened 2021-09-16
ci: only copy test results reports for tests which actually executed, ignoring cache hits

josephperrott:
Ignoring cache hits for passing tests improves the quality of the information uploaded as it will not repeatedly upload the same test results on every run.


.circleci/copy-junit-reports.js
```diff
@@ -47,9 +48,20 @@ const destDirPath = join(__dirname, '../test-results/jasmine');
...
 // Copy each of the test result files to the central test result directory which CircleCI discovers
 // test results in.
-testResultPaths.forEach((filePath, i) => {
-  const destFilePath = join(destDirPath, `results-${i}.xml`);
-  copyFileSync(filePath, destFilePath);
+testResultPaths.forEach(([xmlFilePath, cacheStatusFilePath]) => {
+  const shortFilePath = xmlFilePath.substr(testLogPath.length + 1);
+  const testResultData = TestResultData.deserializeBinary(readFileSync(cacheStatusFilePath));
+  if (testResultData.getRemotelyCached()) {
+    console.debug(`Skipping copy of ${shortFilePath} as it was a passing remote cache hit`);
```
devversion, line 59: Are they necessarily passing if they are remotely cached? IIUC it can also be cached in a failing (if all inputs are the same with a previously failing run)

devversion (approved):

.circleci/copy-junit-reports.js
josephperrott: By default (which we do not set the flag otherwise), bazel will[ rerun a test if "the test previously failed"](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/test/TestConfiguration.java;l=112).

The other options of `true` or `false` would cause it to always cache the results no matter what.

Though for our case, even if the test was cached and failed again, we didn't actually run it again so we wouldn't want to upload the result since its not a new run of the test.

devversion: Ahh, good point! should be good to go

josephperrott:
This PR was merged into the repository by commit 3fea02f7d20fb749cd144c2c30b52e591caf9b48.

PR #231, opened 2021-09-17
refactor: use proto file directly compared to prebuilt protobufjs helper

devversion:
This commit switches from the prebuilt protobufjs to the actual
proto file as this one is easier to update when the proto changes
in more recent Bazel versions. This is also a step forward to actually
consuming this proto from the local bazel installation using the
`@bazel_tools` repository, allowing for the script to be made version
agnostic.


josephperrott (approved):
LGTM

I initially did it this way during my development but found that it was significantly slower (~1.5s vs ~10s), but when I checkout and tested it now its much closer (~1.5s vs ~3s).  I think this way is better, but wanted to make sure I noted somewhere that we might be incurring a slightly processing penalty doing it dynamically like this.

tools/protos/test_status.proto
```diff
@@ -0,0 +1,118 @@
+// Copied from `external/@bazel_tools/src/main/protobuf/test_status.proto`.
+// https://github.com/bazelbuild/bazel/blob/master/src/main/protobuf/test_status.proto.
```
josephperrott, line 2: Lets use the permalink to the exact version we are using since it is tagged on github
```suggestion
// https://github.com/bazelbuild/bazel/blob/4.2.1/src/main/protobuf/test_status.proto
```

josephperrott (approved):

josephperrott:
This PR was merged into the repository by commit 06ecdc29197ec441e0938d97df1e412348046915.

PR #232, opened 2021-09-17
ci: upload empty test result report when all tests are cache hits on CI

josephperrott:
When tests are requested and result in 100% cache hits, CircleCI should still be informed that a test run occurred. If no files are copied, it signals that no tests were actually executed. Always uploading at least one test result report will cause CircleCI to include every workflow run in its aggregated data and provide better metrics about the number of executed tests per run.


devversion (approved):

josephperrott:
This PR was merged into the repository by commit 90bbef480112a20c010f0b972e5de88f73d0a5fa.

PR #233, opened 2021-09-17
refactor(ng-dev/pr): move targeting and validation pr functions into common folder

josephperrott:
As part of #203, refactoring the pr directory into more reusable pieces


ng-dev/pr/common/validation/validations.ts
```diff
@@ -6,13 +6,15 @@
...
+import {MergeConfig} from '../../config';
 import {PullRequestFailure} from './failures';
-import {red, warn} from '../../utils/console';
-import {breakingChangeLabel} from './constants';
-import {RawPullRequest} from '../common/fetch-pull-request';
+import {red, warn} from '../../../utils/console';
+import {RawPullRequest} from '../fetch-pull-request';
+
+/** Label for pull requests containing a breaking change. */
+export const breakingChangeLabel = 'flag: breaking change';
```
devversion, line 17: This seems wrong to have here to be honest. There should be one canonical place where this is hard-coded and it should be in a easy-to-discover file that is not concerned with something else, like validation IMO

josephperrott: Agreed, I will see where I can find that is a better option

josephperrott:
@devversion PTAL

github-actions/breaking-changes-label/lib/BUILD.bazel
```diff
@@ -16,7 +16,7 @@ ts_library(
     deps = [
         "//github-actions:utils",
         "//ng-dev/commit-message",
-        "//ng-dev/pr/common",
+        "//ng-dev/pr:config",
```
devversion, line 19: Have you considered moving this into a `config/` folder instead? That is more aligned with what we do with `/release` and is more consistent overall with target naming.

devversion (approved):
LGTM overall. I think we _could_ just move config into a separate folder as well (but can be a follow-up too;)

github-actions/breaking-changes-label/lib/BUILD.bazel
josephperrott: If you would like to we can, but I personally prefer not doing it that way.  We can do it in  a followup later if we end up wanting to.

devversion: I think my main argument for this is that maintaining a target at the same level results in unnecessary extra code for excluding the `config.ts` file from the other glob. and I like the idea of a TS Bazel target matching just the directory where I can import it from (makes it easy to deal with deps). 

Either way works for me. Not feeling super strong but I prefer the other way haha 😄 We should clean up either release or pr IMO, but a follow-up works!

josephperrott: I went ahead and did it the way you outlined as I realized that my primary argument for doing it this way was around not relying on importing from a directories `index` file, but we sorted out that problem in the repo and rely on this throughout the repo now.

devversion: 👍 cool, thanks for doing that change!

josephperrott:
This PR was merged into the repository by commit c84672f615bd637498a06a98f5df549f79d8b101.

devversion (approved):

PR #234, opened 2021-09-17
refactor: move generic testing utilities into utils/testing from release/testing

josephperrott:
Many of the util functions and classes from release/testing could be leveraged elsewhere in the project and would be better served being located in utils/testing.


devversion (approved):
💯 

I wasn't sure initially whether it makes sense to move the sandbox client to the common utils because it's still somewhat specific to the release tool (making the assumption of `fetch` not needing to be faked somehow), but I agree that it can be a good basis for future tests (where we can evolve the sandbox client).

josephperrott:
This PR was merged into the repository by commit b72bd5f8ae62451026568fa3309817d1fa9dd573.

PR #236, opened 2021-09-20
build: update tsconfig.json to allow editor to properly follow typings/deps

josephperrott:
Update `tsconfig.json` to only include the directories which contain source code in the repository to prevent editors from discovering code in `bazel-out/` and `dist/` as source code.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,7 +7,7 @@
     "noImplicitOverride": true,
     "noImplicitReturns": true,
     "moduleResolution": "node",
-    "outDir": "dist/tsc-output"
+    "noEmit": false
   },
   "include": [
     "bazel/",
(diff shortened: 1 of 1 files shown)


tsconfig.json
```diff
@@ -2,10 +2,18 @@
...
     "target": "es2020",
-    "lib": ["es2020"],
+    "lib": ["DOM", "es2020"],
     "strict": true,
     "noImplicitOverride": true,
     "noImplicitReturns": true,
-    "moduleResolution": "node"
-  }
+    "moduleResolution": "node",
+    "outDir": "dist/tsc-output"
```
devversion, line 10: We can use `noEmit: true`

devversion (approved):

josephperrott:
This PR was merged into the repository by commit 3b41957b53f09f9e37be9eea08fc6a2e66eed17a.

PR #239, opened 2021-09-22
Create `ng-dev ci gather-test-results` command

josephperrott:
See individual commits.


tools/defaults.bzl
```diff
@@ -87,3 +87,73 @@ def pkg_npm(build_package_json_from_template = False, deps = [], **kwargs):
...
+def generate_ts_proto_module(name, protofile, visibility = ["//visibility:public"]):
+    """
+    Generate a typescript module for decoding a proto binary file based on a provided .proto file.
+    """
+    protobufjs_dependencies = [
+        "@npm//semver",
+        "@npm//chalk",
+        "@npm//jsdoc",
+        "@npm//minimist",
+        "@npm//uglify-js",
```
devversion, line 100: Is uglifyjs actually used? If so, we usually use terser in the Angular org instead.

devversion: Any details on why we need to manage dependencies manually here? aren't they brought in automatically as needed?

tools/defaults.bzl
```diff
@@ -87,3 +87,73 @@ def pkg_npm(build_package_json_from_template = False, deps = [], **kwargs):
         deps = deps,
         **kwargs
     )
+
+def generate_ts_proto_module(name, protofile, visibility = ["//visibility:public"]):
```
devversion, line 91: Can we move this into a separate `.bzl` file? I think this is rather a macro than a default at this point (given the amount of code)

devversion (approved):
LGTM. Nicely done. Just a few minor nits

tools/defaults.bzl
josephperrott: I initially did this, then second guessed myself :D

tools/defaults.bzl
josephperrott: I don't know if it is actually used, but if it is not available at runtime, already protobufjs attempts to npm install it during its execution.  

So since running it via `npm_package_bin` only includes access to the `dependencies` of the package you are running, these "optional" dependencies have to be provided in the `data` attribute.

devversion: lol, this is surprisingly unexpected for a NPM package. So uncommon to run `npm` as part of the CLI, but makes sense. We could optionally link/explain this in a comment. I think that would be helpful; but up to you

devversion (approved):
🎉

josephperrott:
This PR was merged into the repository by commit f64f8085a7b75368a97d05f0274471d42cf0f12f.

PR #242, opened 2021-09-24
Update status/checks retrieval from Github, move validation of CLA to use status instead of the label, move other validators to common location

josephperrott:
See individual commits.


ng-dev/pr/common/validation/validations.ts
```diff
@@ -102,3 +106,49 @@ export function assertPendingState(pullRequest: RawPullRequest) {
...
+  }
+  throw PullRequestFailure.notMergeReady();
+}
+
+/**
+ * Assert the pull request has been marked ready for merge by the author.
+ * @throws {PullRequestFailure} if the pull request is missing the merge ready label.
+ */
+export function assertPassingCi(pullRequest: PullRequestFromGithub) {
+  if (getStatusesForPullRequest(pullRequest).combinedStatus === PullRequestStatus.FAILING) {
```
devversion, line 142:
```suggestion
  if (getStatusesForPullRequest(pullRequest).combinedStatus === PullRequestStatus.PENDING) {
```

ng-dev/pr/common/validation/validations.ts
```diff
@@ -102,3 +106,49 @@ export function assertPendingState(pullRequest: RawPullRequest) {
...
+    throw PullRequestFailure.pendingCiJobs();
+  }
+  if (getStatusesForPullRequest(pullRequest).combinedStatus === PullRequestStatus.FAILING) {
```
devversion, line 145: We should store the status in a variable for more readability and to save us from another invocation.

ng-dev/pr/common/validation/validations.ts
```diff
@@ -102,3 +106,49 @@ export function assertPendingState(pullRequest: RawPullRequest) {
       throw PullRequestFailure.isMerged();
   }
 }
+
+/**
+ * Assert the pull request has all necessary CLAs signed.
+ * @throws {PullRequestFailure} if the pull request is missing a necessary CLA signature.
+ */
+export function assertSignedCla(pullRequest: PullRequestFromGithub) {
```
devversion, line 114: I guess, now I see why you wanted to store the statuses in a variable. Because now we need to compute them in two assertion functions here. Not really a big deal, but if we'd want to avoid this, then I think it would be reasonable to accept the statuses directly as parameter here. No strong opinion here though. It doesn't matter a lot I suppose.

ng-dev/pr/common/fetch-pull-request.ts
```diff
@@ -65,13 +101,92 @@ const PR_SCHEMA = {
...
+  return await getPendingPrs(PR_SCHEMA, git);
+}
+
+/**
+ * Normalize the statuses for a commit from a pull requeste, using a consistent interface for both
+ * status and checks results.
+ */
+export function getStatusesForPullRequest(pullRequest: PullRequestFromGithub) {
+  /** The combined github status and github checks object. */
+  const {statusCheckRollup} = pullRequest.commits.nodes.slice(-1)[0].commit;
```
devversion, line 127: Nit: I think using the index is more clean and avoids us from cloning the array. e.g.

```ts
  const nodes = pullRequest.commits.nodes;
  const {statusCheckRollup} = nodes[nodes.length - 1].commit;
```

ng-dev/pr/common/fetch-pull-request.ts
```diff
@@ -65,13 +101,92 @@ const PR_SCHEMA = {
...
+
+/** Fetches a pull request from Github. Returns null if an error occurred. */
+export async function fetchPendingPullRequestsFromGithub(
+  git: AuthenticatedGitClient,
+): Promise<PullRequestFromGithub[] | null> {
+  return await getPendingPrs(PR_SCHEMA, git);
+}
+
+/**
+ * Normalize the statuses for a commit from a pull requeste, using a consistent interface for both
```
devversion, line 122:
```suggestion
 * Gets  the statuses for a commit from a pull requeste, using a consistent interface for both
```

ng-dev/pr/common/fetch-pull-request.ts
```diff
@@ -25,8 +41,28 @@ const PR_SCHEMA = {
       nodes: [
         {
           commit: {
-            status: {
-              state: graphqlTypes.oneOf(['FAILURE', 'PENDING', 'SUCCESS'] as const),
+            statusCheckRollup: {
+              state: graphqlTypes.custom<StatusState>(),
+              contexts: params(
+                {last: 100},
```
devversion, line 47: What are we doing about the 100 here? should we add pagination? It's unlikely to hit 100, but it seems like this is a place where we _should_ be correct (as it's a common utility now)

devversion (approved):
This turned out pretty nice IMO! nice work 🎉

ng-dev/pr/common/validation/validations.ts
josephperrott: Lets keep it as is, I personally like the fact that you provide the entire `PullRequestFromGithub` object that you are validating the assumption for rather than having to know what parts of a pull request have to be checked for each assertion.

devversion: 👍

ng-dev/pr/common/fetch-pull-request.ts
josephperrott: We should probably address it, but I think it makes sense in a follow up as it will be a bit complicated to do.

devversion: Works for me 👍

josephperrott:
This PR was merged into the repository by commit 93e501e32b59f0330bb281b00537bd4b9a78a9b6.

PR #247, opened 2021-10-01
fix(ng-dev/release): prepare-commit-message hook accidentally running when bump commit is created

devversion:
The `prepare-commit-message` hook can still accidentally run when the
changelog / version bump commit is created. This is problematic because
when a version branch is checked out (like the patch branch), the node
modules are not necessarily updated/re-installed, and the node modules
from master remain installed. This can cause runtime errors when Git
hooks accidentally run with a tool from the node modules (like `ng-dev`
itself).

We fix this by also setting the `HUSKY` environment variable, similar to
how the merge tool does. We can assume that Husky is used in all
repositories. This should be our standard for hooks.


josephperrott (approved):
LGTM

PR #256, opened 2021-10-08
fix(ng-dev): allow deprecations in PRs during feature freeze

josephperrott:
Currently our tooling does not distinguish between when RC is occurring and when feature freeze is occurring.
Since they are treated as the same state, we are unable to allow deprecations messages on feature freeze without
allowing it during RC periods.  By creating this distinction we can properly handle this case.

### DIFF
--- a/ng-dev/release/versioning/active-release-trains.ts
+++ b/ng-dev/release/versioning/active-release-trains.ts
@@ -16,14 +16,27 @@ import {
   VersionBranch,
 } from './version-branches';
 
-/** Interface describing determined active release trains for a project. */
-export interface ActiveReleaseTrains {
+/** The active release trains for a project. */
+export class ActiveReleaseTrains {
   /** Release-train currently in the "release-candidate" or "feature-freeze" phase. */
-  releaseCandidate: ReleaseTrain | null;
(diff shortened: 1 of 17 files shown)


ng-dev/release/versioning/active-release-trains.ts
```diff
@@ -16,14 +16,21 @@ import {
...
+    public readonly releaseCandidate: ReleaseTrain | null,
+    /** Release-train in the `next` phase. */
+    public readonly next: ReleaseTrain,
+    /** Release-train currently in the "latest" phase. */
+    public readonly latest: ReleaseTrain,
+  ) {}
+
+  /** Whether the active release trains indicate the repository is in a feature freeze state. */
+  isFeatureFreeze() {
+    return this.releaseCandidate !== null && this.releaseCandidate.version.prerelease[0] === 'next';
```
devversion, line 32: Can we use this helper in the cut release candidate from feature freeze action?

devversion (approved):
Looks awesome 🎉 

I'm slightly unhappy that the creation of the active release trains is not as obvious as before. i.e. with the object it was clear what `ReleaseTrain` goes into `next` for example, while now we have positional arguments.

PR #258, opened 2021-10-11
feat(bazel): introduce rule for running integration tests with Bazel

devversion:
See individual commits


bazel/integration/test_runner/runner.ts
```diff
@@ -0,0 +1,215 @@
...
+    if (process.env.TEST_TMPDIR) {
+      // Bazel itself cleans the temporary directory, so the direct cleanup
+      // function is a noop.
+      return process.env.TEST_TMPDIR;
+    }
+
+    return new Promise((resolve, reject) => {
+      tmp.dir(
+        {
+          template: 'ng-integration-test-XXXXXX',
```
josephperrott, line 75: Where is this temp directory created?  In `/tmp`?

.nvmrc
```diff
@@ -1 +1 @@
-16.6.0
+16.10.0
```
josephperrott, line 1: Should we just go to `16.11.0` since its out?

josephperrott (commented):
Overall its looking good, looks like there is a failure on CI currently

bazel/integration/test_runner/runner.ts
devversion: yeah, that is the system temp directory, or whatever the `tmp` library maintains.

PR #260, opened 2021-10-11
feat(ng-dev): remove prerelease changelog entries when cutting a stable release

josephperrott:
When cutting a stable release, the changelog entries for the prerelease entries are removed as stable release
note entries contain the entirety from the previous stable release and the newly created stable release.


ng-dev/release/notes/release-notes.ts
```diff
@@ -63,9 +63,14 @@ export class ReleaseNotes {
...
    * Prepend generated release note to the CHANGELOG.md file in the base directory of the repository
-   * provided by the GitClient.
+   * provided by the GitClient. Removes entries for related prerelease entries as appropriate.
    */
   async prependEntryToChangelogFile() {
+    // When the version for the entry is a non-prelease (i.e. 1.0.0 rather than 1.0.0-next.1), the
+    // pre-release entries for the version should be removed from the changelog.
+    if (semver.prerelease(this.version) === null) {
+      Changelog.removePrereleaseEntriesForVersion(this.git, this.version);
+    }
```
devversion, line 73: I'm not sure what it would look like if we'd manually invoke the remove pre-release function in the release action when staging (keeping this function more concerned with just prepending), but not sure 🤷‍♂️.

Anyway, have you considered adding a test that validates the removal as part of a release action? that is more reflecting what we'll end up with the release tool.

josephperrott: I initially looked at doing it that way, it its a somewhat sizable refactor to bring the logic required out of the base class to allow it to be different depending on what action is running.

Since we are always wanting the first stable release notes to remove the prerelease notes, this logic of removing all prerelease notes for non-prerelease versions has the same effect.

I will add a validation test as suggested.

devversion (approved):
@josephperrott looks awesome!

Can you also please respond to the review message I added as part of my intitial review? https://github.com/angular/dev-infra/pull/260#pullrequestreview-777381925

devversion:
I think this PR is good to go though; but I think it's the perfect spot to discuss this problem because this PR introduces the problem.

clydin:
I think it's a useful ability to be able to adjust the change log in the release PR (typos happen and breaking changes may need to be augmented).

What if the tool warned the user that customized entries exist so that the change log can be manually reviewed?  This could be a followup and definitely not blocking this PR.

Also, would it be possible to leverage the change log content itself for a particular version (edited or otherwise) as the basis for the Github release content? This could remove the disconnect between edited change log file content and the release description.

PR #262, opened 2021-10-12
test: enable sharding for publish release test

devversion:
Enables sharding for the publish release test as it became
slower over time, most likely the actual Sandbox git client
testing had an effect on this as well.

The test currently took ~20sec with RBE on CI which is rather
slow compared to other tests.


josephperrott:
This PR was merged into the repository by commit a1b31520146be530caf7cbacbefc75ce536e785c.

PR #264, opened 2021-10-13
fix(ng-dev/release): run yarn integrity check before performing a release

josephperrott:
If the ng-dev command is run with an outdate version of the package, a dev may perform a release without
the latest changes to the tooling.  This can lead to not creating up to date artifacts and general issues.
By checking if the installed dependencies match the expected dependencies before performing the action we
can instruct the user to update their installed node_modules.

An example where this change would have prevented issue:
Framework releases from the patch branch which is currently a different version of the package as defined in
package.json. Calling them version 1 and version 2, with version 2 being the latest version. A patch release
is performed by checking out the latest commit from the upstream next branch, running yarn install and then
running ng-dev release publish.  This release process will occur using version 2 of the dev-infra package.
During the release process, the patch branch is checked out and yarn install is performed, causing version 1
to be installed.  Upon completion of the patch release, the next release is attempted unknowingly triggered
using version 1 of the package.  While generated artifacts are already ensured to built with the correct
dependency by the process, the tool itself is already running before this check can occur. Instead version 1
of the package is used to process the release, potentially resulting in an errant or unexpected process.

This situation will now be prevented by validating the expected version is being used for the action process,
in the situation above the user would be informed that yarn install was needed before performing the second
release.

### DIFF
--- a/ng-dev/release/publish/external-commands.ts
+++ b/ng-dev/release/publish/external-commands.ts
@@ -118,11 +118,7 @@ export async function invokeYarnIntegryCheck(projectDir: string): Promise<void> 
 export async function invokeYarnVerifyTreeCheck(projectDir: string): Promise<void> {
   try {
     await spawn('yarn', ['check', '--verify-tree'], {cwd: projectDir, mode: 'silent'});
-    info(
-      green(
-        '  ✓   Confirmed installed installed dependencies match those defined in package.json.',
-      ),
-    );
+    info(green('  ✓   Confirmed installed dependencies match those defined in package.json.'));
   } catch (e) {
     error(red('  ✘   Failed yarn verify tree check, your installed dependencies are likely out'));
     error(red('      of date. Please run `yarn install` to update your installed dependencies.'));


ng-dev/release/publish/external-commands.ts
```diff
@@ -97,3 +97,17 @@ export async function invokeYarnInstallCommand(projectDir: string): Promise<void
     throw new FatalReleaseActionError();
   }
 }
+
+/**
+ * Invokes the `yarn check --integrity` command in order to verify up to date dependencies.
+ */
+export async function invokeYarnIntegryCheck(projectDir: string): Promise<void> {
+  try {
+    await spawn('yarn', ['check', '--integrity'], {cwd: projectDir, mode: 'silent'});
```
devversion, line 106: Do we actually need https://classic.yarnpkg.com/en/docs/cli/check/#toc-yarn-check-verify-tree?

seems like the integrity check just checks the package json against the lock file unless I misunderstand the documentation.

ng-dev/release/publish/external-commands.ts
```diff
@@ -104,10 +104,28 @@ export async function invokeYarnInstallCommand(projectDir: string): Promise<void
...
+
+/**
+ * Invokes the `yarn check --verify-tree` command in order to verify up to date dependencies.
+ */
+export async function invokeYarnVerifyTreeCheck(projectDir: string): Promise<void> {
+  try {
+    await spawn('yarn', ['check', '--verify-tree'], {cwd: projectDir, mode: 'silent'});
+    info(
+      green(
+        '  ✓   Confirmed installed installed dependencies match those defined in package.json.',
```
devversion, line 123: double installed

ng-dev/release/publish/external-commands.ts
devversion: I think that makes sense, yeah 👍

PR #270, opened 2021-10-15
feat(bazel): add support for custom environment variables in integration tests

devversion:
Adds support for custom environment variables in integration tests. The
environment variable dictionary can use location expansion. Also there
is a special placeholder called `<TMP>` that can be used to acquire a
temporary directory. This is useful when setting up tools like BAZELISK,
which requires a `HOME` environment variable as an example.

### DIFF
--- a/bazel/integration/test_runner/runner.ts
+++ b/bazel/integration/test_runner/runner.ts
@@ -51,7 +51,7 @@ export class TestRunner {
   async run() {
     const testDir = await this._getTestTmpDirectoryPath();
     const toolMappings = await this._setupToolMappingsForTest(testDir);
-    const testEnv = await this._createTestProcessEnvironment(testDir, toolMappings.binDir);
+    const testEnv = await this._buildTestProcessEnvironment(testDir, toolMappings.binDir);
 
     console.info(`Running test in: ${path.normalize(testDir)}`);
 
@@ -175,13 +175,13 @@ export class TestRunner {
   }
 
   /**
-   * Creates the test process environment. The specified tools bin directory
-   * will be added to the environment `PATH` variable.
+   * Builds the test process environment object literal. The specified tools bin
+   * directory will be added to the environment `PATH` variable.
    *
    * User-specified environment variables can use Bazel location expansion as well
    * as a special placeholder for acquiring a temporary directory for the test.
    */
-  private async _createTestProcessEnvironment(
+  private async _buildTestProcessEnvironment(
     testDir: string,
     toolsBinDir: string,
   ): Promise<NodeJS.ProcessEnv> {


bazel/integration/test_runner/runner.ts
```diff
@@ -38,18 +44,20 @@ export class TestRunner {
     private readonly testPackage: string,
     private readonly toolMappings: Record<string, BazelFileInfo>,
     private readonly npmPackageMappings: Record<string, BazelFileInfo>,
-    private readonly commands: [[binary: string, ...args: string[]]],
+    private readonly commands: [[binary: BazelExpandedValue, ...args: BazelExpandedValue[]]],
+    private readonly environment: Record<string, BazelExpandedValue>,
```
josephperrott, line 48: Should `TestRunner` just take in a `TestConfig` object rather than separating it out like this and managing the list twice?

bazel/integration/test_runner/runner.ts
```diff
@@ -167,24 +175,50 @@ export class TestRunner {
   }
 
   /**
-   * Runs the test commands sequentially in the test directory. An additional directory
-   * that is added to the command process `$PATH` environment variables can be specified.
+   * Creates the test process environment. The specified tools bin directory
```
josephperrott, line 178: Can we name this as `_buildTestProcessEnvironment`, when I read create I thought it was changing the environment rather than creating a new `ProcessEnv` object.  For whatever reason, when I see "create environment" I think of it as "setup environment".

```suggestion
   * Creates a new `ProcessEnv` for the test. The specified tools bin directory
```

bazel/integration/index.bzl
```diff
@@ -6,10 +6,30 @@ def _serialize_file(file):
...
     return struct(path = file.path, shortPath = file.short_path)
 
+def _serialize_and_expand_location(ctx, value):
+    """Expands Bazel make location expressions for the given value. Returns a JSON
+      serializable dictionary matching the `BazelExpandedValue` type in the test runner."""
+    new_value = ctx.expand_location(value, targets = ctx.attr.data)
+
+    return {
+        "value": new_value,
+        "containsExpandedValue": new_value != value,
```
josephperrott, line 16: If I understand correctly, while this is JSON serializable, its not actual put into JSON, so does this properly match the typings expect?

I would think that this is:
```ts
{
  value: string;
  containsExpandedValue: 'True' | 'False';
}
```

rather than 
```ts
{
  value: string;
  containsExpandedValue: boolean;
}
```

If this is the case and '"False"' is provided, if we expect a boolean, then we are going to get alot of unexpected results.

bazel/integration/test_runner/runner.ts
devversion: that's fair. Will change

bazel/integration/test_runner/runner.ts
devversion: I'd prefer keeping it that way as the test config was just a way of transferring state from Starlark to JS. Surely this could be the interface for the test runner itself, but subjectively the property access is a a little more convenient that way. We can revisit in the future.

(also we only use the ctor once where the parameters are passed in an obvious way, almost as if they were keyed)

josephperrott: Okay, thats fine with me, I just wanted to make sure I pointed it out as its is a bit atypical.  But 👍

bazel/integration/index.bzl
josephperrott: That is what I meant, I just was unsure of where the dictionary gets encoded as JSON in the process and wanted to make sure to confirm

bazel/integration/test_runner/runner.ts
devversion: yeah, definitely a very good point. Looks like both approaches have small trade-offs

josephperrott (approved):
LGTM

PR #279, opened 2021-11-03
feat(tslint-rules): move validate-import-for-esm-cjs-interop from framework

devversion:
See individual commits


josephperrott (approved):
LGTM

PR #284, opened 2021-11-05
Allow integration test rules to access browser binaries managed within Bazel

devversion:
See individual commits. Allows us to use Chromium from Bazel, instead of having to always manage/install puppeteer for Browser-based integration tests.

### DIFF
--- a/bazel/integration/test_runner/process_utils.ts
+++ b/bazel/integration/test_runner/process_utils.ts
@@ -47,7 +47,7 @@ export function runCommandInChildProcess(
   workingDir: string,
   env: NodeJS.ProcessEnv,
 ): Promise<boolean> {
-  const humanReadableCommand = `${binary} ${args.join(' ')}`;
+  const humanReadableCommand = `${binary}${args.length ? ` ${args.join(' ')}` : ''}`;
 
   debug(`Executing command: ${humanReadableCommand} in ${workingDir}`);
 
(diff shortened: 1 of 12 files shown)


bazel/integration/test_runner/process_utils.ts
```diff
@@ -44,7 +47,7 @@ export function runCommandInChildProcess(
   workingDir: string,
   env: NodeJS.ProcessEnv,
 ): Promise<boolean> {
-  const humanReadableCommand = `${binary} ${args.join(' ')}`;
+  const humanReadableCommand = `${binary}${args.length ? ` ${args.join(' ')}` : ''}`;
```
josephperrott, line 50: I think we can leave it as since when `[].join(' ')` is `''`

josephperrott (approved):
LGTM

bazel/integration/test_runner/process_utils.ts
devversion: you are right, but unless I miss something, doing this will always result in the additional trailing space if there are no arguments.

josephperrott: I suppose thats true.  Either way you want to do it is fine.

devversion: yeah, let's keep as is as I just did that to not have the weird space in the debug messages.

PR #285, opened 2021-11-06
feat(bazel): allow integration commands to resolve executables through expansion

devversion:
Given a tool built using `nodejs_binary` or `sh_binary`, the tool cannot
be used as binary in an integration test command currently because:

* Bazel location expansion does not respect the executable "FilesToRun"
  information of the binaries, and rather sees _all_ outputs, erroring
  that `$(locations X)` need to be used instead. See e.g. https://github.com/bazelbuild/bazel/issues/11820)

* The command is currently split by space and this decouples the Make
  funtion expression incorrectly into a broken expansion. e.g.
  `$(rootpath X)` will become `["$(rootpath", "X"]`).

This commit fixes both things by relying on the Bazel
`CommandHelper.java` class which handles the expansion of commands as
expected, with similar semantics of a `genrule` as per `GenRuleBase.java`.

This allows test authors to conveniently use a `nodejs_binary` or
`sh_binary` as command binary because the `CommandHelper` knows
exactly which targets provide executables or not. There is no good API
for the plain Make expansion of a command itself, so we use a little trick
(which is documented sufficiently in the code).

### DIFF
--- a/bazel/integration/index.bzl
+++ b/bazel/integration/index.bzl
@@ -54,7 +54,8 @@ def _expand_and_split_command(ctx, command):
     # https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/BashCommandConstructor.java;l=40;drc=2255ce4165f936f695111020fa664b259a875c4a.
     # https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/CommandHelper.java;l=275-282;drc=2255ce4165f936f695111020fa664b259a875c4a.
     if len(resolved_bash_command) != 3:
-        fail("Command too long. Please split this into a dedicated script: %s" % command)
+        fail("Test command too long. Please use a shorter one, or extract this " +
+             "into a dedicated script: %s" % command)
 
     # The third argument of the resolved `bash -c` command will hold the actually expanded command.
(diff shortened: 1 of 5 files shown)


bazel/integration/index.bzl
```diff
@@ -17,14 +25,50 @@ def _serialize_and_expand_value(ctx, value, description):
...
+        make_variables = ctx.var,
+        tools = ctx.attr.data,
+    )
+
+    # If the resolved command does not have three arguments, then there were too many arguments
+    # and Bazel extracted the command into a dedicated Bash script that we cannot read here.
+    # https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/BashCommandConstructor.java;l=40;drc=2255ce4165f936f695111020fa664b259a875c4a.
+    # https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/CommandHelper.java;l=275-282;drc=2255ce4165f936f695111020fa664b259a875c4a.
+    if len(resolved_bash_command) != 3:
+        fail("Command too long. Please split this into a dedicated script: %s" % command)
```
gkalpak, line 57: Is "split" the right verb here?

gkalpak (approved):

bazel/integration/index.bzl
devversion: good point. fixed 👍

PR #286, opened 2021-11-07
fix(bazel): integration test rule not able to setup mappings for resolutions

devversion:
Fixes that the Bazel integration test rule is not able to setup mappings
for Yarn resolutions because the record keys do not exactly match to a
package name, but rather follow certain allowed patterns, like
`**/<pkg-name>`.


bazel/integration/test_runner/package_json.ts
```diff
@@ -49,10 +49,18 @@ export function updateMappingsForPackageJson(
 ): PackageJson {
   const newPackageJson = {...packageJson};
 
-  updateMappingForRecord(newPackageJson.dependencies ?? {}, mappings);
-  updateMappingForRecord(newPackageJson.devDependencies ?? {}, mappings);
-  updateMappingForRecord(newPackageJson.optionalDependencies ?? {}, mappings);
-  updateMappingForRecord(newPackageJson.resolutions ?? {}, mappings);
+  updateMappingForRecord(newPackageJson, 'dependencies', mappings);
+  updateMappingForRecord(newPackageJson, 'dependencies', mappings);
```
gkalpak, line 53:
```suggestion
  updateMappingForRecord(newPackageJson, 'devDependencies', mappings);
```

bazel/integration/tests/package_mappings/package.json
```diff
@@ -5,5 +5,10 @@
   "license": "MIT",
   "dependencies": {
     "fake_pkg": "0.0.0"
+  },
+  "resolutions": {
+    "**/fake_pkg": "0.0.0",
+    "some_pkg/**/fake_pkg": "0.0.0",
+    "fake_pkg": "0.0.0"
   }
```
gkalpak, line 13: Could we also use `devDependencies` and `optionalDependencies` here?

bazel/integration/tests/package_mappings/test.mjs
```diff
@@ -1,8 +1,16 @@
...
 // Sanity check that the installed package matches the one we have
 // built from source using `pkg_npm`.
 if (fakePkg !== 'This is a fake package!') {
   console.error('Fake package is not matching with locally-built one.');
   process.exitCode = 1;
 }
+
+const pkgJson = JSON.parse(fs.readFileSync('./package.json', 'utf8'));
+
+if (Object.values(pkgJson.resolutions).some((v) => v === '0.0.0')) {
```
gkalpak, line 13: Nit: This could be simplified as:

```suggestion
if (Object.values(pkgJson.resolutions).includes('0.0.0')) {
```

bazel/integration/test_runner/package_json.ts
```diff
@@ -82,6 +106,24 @@ function updateMappingForRecord(record: DependencyRecord, mappings: PackageMappi
...
+    record[entryKey] = mappedAbsolutePath;
+  }
+}
+
+/**
+ * Gets the package name from a dependency record entry.
+ *
+ * @param entryKey Key of the dependency record entry.
+ * @param nameMatchRegex Optional regular expression that can be specified to match the package
+ *   name in a dependency entry. This is useful for e.g. Yarn resolutions using patterns.
```
gkalpak, line 118: Might be worth mentioning that the value matched by the first capturing group of the regex is used as the package name.

bazel/integration/test_runner/package_json.ts
```diff
@@ -82,6 +106,24 @@ function updateMappingForRecord(record: DependencyRecord, mappings: PackageMappi
...
+ */
+function getPackageNameFromDependencyEntry(
+  entryKey: string,
+  nameMatchRegex?: RegExp,
+): string | null {
+  if (nameMatchRegex) {
+    const matches = entryKey.match(nameMatchRegex);
+    return matches === null ? null : matches[1];
```
gkalpak, line 126: Nit: This could be simplified as:

```suggestion
    return matches?.[1];
```

bazel/integration/test_runner/package_json.ts
```diff
@@ -49,10 +49,18 @@ export function updateMappingsForPackageJson(
...
+  updateMappingForRecord(newPackageJson, 'optionalDependencies', mappings);
+  // The object for Yarn resolutions will not directly match with the mapping keys
+  // specified here, as resolutions usually follow a format as followed:
+  //  1. `**/<pkg-name>`
+  //  2. `<other-pkg>/**/<pkg-name>`
+  //  3. `<pkg-name>`
+  // More details here: https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/.
+  // We pass a regular expression which matches the `<pkg-name>` so that the mappings can
+  // be applied for resolutions as well.
+  updateMappingForRecord(newPackageJson, 'resolutions', mappings, /(?:^|\/)?([^/]+)$/);
```
gkalpak, line 63: By making the `(?:^|\/)` optional, it basically makes it redundant. So, you could as well remove it altogether (and the regex would remain the same):

```suggestion
  updateMappingForRecord(newPackageJson, 'resolutions', mappings, /([^/]+)$/);
```

bazel/integration/test_runner/package_json.ts
devversion: ohh 🤦‍♂️ good catch!

bazel/integration/tests/package_mappings/package.json
devversion: 👍

bazel/integration/test_runner/package_json.ts
devversion: agreed 👍

bazel/integration/test_runner/package_json.ts
devversion: good point. this would return `undefined` through, unless we use an `??` later though. my preference has been to return `null` in explicit cases, so I'll keep it as is, but please let me know if you feel differently about `undefined vs. null`

PR #289, opened 2021-11-10
fix(ng-dev/release): properly ensure local ng-dev version is up-to-date

devversion:
Fixes that the ng-dev release tool currently relies on a historically broken
`yarn check` command for ensuring it is up-to-date. This check has proven to
not work well within the Angular repositories, so we switch to a more future-proof
and reliable approach by checking the lock file and comparing with the running
version through a substitued placeholder.


josephperrott (approved):
LGTM

PR #292, opened 2021-11-11
refactor(ng-dev): move check for outdated ng-dev local version to shared location

devversion:
See individual commits


crisbeto (approved):
LGTM

PR #303, opened 2021-11-16
feat(bazel): introduce rule for bundling specs with optional angular linker

devversion:
See individual commits.

This is a part of the rule/logic from `angular/components`. The code can then be used in the CLI, COMP and other Angular-owned projects (like the devtools as well).


devversion:
Note: I did not add `@angular/compiler-cli` as dependency because that would complicate things when the package is installed in FW. We did not add a dependency on FW packages for the benchmark code as well, so that should be fine for now.

bazel/esbuild/esbuild-amd-config.mjs
```diff
@@ -0,0 +1,24 @@
...
+const userConfigExecPath = 'TMPL_CONFIG_PATH';
+
+/** User ESBuild config. Empty if none is loaded. */
+let userConfig = {};
+
+if (userConfigExecPath !== '') {
+  const userConfigPath = path.join(process.cwd(), userConfigExecPath);
+  const userConfigUrl = url.pathToFileURL(userConfigPath);
+
+  // Load the user config, assuming it is set as `default` export.
```
josephperrott, line 14: Can we have this comment note that it is expecting the `default` export because we define it in our own file.

josephperrott (approved):
LGTM

PR #304, opened 2021-11-16
Create yarn update tooling

josephperrott:
See individual commits.

Note: This tool has been used locally to update yarn across our repos for the last few updates.


ng-dev/release/publish/actions.ts
```diff
@@ -200,26 +199,17 @@ export abstract class ReleaseAction {
 
   /**
    * Gets an owned fork for the configured project of the authenticated user. Aborts the
-   * process with an error if no fork could be found. Also caches the determined fork
-   * repository as the authenticated user cannot change during action execution.
+   * process with an error if no fork could be found.
    */
   private async _getForkOfAuthenticatedUser(): Promise<GithubRepo> {
-    if (this._cachedForkRepo !== null) {
-      return this._cachedForkRepo;
```
devversion, line 208: The old variable can be removed right?

devversion: it's still defined up in the class

ng-dev/utils/git/authenticated-git-client.ts
```diff
@@ -81,6 +85,30 @@ export class AuthenticatedGitClient extends GitClient {
...
+      return this._cachedForkRepo;
+    }
+
+    const {owner, name} = this.remoteConfig;
+    const result = await this.github.graphql(findOwnedForksOfRepoQuery, {owner, name});
+    const forks = result.repository.forks.nodes;
+
+    if (forks.length === 0) {
+      throw Error(
+        'Unable to find fork for currently authenticated user. Please ensure you created a fork ' +
```
devversion, line 103: We should rather return `null` as this is much easier to check and less ambiguous. Especially since we throw a non-unique error here it is difficult to filter out if some logic just failed by mistake vs. if just a fork could not be found.

ng-dev/misc/update-yarn/cli.ts
```diff
@@ -0,0 +1,125 @@
...
+import {error, info, red} from '../../utils/console';
+import {Spinner} from '../../utils/spinner';
+import {AuthenticatedGitClient} from '../../utils/git/authenticated-git-client';
+import {addGithubTokenOption} from '../../utils/git/github-yargs';
+
+async function builder(yargs: Argv) {
+  return addGithubTokenOption(yargs);
+}
+
+/** Handles the command. */
```
devversion, line 23: Nit: add a description explaining what the command exactly does?

ng-dev/misc/update-yarn/cli.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * New environment object with identifiers removed which prevent yarn from correctly loading
+     * new version from within a yarn call.
+     */
+    const env = {
+      ...process.env,
+      YARN_IGNORE_PATH: '0',
+      'PATH': process.env['PATH']!.replace(/\/tmp\/yarn.+?\:/, ''),
+    };
```
devversion, line 86: Why is this custom environment needed? I think this needs more explanation in the code and I'm not sure if this is actually needed. can you please help clarifying that?

ng-dev/misc/update-yarn/cli.ts
```diff
@@ -0,0 +1,125 @@
...
+      return;
+    }
+
+    /* File path of old yarn version */
+    const oldYarnPath = join(git.baseDir, `.yarn/releases/yarn-${yarnVersion.old}.cjs`);
+    const title = `build: update to yarn v${yarnVersion.new}`;
+    const body = `Update to the latest version of yarn, ${yarnVersion.new}.`;
+    const commitMessage = `${title}\n\n${body}`;
+    const {owner, repo} = git.remoteParams;
+    const branchName = `yarn-update-v${yarnVersion.new}`;
```
devversion, line 66: Super optional nit: To make this more readable below when consumed, I think we should explicitly add the `pr` prefix. 

```suggestion
    const prBranchName = `yarn-update-v${yarnVersion.new}`;
```

devversion (commented):
LGTM overall just a few nits

ng-dev/misc/update-yarn/cli.ts
```diff
@@ -0,0 +1,140 @@
...
+     * New environment object with identifiers removed which prevent yarn from correctly loading
+     * new version from within a yarn call.
+     */
+    const env = {
+      ...process.env,
+      YARN_IGNORE_PATH: '0',
+      // The /tmp/yarn/ path needs to be removed when running the `git commit` command, because the
+      // environment in which this script runs, was executed by yarn and yarn binary forks itself.
+      // But since the yarn script it referenced was removed by this script, we need to run yarn
+      // "fresh".
```
devversion, line 99: is that due to the husky hooks? e.g prepare-commit-msg? If so, I think we should just set HUSKY=0 similarly to how we do it in the merge and release tool

josephperrott: Yes it is, and I created this change initially before we started using `HUSKY=0` and had forgotten about it.

Will updated to using that.

devversion (approved):

PR #306, opened 2021-11-16
fix(github-actions): properly set the authentication token for feature request action

josephperrott:
The `Octokit` constructor was initially created with an older version of `Octokit` which supported
providing the token as a property `token`, the property is now expected to be `auth` to properly
use the token for authentication.


devversion (approved):

devversion:
The octokit types are pretty bad in that perspective

josephperrott:
This PR was merged into the repository by commit 912311af713004c92ddc3ac73eb0f43ceb55867f.

PR #307, opened 2021-11-17
refactor: ship `BUILD.bazel` files for shared scripts

devversion:
In order to simplify the use of the shared scripts within the dev-infra
repo itself, and within consumer projects, while also improving the
dependency managment and runfile tree creation, we start shipping the
actual `BUILD.bazel` files for the shared scripts.

This allows us to just continue using the shared-scripts Bazel targets
within Bazel tools, without needing any substitutions. And it also
ensures that the scripts will have the proper dependencies within Bazel
actions (e.g. by relying on a `js_library`). This also reduces the set
of unnecessary runfiles being brought in as the whole
`@angular/dev-infra-private` package and transitive deps wouldn't be
always needed.

This also works better within the FW repo where e.g.
`@angular/compiler` needs to be resolved through the Bazel NodeJS
linker. Running the scripts (like spec-bundling) directly in NodeJS
would not work as it wouldn't find the e.g. compiler.


josephperrott (approved):
LGTM

PR #310, opened 2021-11-18
refactor: add workaround for shared-scripts not working with linker

devversion:
Adds a small workaround for
https://github.com/bazelbuild/rules_nodejs/pull/3083.

Currently it is not possible to expose a `js_library` from an external
workspace and make it work with the linker. The JS library rule always
assumes that files are located in the `bazel-out/<..>/bin` directory.
This is not necessarily true and it would also be inefficient always
copying sources to the bazel-out when they originate from an external
workspace (although that is what we are doing now as a workaround).


josephperrott (approved):
LGTM

PR #316, opened 2021-11-25
feat(ng-dev/release): support experimental release packages

devversion:
Adds support for experimental release packages with the new release
tool. Experimental packages could already work partially by accident,
but there are issues when e.g. a new stable is cut an the NPM dist
tags are set for all packages. This surfaces in v13 for the CLI.

This commit requires an explicit setting for packages which are
supposed to be experimental and it replaces the logic for determining
an representative NPM package with an explicit configuration. This is
necessary because using an experimental package accidentally would break
the release tool down the line. The author of the release config will
know best which package can serve as representative, like
`@angular/core`, `@angular/cdk` or `@angular/cli` in all the Angular
main repositories.


devversion:
This is also a preparation/pre-requisite for us acutally running release output checks automatically as part of this tool..

gkalpak (approved):
Looks reasonable (although I am not too familiar with the release tooling 😅) :+1:

ng-dev/release/config/index.ts
```diff
@@ -77,13 +98,34 @@ export function assertValidReleaseConfig<T>(
...
   }
+
+  if (config.release.representativeNpmPackage && config.release.npmPackages) {
+    const representativePkgEntry = config.release.npmPackages.find(
+      (pkg) => pkg.name === config.release?.representativeNpmPackage,
+    );
+
+    if (representativePkgEntry === undefined) {
+      errors.push(
+        `Configured "representativeNpmPackage" does not match a package in "npmPackages".`,
```
josephperrott, line 118:
```suggestion
        `Configured "representativeNpmPackage" (${representativePkgEntry}) does not match a package in "npmPackages".`,
```

ng-dev/release/config/index.ts
```diff
@@ -77,13 +98,34 @@ export function assertValidReleaseConfig<T>(
...
+      );
+    } else if (representativePkgEntry.experimental) {
+      errors.push(
+        `Configured "representativeNpmPackage" corresponds to an experimental package. ` +
```
josephperrott, line 122:
```suggestion
        `Configured "representativeNpmPackage" (${representativePkgEntry}) corresponds to an experimental package. ` +
```

ng-dev/release/set-dist-tag/cli.ts
```diff
@@ -46,21 +55,36 @@ async function handler(args: Arguments<ReleaseSetDistTagOptions>) {
...
   debug(`Setting "${tagName}" NPM dist tag for release packages to v${version}.`);
   const spinner = new Spinner('');
 
-  for (const pkgName of npmPackages) {
-    spinner.update(`Setting NPM dist tag for "${pkgName}"`);
+  for (const pkg of npmPackages) {
+    // If `--skip-experimental-packages` is specified, all NPM packages which
+    // are marked as experimental will not receive the NPM dist tag update.
+    if (pkg.experimental && skipExperimentalPackages) {
+      continue;
```
josephperrott, line 75: I think we should log/update the spinner expressing that we skipped a package because its experimental

ng-dev/release/publish/external-commands.ts
```diff
@@ -36,8 +36,15 @@ import {FatalReleaseActionError} from './actions-error';
...
  * NPM dist tag for all packages in the checked out branch to the given version.
+ *
+ * Optionally, the NPM dist tag update can be skipped for experimental packages. This
+ * is useful when tagging long-term-support packages within NPM.
  */
-export async function invokeSetNpmDistCommand(npmDistTag: NpmDistTag, version: semver.SemVer) {
+export async function invokeSetNpmDistCommand(
+  npmDistTag: NpmDistTag,
+  version: semver.SemVer,
+  options: {skipExperimentalPackages: boolean} = {skipExperimentalPackages: false},
```
josephperrott, line 46: So for setting the npm tag we are set up to skip experimental packages for LTS, but when we cut an LTS version we still publish the package tagged as LTS.

If `@angular/package` is experimental, we publish `0.100.0`, then we publish `0.200.0` but `0.100.0` isn't tagged as `lts-v1`.  Then when we release `0.100.1` as LTS it would be published with `lts-v1` tag.

We should add this same logic to not publish experimental in `CutLongTermSupportPatchAction` as well.

devversion: I think for the sake of simplicity and keeping version numbers in sync, we should still publish such experimental packages in those cases. It should not have any negative impact, but to simplify our setup.

We are just omitting the LTS NPM dist tags so that we do not indicate any LTS guarantees / have people install the package through e.g @12-lts.

josephperrott: In `Action._publishBuiltPackageToNpm` we have:
```ts
await runNpmPublish(pkg.outputPath, npmDistTag, this.config.publishRegistry);
```

So when we publish an LTS (after the initial retag during the major release), we would begin to tag it as `{version}-lts`.  So I believe that we need to add this same skipping logic into the npm publishing process.  I might be misunderstanding what we are looking to avoid.

devversion: ah I see! good point. 

yeah, that is something we need to address. Maybe if we even go that far to exclude experimental NPM packages in such cases, should we also consider what I thought you proposed initially? 

i.e. we do not actually release the experimental package for an LTS? that would mean the package is not in sync with the other packages, but I'm trying to understand if that is acceptable/desirable? any thoughts?

ng-dev/release/set-dist-tag/cli.ts
devversion: It should rarely be readable anyway, but I'll add a message.

ng-dev/release/publish/external-commands.ts
devversion: updated

ng-dev/release/info/cli.ts
```diff
@@ -6,31 +6,61 @@
...
+}
+
 /** Yargs command handler for printing release information. */
-async function handler() {
+async function handler(argv: Arguments<ReleaseInfoOptions>) {
+  const config = getConfig();
+  assertValidReleaseConfig(config);
+
+  // If JSON output is requested, print the information as JSON to stdout.
+  if (argv.json) {
```
josephperrott, line 46: I am not quite sure how this makes sense to list within this command.  The JSON information being returned is entirely different than what is printed without the `--json` flag.

ng-dev/release/publish/external-commands.ts
```diff
@@ -36,8 +37,15 @@ import {FatalReleaseActionError} from './actions-error';
 /**
  * Invokes the `ng-dev release set-dist-tag` command in order to set the specified
  * NPM dist tag for all packages in the checked out branch to the given version.
+ *
+ * Optionally, the NPM dist tag update can be skipped for experimental packages. This
+ * is useful when tagging long-term-support packages within NPM.
  */
-export async function invokeSetNpmDistCommand(npmDistTag: NpmDistTag, version: semver.SemVer) {
+export async function invokeSetNpmDistCommand(
```
josephperrott, line 44: As I am looking at this, it seems odd that this is the place where we are making the decision about if an experimental package should be updated.

Does it make more sense to do this check before we invoke this command instead of invoking a command just to potentially tell it to skip itself.

ng-dev/release/publish/test/output-checks.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+
+import {parse, setupReleaseActionForTesting} from './test-utils/test-utils';
+import {DelegateTestAction} from './delegate-test-action';
+import * as console from '../../../utils/console';
+
+describe('package output checks', () => {
+  const baseReleaseTrains = new ActiveReleaseTrains({
+    releaseCandidate: null,
+    latest: new ReleaseTrain('13.0.x', parse('13.0.1')),
+    next: new ReleaseTrain('master', parse('13.1.0-next.2')),
```
josephperrott, line 22:
```suggestion
    next: new ReleaseTrain('main', parse('13.1.0-next.2')),
```

josephperrott (commented):
Commented with a few thoughts on things.


Additionally, as I was reviewing this I realized that we can't really land this change to not publishing LTS versions for experimental until `v13` LTSs.  We have already established that for `<=12` we publish experimental LTSs so it will be worse I think if we start to not do so partway through the cycle.

ng-dev/release/info/cli.ts
devversion: Yeah, I wasn't entirely sure about this, but given this is just a layer of communication for us, and it being strictly typed, I chose that way.

I initially considered another command like ng-dev release is-experimental.. but that is not very future-proof

ng-dev/release/publish/external-commands.ts
devversion: We need to do it there because the  configuration of experimental packages is local to the currently checked-out branch.

i.e. calling this command on an LTS branch should cause packages to be updated that e.g. no longer exist in master, and vice-verca

devversion:
@josephperrott I think we should be good even with current active LTS versions as we have not marked any package as experimental, and this change won't even go into LTS branches. That is actually the benefit of us keeping some operations local to the publish branches (with external commands)

ng-dev/release/publish/test/output-checks.spec.ts
devversion: 👍

ng-dev/release/publish/external-commands.ts
devversion: It's also worth noting that this command updates all packages in a branch by default, so filtering beforehand would not work because we either would need to skip the whole npm dist update, or update all (even experimental ones). I think this is the right way to proceed here.

devversion:
@josephperrott Can you please have another look at this? this is still pending your review/open comment.

josephperrott (approved):
LGTM

PR #317, opened 2021-12-01
fix(bazel): benchmark angular compilation pipeline not working with APF v13

devversion:
See individual commits


josephperrott (approved):
LGTM

PR #325, opened 2021-12-09
fix(ng-dev/release): do not error when yarn version of publish branch is older

devversion:
Currently there is one surprising issue within the release tool that
breaks the release of LTS versions.

If the ng-dev release tool is invoked through Yarn. e.g. `yarn ng-dev`,
Yarn will create a temporary directory and create wrappers for a binary
called `yarn` that will directly point to the checked-in Yarn file that
originally invoked the `ng-dev` tool. This means that child processes
calling to `yarn` will actually depend on the checked-in Yarn version
that invoked the `ng-dev` tool.

This logic works fine for most scripts/commands, but it breaks if the
release tool checks out an LTS branch and runs `yarn`. The checked-in
Yarn file from the original invocation may not exist anymore.

We fix this by resolving Yarn directly within the project, falling back
to the NPM global bin, or ultimately by leaving to the process `PATH`
standard resolution. This prevents this issue in all of our Angular
repositories where we check-in Yarn and have a `yarnrc` configuration.

### DIFF
--- a/ng-dev/misc/update-yarn/cli.ts
+++ b/ng-dev/misc/update-yarn/cli.ts
@@ -15,6 +15,7 @@ import {error, info, red} from '../../utils/console';
 import {Spinner} from '../../utils/spinner';
 import {AuthenticatedGitClient} from '../../utils/git/authenticated-git-client';
 import {addGithubTokenOption} from '../../utils/git/github-yargs';
+import {getYarnPathFromNpmGlobalBinaries} from '../../utils/resolve-yarn-bin';
 
 async function builder(yargs: Argv) {
   return addGithubTokenOption(yargs);
@@ -33,10 +34,16 @@ const skipHuskyEnv = {
 };
 
 async function handler() {
-  /** Directory where node binary are globally installed. */
-  const npmBinDir = spawnSync('npm', ['bin', '--global', 'yarn']).stdout.trim();
-  /** The full path to the globally installed yarn binary. */
-  const yarnBin = `${npmBinDir}/yarn`;
+  /**
+   * Process command that refers to the global Yarn installation.
+   *
+   * Note that we intend to use the global Yarn command here as this allows us to let Yarn
+   * respect the `.yarnrc` file, allowing us to check if the update has completed properly.
+   * Just using `yarn` does not necessarily resolve to the global Yarn version as Yarn-initiated
+   * sub-processes will have a modified `process.env.PATH` that directly points to the Yarn
+   * version that spawned the sub-process.
+   */
+  const yarnGlobalBin = (await getYarnPathFromNpmGlobalBinaries()) ?? 'yarn';
   /** Instance of the local git client. */
   const git = AuthenticatedGitClient.get();
   /** The main branch name of the repository. */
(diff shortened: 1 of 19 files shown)


ng-dev/release/publish/actions.ts
```diff
@@ -33,6 +33,7 @@ import {getPullRequestState} from './pull-request-state';
 import {getReleaseTagForVersion} from '../versioning/version-tags';
 import {GithubApiRequestError} from '../../utils/git/github';
 import {workspaceRelativePackageJsonPath} from '../../utils/constants';
+import {YarnCommandInfo} from '../../utils/resolve-yarn-bin';
```
josephperrott, line 36: Looks like this import is unused?

ng-dev/release/publish/external-commands.ts
```diff
@@ -37,17 +38,31 @@ import {FatalReleaseActionError} from './actions-error';
  * Invokes the `ng-dev release set-dist-tag` command in order to set the specified
  * NPM dist tag for all packages in the checked out branch to the given version.
  */
-export async function invokeSetNpmDistCommand(npmDistTag: NpmDistTag, version: semver.SemVer) {
+export async function invokeSetNpmDistCommand(
```
josephperrott, line 41: With 3 yarn commands here, as well as the `npm` commands (which should actually probably be yarn commands) in the release versioning files.  Does it make more sense to have a `YarnClient` similar to our `GitClient` rather than all of this repeated code?

This could also make tests involving the yarn pieces easier potentially.

josephperrott (commented):
I like the logic for resolving the yarn binary location, just the one question about how we utilize it throughout the tooling.

ng-dev/release/publish/external-commands.ts
devversion: I think there were some issues with publishing through Yarn, so we went with NPM (no idea why though)

Generally I like the idea of keeping it named around "external commands" in this super-dangerous file w/ a file comment. A class could wrap these methods yeah, but I don't think there is any benefit. Yarn would/should need to be re-resolved for every command invocation anyway, and a class has the risk of us caching some logic accidentally (and breaking when we switch branches).

Not sure if that make sense but this just my subjective feeling. This is probably something to revisit in a follow-up regardless?

ng-dev/release/publish/actions.ts
devversion: good catch

ng-dev/misc/update-yarn/cli.ts
```diff
@@ -33,10 +34,16 @@ const skipHuskyEnv = {
...
 
 async function handler() {
-  /** Directory where node binary are globally installed. */
-  const npmBinDir = spawnSync('npm', ['bin', '--global', 'yarn']).stdout.trim();
-  /** The full path to the globally installed yarn binary. */
-  const yarnBin = `${npmBinDir}/yarn`;
+  /**
+   * Process command that refers to the global Yarn installation.
+   *
+   * Note that intend to use the global Yarn command here as this allows us to let Yarn
```
josephperrott, line 40:
```suggestion
   * Note that we intend to use the global Yarn command here as this allows us to let Yarn
```

josephperrott (approved):
LGTM

ng-dev/misc/update-yarn/cli.ts
devversion: 👍

PR #328, opened 2021-12-13
fix(bazel): do not always treat parts of `app_bundle` rule as testonly

devversion:
This is a new difference that came up accidentally when the
ng_rollup_bundle rule has been reworked. A subset of targets
generated by the `app_bundle` macro is always testonly. This results
in unnecessary/unavoidable churn when the rule is used like in
Angular framework.


josephperrott (approved):
LGTM

PR #329, opened 2021-12-14
fix(bazel): unable to set tags for `app_bundle` rule due to conflict

devversion:
Fixes that no tags can be specified for `app_bundle` benchmark
targets due to a conflict with the debug target and its tags.

### DIFF
--- a/bazel/benchmark/app_bundling/index.bzl
+++ b/bazel/benchmark/app_bundling/index.bzl
@@ -65,7 +65,7 @@ def app_bundle(
         visibility = visibility,
     )
 
-    common_esbuild_options = dict(kwargs, **{
+    common_esbuild_options = {
         "config": "%s_esbuild_config" % name,
         "entry_point": entry_point,
         "target": target,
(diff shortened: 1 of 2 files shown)


bazel/benchmark/app_bundling/index.bzl
```diff
@@ -86,14 +86,13 @@ def app_bundle(
...
-        **common_esbuild_options
+        **dict(kwargs, **common_esbuild_options)
     )
 
     esbuild(
         name = "%s.debug" % name,
         args = _create_esbuild_minify_options(True),
-        tags = ["manual"],
-        **common_esbuild_options
+        **dict(kwargs, tags = ["manual"], **common_esbuild_options)
```
josephperrott, line 95: Do we want `tags` to apply here as well rather than always being just `manual`

josephperrott (approved):
LGTM

Just one question to confirm about

bazel/benchmark/app_bundling/index.bzl
devversion: yeah, I wasn't sure about that as well, but just left it like that because I think realistically the `debug` target should not receive any tags. If one adds e.g. `integration-test` to the app bundle target, then I'd expect the Bazel tag filter to only match the actual target and not the debug-one.

PR #330, opened 2021-12-15
fix(ng-dev): do not accidentally leak github tokens for errors

devversion:
NodeJS seems to print public properties of error instances. This
can result in the Github token for Git client errors to be printed.

We do not want the token to be printed as often team members might
escalate to dev-infra and just share the log output. In those cases
the token should be reducted to avoid unnecessary rotations.


crisbeto (approved):
LGTM

PR #340, opened 2022-01-10
feat(bazel): allow options to be passed into esbuild ng linker plugin for bazel

AleksanderBodurri:
Previously `createLinkerEsbuildPlugin` would invoke `createEs2015LinkerPlugin` without any options.

Now options passed into `createLinkerEsbuildPlugin` propagate into the internal invocation of `createEs2015LinkerPlugin`.

### DIFF
--- a/shared-scripts/angular-linker/esbuild-plugin.mjs
+++ b/shared-scripts/angular-linker/esbuild-plugin.mjs
@@ -47,13 +47,20 @@ async function assertNoPartialDeclaration(filePath, ast, traverseFn) {
  *   more here: https://esbuild.github.io/plugins/#filters.
  * @param ensureNoPartialDeclaration Whether an additional check ensuring there are
  *   no partial declarations should run.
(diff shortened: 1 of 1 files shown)


shared-scripts/angular-linker/esbuild-plugin.mjs
```diff
@@ -48,12 +48,13 @@ async function assertNoPartialDeclaration(filePath, ast, traverseFn) {
  * @param ensureNoPartialDeclaration Whether an additional check ensuring there are
  *   no partial declarations should run.
  */
-export async function createLinkerEsbuildPlugin(filter, ensureNoPartialDeclaration) {
+export async function createLinkerEsbuildPlugin(filter, ensureNoPartialDeclaration, options) {
```
devversion, line 51: Just some nit: Want to name this `linkerOptions` and maybe initialize the default directly here?

devversion (approved):

devversion (approved):

PR #361, opened 2022-01-27
fix(bazel/benchmark): switch to http-server for perf tests to support Apple M1

devversion:
See individual commits

### DIFF
--- a/bazel/benchmark/component_benchmark/defaults/index.ts
+++ b/bazel/benchmark/component_benchmark/defaults/index.ts
@@ -11,7 +11,7 @@ import 'zone.js/dist/zone';
 import {enableProdMode} from '@angular/core';
 import {platformBrowser} from '@angular/platform-browser';
 
-// @ts-ignore This file needs to be defined by the consumer.
+// @ts-ignore This file will be provided by the consumer.
 import {AppModule} from './app.module';
 
 enableProdMode();
--- a/bazel/benchmark/component_benchmark/defaults/index.ts
+++ b/bazel/benchmark/component_benchmark/defaults/index.ts
@@ -6,14 +6,13 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-// @ts-ignore Cannot find module
-import {enableProdMode} from '@angular/core';
+import 'zone.js/dist/zone';
 
-// @ts-ignore Cannot find module
+import {enableProdMode} from '@angular/core';
 import {platformBrowser} from '@angular/platform-browser';
 
-// @ts-ignore Cannot find module
-import {AppModuleNgFactory} from './app.module.ngfactory';
+// @ts-ignore This file needs to be defined by the consumer.
+import {AppModule} from './app.module';
 
 enableProdMode();
-platformBrowser().bootstrapModuleFactory(AppModuleNgFactory);
+platformBrowser().bootstrapModule(AppModule);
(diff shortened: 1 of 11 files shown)


bazel/benchmark/component_benchmark/defaults/index.ts
```diff
@@ -6,14 +6,13 @@
...
-import {enableProdMode} from '@angular/core';
+import 'zone.js/dist/zone';
 
-// @ts-ignore Cannot find module
+import {enableProdMode} from '@angular/core';
 import {platformBrowser} from '@angular/platform-browser';
 
-// @ts-ignore Cannot find module
-import {AppModuleNgFactory} from './app.module.ngfactory';
+// @ts-ignore This file needs to be defined by the consumer.
```
josephperrott, line 14:
```suggestion
// @ts-ignore This file will be provided by the consumer.
```

bazel/benchmark/component_benchmark/test/app.perf.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {element, By} from 'protractor';
+import {runBenchmark} from '../../driver-utilities/index';
+
+describe('app performance', () => {
+  it('should update on button press', async () => {
+    await runBenchmark({
```
josephperrott, line 14: This is more a comment on the component benchmarking tests I suppose, but we don't actually assert this does anything.

bazel/benchmark/driver-utilities/perf_util.ts
```diff
@@ -5,10 +5,14 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {mkdirSync} from 'fs';
+import {resolve} from 'path';
 
 export {verifyNoBrowserErrors} from './e2e_util';
 
+/** Absolute path to the directory where benchmark results should be written to. */
```
josephperrott, line 12:
```suggestion
/** Absolute path to the directory where benchmark results should be written. */
```

bazel/benchmark/driver-utilities/perf_util.ts
```diff
@@ -71,16 +75,13 @@ function createBenchpressRunner(): Runner {
   if (process.env.GIT_SHA) {
     runId = process.env.GIT_SHA + ' ' + runId;
   }
-  const resultsFolder = './dist/benchmark_results';
```
josephperrott, line 74: Is there a reason that we no longer need to create this directory?  I know that bazel should already provide the `TEST_UNDECLARED_OUTPUTS_DIR` value, but if it doesn't we need to make sure that the directory we use exists still correct?

devversion: good point. is it even worth caring about the non-bazel case when this utility is part of `/bazel`

josephperrott: Probably not, I think its fine just operate on the assumption that the environment value will be there.  Or to error if its not present instead of falling back.

bazel/benchmark/component_benchmark/test/app.perf.spec.ts
devversion: added an actual assertion test

josephperrott: 👍  Obviously outside of the scope of this PR, but should the component benchmark also include a requirement of providing an assertion that verifies the `work` performs the expected action?  Then `run_benchmark` could one time run the assertion to make sure the test passes, and then perform the actual benchmarks as it currently does.

josephperrott (approved):
LGTM

PR #362, opened 2022-01-27
test: ensure benchmark component rule test runs on CI

devversion:
Explicitly adds the component benchmark rule test to ensure
the tests run. Benchmark targets are marked as `manual` by
default.


josephperrott (approved):
LGTM

PR #363, opened 2022-01-27
feat(ng-dev): format json files by default with prettier

devversion:
Self-explanatory 😛


josephperrott (approved):
LGTM, will just need to have the generated file updated.

PR #367, opened 2022-01-29
Various improvements for Yarn 2 experiments

devversion:


.circleci/config.yml
```diff
@@ -8,11 +8,10 @@ orbs:
 # **Note**: When updating the beginning of the cache key, also update the cache key to match
 # the new cache key prefix. This allows us to take advantage of CircleCI's fallback caching.
 # Read more here: https://circleci.com/docs/2.0/caching/#restoring-cache.
-var_1: &cache_key v1-{{ checksum ".bazelversion" }}-{{ checksum "WORKSPACE" }}-{{ checksum "yarn.lock" }}
-# We want to invalidate the cache if the postinstall patches change. In order to apply new
-# patches, a clean version of the node modules is needed. Additionally, we invalidate the cache
-# if the Bazel version changes. We do this because otherwise the `bazelisk` cache folder will
-# contain all previously used versions and ultimately cause the cache restoring to be slower.
+var_1: &cache_key v1-{{ checksum ".bazelversion" }}
```
josephperrott, line 11: Why are we able to remove the `WORKSPACE` from the cache key?  That could still change right?

devversion: ops, that was an oversight. good catch

josephperrott (approved):
LGTM

PR #368, opened 2022-01-29
build: update to the latest version of angular benchpress

devversion:
Updates to the latest version of Angular benchpress. This one has
a proper peer dependency on v13. Note that we still use a Yarn
resolution as peer dependencies for NPM/Yarn will not match pre-releases
unless explicitly declared (for every possible minor then!)

Since we use pre-release versions of Angular in dev-infra, we want to
use such versions as the peer dependency of benchpress as well (with
the goal of reducing duplicated deps).

Relevant discussion on that note: https://stackoverflow.com/questions/67934358/npm-including-all-range-of-pre-release-when-defining-peer-dependency. I also experimented with that, but there seems no good way around that.


package.json
```diff
@@ -15,7 +15,7 @@
     "@actions/core": "^1.4.0",
     "@actions/github": "^5.0.0",
     "@angular-devkit/build-angular": "13.2.0-rc.1",
-    "@angular/benchpress": "^0.2.2",
+    "@angular/benchpress": "file:./angular-benchpress-v0.2.2.tgz",
```
josephperrott, line 18: Doesn't this mean we are using a local copy rather than NPM? I am just a little confused on why we need to do this part.

devversion: This is just temporary until the fix landed in framework. We can hold off for now

josephperrott: 👍  I ended up seeing that in a review later in my list.  Sounds good

devversion:
@josephperrott This should be ready for another pass, now that this PR is unblocked

josephperrott (approved):
LGTM

PR #369, opened 2022-01-30
build: update angular to stable version and remove range caret

devversion:
Updates Angular to stable v13.2.0 and removes the caret semver
range. This is necessary so that Renovate can properly update
the versions with `followTag: next` (in the renovate settings).

See discussion: https://github.com/renovatebot/renovate/issues/2864#issuecomment-441369513


crisbeto (approved):
LGTM

PR #371, opened 2022-01-30
refactor(bazel/benchmark): make it easier to substitute zone.js target with source target

devversion:
In the default entry-point index, `zone.js` is imported in a way that is not
checked by TypeScript. We add the dependency only for bundling to reduce the compilation
scope and to make it easier to replace this dependency inside the `angular/angular`
repository with its corresponding source target that does not come with any typings.

Replacing `@npm//zone.js` with a plain source file in the FW repo, or pointing
it to a `js_library` does not work because `ng_module`/`ts_library` would then
complain about a dependency that does not provide any typings. Hence the solution
to that problem is just adding zone.js as dependency for bundling (it's not needed
for the TS compilation anyway). In this repo, zone.js can be a simple dep for `ng_module`
or `ts_library` because it's covered as an `ExternalNpmPackage` target (which is an
exception and doesn't need any types to work..)

### DIFF
--- a/bazel/benchmark/component_benchmark/component_benchmark.bzl
+++ b/bazel/benchmark/component_benchmark/component_benchmark.bzl
@@ -39,7 +39,6 @@ def component_benchmark(
         ng_deps = [
             "@npm//@angular/core",
             "@npm//@angular/platform-browser",
-            "@npm//zone.js",
         ],
         ng_assets = [],
         assets = None,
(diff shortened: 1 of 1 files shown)


devversion:
Needed for https://github.com/angular/angular/pull/44830

crisbeto (approved):
LGTM

PR #382, opened 2022-02-03
build: disable yarn zero installs

devversion:
Disables Yarn zero installs since it's not really providing any
value given not using PnP.


josephperrott (approved):
LGTM

PR #384, opened 2022-02-04
feat(ng-dev): When checking out a PR, print the command to push changes.

dylhunn:
This is useful because the command for pushing the detached head back to the remote branch is a bit complicated and hard to remember.


ng-dev/pr/common/checkout-pr.ts
```diff
@@ -109,6 +109,7 @@ export async function checkOutPullRequestLocally(
   try {
     // Fetch the branch at the commit of the PR, and check it out in a detached state.
     info(`Checking out PR #${prNumber} from ${fullHeadRef}`);
+    info(`To push to this PR branch: git push ${GITHUB_SSH_PREFIX}:${pr.headRef.repository.nameWithOwner} HEAD:${headRefName} -f`);
```
dylhunn, line 112: I'm not sure it's generally correct to always use the SSH prefix?

devversion: I think not everybody uses SSH, and it's opt-in in the `ng-dev` config. I think it would make sense to either print both, or go with just showing the fork branch user and name

dylhunn: I updated it to print both URLs.

ng-dev/pr/common/checkout-pr.ts
```diff
@@ -109,6 +113,13 @@ export async function checkOutPullRequestLocally(
   try {
     // Fetch the branch at the commit of the PR, and check it out in a detached state.
     info(`Checking out PR #${prNumber} from ${fullHeadRef}`);
+    info(`To push to this PR branch:`);
+    info(
+      `  git push ${GITHUB_SSH_PREFIX}:${pr.headRef.repository.nameWithOwner} HEAD:${headRefName} -f`,
+    );
+    info(
+      `  git push ${GITHUB_URL_PREFIX}:${pr.headRef.repository.nameWithOwner} HEAD:${headRefName} -f`,
```
devversion, line 121: Does this URL work? I was assuming it should be `https://github.com/{nameWithOwner}<.git>` and optionally `.git` suffix. The colon seems to cause an invalid URL

ng-dev/utils/git/github-urls.ts
```diff
@@ -31,7 +37,7 @@ export function getRepositoryGitUrl(
   githubToken?: string,
 ): string {
   if (config.useSsh) {
-    return `git@github.com:${config.owner}/${config.name}.git`;
+    return `${GITHUB_SSH_PREFIX}:${config.owner}/${config.name}.git`;
   }
   const baseHttpUrl = `https://github.com/${config.owner}/${config.name}.git`;
```
devversion, line 42: Nit: you could potentially use the variable you introduced here and below as well, but not sure if it's even worth maintaing the variable. If you want to keep it though, it would make sense to use it consistenly I think

ng-dev/pr/common/checkout-pr.ts
```diff
@@ -109,6 +113,13 @@ export async function checkOutPullRequestLocally(
   try {
     // Fetch the branch at the commit of the PR, and check it out in a detached state.
     info(`Checking out PR #${prNumber} from ${fullHeadRef}`);
+    info(`To push to this PR branch:`);
+    info(
+      `  git push ${GITHUB_SSH_PREFIX}:${pr.headRef.repository.nameWithOwner} HEAD:${headRefName} -f`,
```
josephperrott, line 118: Personally, I don't think we should provide the ssh prefixed version.  The majority of our team now uses https, and we already know that it works for our users as the tooling uses https for its own pushes it performs.  While the ssh is definitely valid, it won't work unless already set up where as the http one we know should work.

---

We don't  want to instruct users to use `-f` in this case, they should use `--force-with-lease`.  I think our best bet will actually be to leverage the information we already have in the shared checkout-pr.ts utils.  We can have the checkout function provide the command to push back to the branch and then use that here.  This would allow us to maintain this expected command in just one place.

ng-dev/pr/common/checkout-pr.ts
josephperrott: Correct, I think this was a result of copy and paste from the ssh url above.

dylhunn:
Thanks for implementing this FR Joey!

PR #386, opened 2022-02-04
refactor: allow http-server to serve dotfiles

kyubisation:
It is currently not possible to serve dotfiles (e.g. .stackblitzrc) from the http-server. This PR changes the configuration to allow this.

We understand that this repository is only intended for internal Angular development and in this regard we completely understand if you reject this PR. Still we wanted to ask if you could accept this change, since this PR only changes a small configuration, without having any semantic impact.

### DIFF
--- a/bazel/http-server/server.ts
+++ b/bazel/http-server/server.ts
@@ -109,7 +109,7 @@ export class HttpServer {
         return;
       }
 
-      send(req, resolvedPath).pipe(res);
+      send(req, resolvedPath, {dotfiles: 'allow'}).pipe(res);
     }
   }
 


bazel/http-server/server.ts
```diff
@@ -34,6 +34,9 @@ export class HttpServer {
     server: {
       directory: false,
       middleware: [(req, res) => this._bazelMiddleware(req, res)],
+      serveStaticOptions: {
+        dotfiles: 'allow',
+      },
```
devversion, line 39: This option seems like a noop since we use a custom middleware anyway, and not the static serve one. Can you remove this please? (unless you see it being needed?)

devversion (approved):
LGTM.

I'm happy landing this because it matches with what the Angular CLI `ng serve` does, and IMO files explicitly part of the http-server `srcs` should be reachable.

bazel/http-server/server.ts
kyubisation: Quickly ran some tests, but you're certainly right. Removed it. 👍

devversion: thank you!

PR #387, opened 2022-02-05
build: tidy up dependencies in `package.json`

gkalpak:
This commit tidies up the dependencies in `package.json` in the following ways:
- Unused dependencies were removed.
- Missing dependencies were added.
- Dependencies were moved to the correct list (`dependencies`/`devDependencies`).

This ensures that no unnecessary dependencies will be installed by consumers and also that all needed dependencies will be present (without relying on the consumer to have them installed themselves).


package.json
```diff
@@ -35,9 +33,6 @@
     "@octokit/request-error": "^2.1.0",
     "@octokit/rest": "^18.7.0",
     "@octokit/types": "^6.16.6",
-    "@types/browser-sync": "^2.26.3",
-    "@types/send": "^0.17.1",
-    "@types/tmp": "^0.2.1",
```
devversion, line 40: I think all of these types are needed in the `http-server` which is built in consumer apps directly (so types are also needed 😄) We don't have a good strategy on Bazel-required NPM deps yet unfortunately.

devversion (approved):
LGTM in general, just one comment

package.json
```diff
@@ -12,75 +12,62 @@
...
-    "@bazel/bazelisk": "^1.10.1",
-    "@bazel/buildifier": "^4.0.1",
     "@bazel/concatjs": "4.6.0",
     "@bazel/esbuild": "4.6.0",
-    "@bazel/jasmine": "4.6.0",
     "@bazel/protractor": "4.6.0",
     "@bazel/runfiles": "4.6.0",
     "@bazel/terser": "4.6.0",
     "@bazel/typescript": "4.6.0",
+    "@google-cloud/storage": "^5.18.1",
```
devversion, line 25: I realize the `update-script` folder in `bazel/browsers` is technically not shipped to the NPM artifact. It's an internal script for us, so we could move it to dev dependencies. Let's just do that as a follow-up though.

devversion (approved):
LGTM, thx for cleaning this up. 

I'm a little unsure what impact this would have on the COMP repo which sometimes imports from deep files of `ng-dev` (now that some deps are no longer auto-installed). Though that is not a problem of this package. I think we might need to expose some "bundled" entry-point for private APIs we'd like to consume (like `fetchActiveReleaseTrains`)

PR #389, opened 2022-02-07
ci: group all scorecard action dependencies

alan-agius4:
With this change we group all the scorecard action dependencies so that Renovate opens a single PR.


josephperrott:
This PR was merged into the repository by commit 0f50da3dcfde727b8e4d26e1d61842f199016dad.

PR #392, opened 2022-02-07
fix(ng-dev/pr): do not fail assertions for changes for target label when the fixup label is applied

### LINKED ISSUE #301: merge tool fails before giving a chance to change the commit message
I'm trying to merge https://github.com/angular/components/pull/23714 which the author stated in the commit message is a `feat: ...`. However, the reviewer determined that it is not a feature, just a bug fix. They therefore applied the `target: patch` and `commit message fixup` labels to indicate that the message should be changed to say `fix: ...` and it can then be merged into a patch release. When I try to run the merge tool to actually do this, it dies, complaining that the `target: patch` doesn't match the `feat: ...` in the commit message. It does not give me a chance to change the commit message first as I would expect

```
$ ng-dev pr merge --no-branch-prompt 23714
ng-dev pr merge <pr>

Merge a PR into its targeted branches.

Positionals:
  pr  The PR to be merged.                                                                           [number] [required]

Options:
  --version        Show version number                                                                         [boolean]
  --github-token   Github token. If not set, token is retrieved from the environment variables.
                                                                                       [string] [default: <LOCAL TOKEN>]
  --help           Show help                                                                                   [boolean]
  --branch-prompt  Whether to prompt to confirm the branches a PR will merge into.             [boolean] [default: true]

PullRequestFailure {
  message: 'Cannot merge into branch for "target: patch" as the pull request has commits with the "feat" type. New features can only be merged with the "target: minor" or "target: major" label.',
  nonFatal: false
}
error Command failed with exit code 1.
```

josephperrott:
As noted in #301, if the commit message fixup label is applied we cannot properly assert
whether or not the changes are allowed for the target label.  Instead we should skip this
assertion.

Fixes #301

### DIFF
--- a/ng-dev/pr/common/validation/validations.ts
+++ b/ng-dev/pr/common/validation/validations.ts
@@ -10,7 +10,7 @@ import {Commit} from '../../../commit-message/parse';
 import {TargetLabel, TargetLabelName} from '../targeting/target-label';
 import {breakingChangeLabel, PullRequestConfig} from '../../config';
 import {PullRequestFailure} from './failures';
-import {red, warn} from '../../../utils/console';
+import {debug, red, warn} from '../../../utils/console';
 import {
   getStatusesForPullRequest,
   PullRequestFromGithub,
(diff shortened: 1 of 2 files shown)


ng-dev/pr/common/validation/validations.ts
```diff
@@ -28,7 +28,17 @@ export function assertChangesAllowForTargetLabel(
   label: TargetLabel,
   config: PullRequestConfig,
   releaseTrains: ActiveReleaseTrains,
+  labelsOnPullRequest: string[],
 ) {
+  if (
+    !!config.commitMessageFixupLabel &&
+    labelsOnPullRequest.some((name) => matchesPattern(name, config.commitMessageFixupLabel))
+  ) {
+    debug('Skipping assertion for changes from commits allowed for the target label because the');
```
devversion, line 37: Optional nit: This seems to read a little confusingly, maybe because I'm not a native-speaker. I think something like the following would be more clear:

```
Skipping commit message target label validation because the commit message fixup label is applied.
```

devversion (approved):
LGTM, just a subjective comment on the message, optional obviously 😄

josephperrott:
This PR was merged into the repository by commit 1c1d5c5d28f1c670ecdf9615c4e62a87f48151ee.

PR #395, opened 2022-02-08
Update browsers for Angular

devversion:


bazel/browsers/update-script/browser-artifact.ts
```diff
@@ -0,0 +1,50 @@
...
+  const separatorIdx = filePath.indexOf('.', startIdx);
+  const extension = filePath.substring(separatorIdx + 1);
+
+  if (separatorIdx === -1 || extension === '') {
+    throw new Error(`Unable to find known extension for file path: ${filePath}`);
+  } else if (KNOWN_EXTENSIONS.has(extension)) {
+    return extension;
+  }
+
+  return getArtifactExtension(filePath, separatorIdx + 1);
```
josephperrott, line 49: Should we leverage `path.extname` here?

```suggestion
  function getArtifactExtension(filePath: string): string {
    const fullFilePath = filePath;
    let extension = '';

    while (path.extname(filePath) !== '') {
      const extensionChunk = path.extname(filePath);
      extension = extensionChunk + extension;
      filePath = path.basename(filePath, extensionChunk);
    }

    if (KNOWN_EXTENSIONS.has(extension)) {
      return extension;
    } 

    throw new Error(`Unable to find known extension for file path: ${fullFilePath}`);
  }
```

josephperrott (approved):
LGTM

Just a couple suggestions/ideas.

bazel/browsers/update-script/browser-artifact.ts
devversion: Yeah, I thought using `indexOf` might be more clear, but `extname` handles the special case `a/b.c/d.tar.gz` better, so 👍 good idea!

PR #403, opened 2022-02-11
feat(bazel): setup karma web test wrapper with "local" debug target

devversion:
Sets up a karma web test wrapper with the "local" debug target that
is commonly set up in framework and the components repo. It seems
reasonable maintaining this logic in one place.


devversion:
We can wire up this in the test from #395

bazel/karma/index.bzl
```diff
@@ -0,0 +1,52 @@
...
+        kwargs["browsers"] = [
+            "//bazel/browsers/chromium:chromium",
+            "//bazel/browsers/firefox:firefox",
+        ]
+
+    # Filter out options which are specific to "karma_web_test" targets. We cannot
+    # pass options like "browsers" to the local web test target.
+    web_test_args = {}
+    for opt_name in kwargs.keys():
+        if not opt_name in ["wrapped_test_tags", "browsers", "wrapped_test_tags", "tags"]:
```
gkalpak, line 44: Did you mean to include `wrapped_test_tags` twice?

bazel/karma/index.bzl
```diff
@@ -0,0 +1,52 @@
...
+            web_test_args[opt_name] = kwargs[opt_name]
+
+    # Custom standalone web test that can be run to test against any browser
+    # that is manually connected to.
+    _karma_local_browsers_target(name = "%s_local" % name)
```
gkalpak, line 49: Nit: In my mind, `local` can be a little confusing. AFAICT, other targets could be "just as local" :grin:
It might be less confusing (for people that are not familiar with how Bazel/our setup works under the hood) to suffix this with `debug` or something.

gkalpak (approved):
Nice! ✨

bazel/karma/index.bzl
devversion: ops, just took this from Angular Components. definitely not intended

bazel/karma/index.bzl
devversion: yeah, I like that. `_local` matched Google internally IIRC, but I prefer debug as well. changing!

gkalpak (approved):
:tada:

PR #405, opened 2022-02-12
feat(ng-dev/format): add staged files back

dario-piotrowicz:
change the staged command so that after formatting it also adds
formatted files back to the staging area (unless --check is used
of course)
___

I just wanted to improve one small thing for development on the angular repo, basically when I create a commit and the pre-commit git hook formats the staged files, those changes aren't actually included in the commit, so I usually need to re-add the formatted files, update my commit and force push it.

I think that it is pretty inconvenient and it would just make sense for the hook to both format them and add them back to the staging area (I don't think this could be an unwanted behavior since the ci checks would anyways required the user to update those files).

So my idea is to just extend this:
https://github.com/angular/angular/blob/444354855b1e19ef383cf63088cbb31d59a14593/.husky/pre-commit#L5
to use this new `staged-re` command instead

___

Alternatively I could have added a flag as `--re-stage` or whatever but that wouldn't have made sense for the other format commands so I figured just a separate command could make sense

___

Also sorry for my boldness on opening a random PR here, I hope it's fine, if it is a bad idea you can always just shut it down anyways :smiling_face_with_tear: :bow:

### DIFF
--- a/ng-dev/format/cli.ts
+++ b/ng-dev/format/cli.ts
@@ -51,6 +51,9 @@ export function buildFormatParser(localYargs: yargs.Argv) {
         const executionCmd = check ? checkFiles : formatFiles;
         const allStagedFiles = GitClient.get().allStagedFiles();
         process.exitCode = await executionCmd(allStagedFiles);
+        if (!check && process.exitCode === 0) {
+          GitClient.get().runGraceful(['add', ...allStagedFiles]);
+        }
       },
     )
     .command(


devversion (commented):
I think this is a cool feature, definitely would have saved me some time in the past.

Just not sure about the API. I could imagine it being an optional flag for just `staged`.

cc. @josephperrott as the author of this command

dario-piotrowicz:
> I think this is a cool feature, definitely would have saved me some time in the past.

Thanks a lot!!! :heart_eyes: 

> 
> Just not sure about the API. I could imagine it being an optional flag for just `staged`.

I totally agree with you, I just did it by adding a new command because I am not too familiar with `yargs` and thought adding options to those chained commands was not possible :sweat_smile: :bow: , sorry I looked into it and updated the thing to have a `--add` flag for the addition of the formatted files to the staging area

thanks a lot for making me look into it! (and sorry for the first version :bow:)

PS: to keep the description short I put "add the __potentially__ formatted files" instead of saying that this option gets ignored if you use `--check`, what do you think? is it clear enough? :slightly_smiling_face:


> 
> Generally I believe formatted files should be staged automatically as this may interfere with the process of creating e.g. multiple commits based on certain criteria
> 

Same :relaxed::+1:

devversion:
No worries at all! I think both were valid approaches. It really comes down to preference and what might be best DX. I would definitely like to hear from @josephperrott about his thoughts.

> PS: to keep the description short I put "add the potentially formatted files" instead of saying that this option gets ignored if you use --check, what do you think? is it clear enough?

Interestingly I think there is a way in Yargs to indicate that something conflicts with another option. You may want to try that. Also should this option be applied to all sub-commands for `ng-dev format`? (that might be a good question to answer as well; e.g. `yarn ng-dev format changed --add`

For CI: You need to run `yarn update-generated-files` because we check-in some bundles into the repository for Github actions. Some github actions rely on code that you modified transitively.

dario-piotrowicz:
@devversion, I am not sure about putting the `--add` to all the format commands, I feel like it makes most sense for the `format staged`, but if you want I can also add it to the others :thinking: :man_shrugging:

ng-dev/format/cli.ts
```diff
@@ -46,11 +46,18 @@ export function buildFormatParser(localYargs: yargs.Argv) {
...
+        args.options('add', {
+          type: 'boolean',
+          description: 'Add the potentially formatted files back to the staging area',
+        }),
+      async ({check, add}) => {
         const executionCmd = check ? checkFiles : formatFiles;
         const allStagedFiles = GitClient.get().allStagedFiles();
         process.exitCode = await executionCmd(allStagedFiles);
+        if (!check && add && process.exitCode === 0) {
+          GitClient.get().stageFiles(allStagedFiles);
```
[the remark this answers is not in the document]
dario-piotrowicz: I _think_ that pre-commit hooks do allow us to add the files to the staging area _before_ the commit happens, thus including them in the commit itself

I'll check it out and let you know how it goes :slightly_smiling_face::+1:

ng-dev/utils/git/git-client.ts
```diff
@@ -197,6 +197,11 @@ export class GitClient {
     );
   }
 
+  /** Adds the provided files to the staging area. */
+  stageFiles(files: string[]) {
+    return gitOutputAsArray(this.runGraceful(['add', files.join(' ')]));
```
josephperrott, line 202: I think it makes more sense to return the command result.  Additionally allowing for the options to be passed to the `runGraceful` method to allow for suppressing output as desired.

Additionally, since the arguments are provided to `spawnSync` as an array, there is not a need to join them, we can just spread the files array into the args array.

```suggestion
  stageFiles(files: string[], options: GitCommandRunOptions = {}): SpawnSyncReturns<string> {
    return this.runGraceful(['add', ...files], options);
```

ng-dev/format/cli.ts
```diff
@@ -46,11 +46,19 @@ export function buildFormatParser(localYargs: yargs.Argv) {
...
+        args.options('add', {
+          type: 'boolean',
+          description: 'Add the potentially formatted files back to the staging area',
+          conflicts: 'check',
+        }),
+      async ({check, add}) => {
         const executionCmd = check ? checkFiles : formatFiles;
         const allStagedFiles = GitClient.get().allStagedFiles();
         process.exitCode = await executionCmd(allStagedFiles);
+        if (!check && add && process.exitCode === 0) {
```
josephperrott, line 59: @devversion what do you think of just always re-staging the files when `format staged` is executed? We would not run it during a check obviously.

```suggestion
        if (!check && process.exitCode === 0) {
```

ng-dev/format/cli.ts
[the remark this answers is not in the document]
dario-piotrowicz: Just adding them to the staging area does seem to do the trick for me:
Peek 2022-02-14 23-18

ng-dev/format/cli.ts
dario-piotrowicz: I've applied the change since I was anyways having issues passing the `--add` flag to the command in the bash pre-commit script (strangely enough it worked perfectly if called directly in the terminal, also `--check` worked just fine in the script :man_shrugging:)

if we don't want it I can always revert the change anyways :slightly_smiling_face: (and keep digging as to why `--add` wouldn't be recognized :sweat_smile: )

Anyways yeah I do agree that it could make sense to have this behavior, I am just wondering, is there any case in which this could be undesirable? (keeping in mind that we do still have the `--check` anyways)

dario-piotrowicz:
Reminder
- [ ] if/when the PR gets approved and it's ready to go I need to update the commit message

ng-dev/format/cli.ts
devversion: @josephperrott Sorry, didn't realize I got mentioned here again. I'm good with always staging (when not checking).

ng-dev/utils/git/git-client.ts
```diff
@@ -197,6 +197,11 @@ export class GitClient {
     );
   }
 
+  /** Adds the provided files to the staging area. */
+  stageFiles(files: string[], options: GitCommandRunOptions = {}): SpawnSyncReturns<string> {
```
devversion, line 201: This really a nit: I'd prefer to not add this here since it seems easy enough to invoke where needed. I don't think we intend to build a nice API around git, rather try to share commonly used, rather complicated code here. Not feeling strongly though (the method just limits the flexibility of the add command)

devversion (approved):
LGTM

ng-dev/utils/git/git-client.ts
devversion: I'm probably unhappy about this because we have other git add places already that are now left behind.

dario-piotrowicz: Sure I understand :slightly_smiling_face::+1:

I am not that familiar with this codebase, but I guess it could be nice to move some functionality in the git client, in case you'd like that just let me know and I would really like to try to move some stuff in the git client (consistently, without leaving stuff behind)

(although as you said these are pretty simple commands and `git`'s api as far as I can tell is pretty static, so there may not be too much benefit in that :shrug:)

josephperrott: I think I agree with Paul here, I would prefer we just put the staging logic at the call site.

If we were to create a generalized method here, I would prefer it to be `add()` since thats the actual thing that's called.

josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 74b89a46292330ba949652680c30b0280249db74.

devversion:
Thank you @dario-piotrowicz

ng-dev/utils/git/git-client.ts
dario-piotrowicz: yeah totally agree with the naming of course :slightly_smiling_face::+1:  (but yeah since it doesn't seem to be worth it anyways never mind :sweat_smile:)

dario-piotrowicz:
Thank you @devversion and @josephperrott for the awesome reviewing :smile::+1: 

___

(PS: please let me know if I need to do anything further to introduce this in the Angular repo :slightly_smiling_face:)

PR #408, opened 2022-02-14
fix(bazel/karma): web test arguments not passed through

devversion:
Fixes that the webtest arguments are not passed through to the
debug target. Also wires up the rule in our own tests as planned
initially (now that the test is merged)


josephperrott (approved):
LGTM

PR #416, opened 2022-02-20
fix(ng-dev): uncommitted changes check returning false-positives

devversion:
In some cases the `hasUncommittedChanges` check returns false-positives
that occur due to us not refreshing & cleaning up the index before diffing
the index. The index needs a cleanup/refresh when a file is simply touched/modified with the actual content not changing. i.e. just the last modified time got updated.

More details: https://stackoverflow.com/questions/34807971/why-does-git-diff-index-head-result-change-for-touched-files-after-git-diff-or-g/34808299#34808299

### DIFF
--- a/github-actions/commit-message-based-labels/main.js
+++ b/github-actions/commit-message-based-labels/main.js
@@ -49407,6 +49407,7 @@ var require_git_client = __commonJS({
         return branchName;
       }
       hasUncommittedChanges() {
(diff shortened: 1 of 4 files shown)


crisbeto (approved):
LGTM

devversion:
I fixed up the PR description since it read a little weird. Just FYI

PR #429, opened 2022-02-23
build: add darwin_arm64 support

jbedard:
May also require a fix or patch to the cc config: https://github.com/bazelbuild/bazel/issues/13514#issuecomment-847917936

I'm not sure if there's anything that should be added, or maybe we'd want to wait until that patch isn't required? But without this and the linked patch the build on an m1 doesn't work...

### DIFF
--- a/bazel/remote-execution/cpp/BUILD.bazel
+++ b/bazel/remote-execution/cpp/BUILD.bazel
@@ -11,6 +11,7 @@ cc_toolchain_suite(
     name = "cc_toolchain_suite",
     toolchains = {
         "k8": ":cc_compiler_k8",
+        "darwin_arm64": ":cc_compiler_k8",
     },
 )
 


devversion:
@jbedard In what cases would you need this to be built with arm64? This is the remote execution platform code that will run as `k8` anyway ([see here](https://github.com/angular/dev-infra/blob/main/bazel/remote-execution/BUILD.bazel#L6-L7)). Building these targets with Arm64, or Windows, MacOS fails currently yes, but that package shouldn't be built unless actually used.

gregmagolan:
What's the error you're getting on M1 @jbedard ?

gregmagolan:
@devversion Jason is working at Aspect and taking over the work the Dylan Martin was working on

gregmagolan:
Perhaps it is time to bring in a proper cpp toolchain into this repo so we can avoid the k8 hack for remote execution on macos.

I have an example of using the grails toolchain up here https://github.com/gregmagolan/bazel-llvm-cgo-cross-compile-example/blob/main/WORKSPACE#L39

jbedard:
```
ERROR: /.../angular/dev-infra/bazel/remote-execution/cpp/BUILD.bazel:10:19: Analysis of target '//bazel/remote-execution/cpp:cc_toolchain_suite' failed
ERROR: Analysis of target '//bazel/remote-execution/cpp:cc_toolchain_suite' failed; build aborted: 
```

Fixing this requires both the patch in the commit message and this change.

devversion:
As stated though, the CC toolchain in `bazel/remote-execution/cpp` is k8-only since that is what our `platform` in `bazel/remote-execution` provides. There is no need for the toolchain to support arm64, win64, darwin64 etc.

I think this error just surfaces due to the use of something like `bazel build //...`? We should rather exclude these targets from the expansion, or use something like `manual` (not sure if it works though)

devversion:
@gregmagolan I don't 100% remember, but I don't think the cpp toolchain is a hack. It's supposed to be a copy of the default k8 toolchain in Bazel. We just need to define it ourselves/copy it since it's not exposed in `bazel_tools` or elsewhere. The larger question might be if we need the cpp toolchain at all (which seems unrealistic given the project scope)

jbedard:
Tagging `bazel/remote-execution/cpp:cc_toolchain_suite` as manual works as well. Is that preferred?

devversion:
Yeah, I would be good with that

jbedard:
See #429

PR #430, opened 2022-02-24
Various improvements

devversion:


ng-dev/pr/merge/strategies/autosquash-merge.ts
```diff
@@ -122,3 +118,12 @@ export class AutosquashMergeStrategy extends MergeStrategy {
     return null;
   }
 }
+
+/** Gets the absolute file path to the commit-message filter script. */
+function getCommitMessageFilterScriptPath(): string {
+  // We resolve the script using module resolution as in the package output
+  // the worker might be bundled but exposed through a subpath export mapping.
+  return require.resolve(
```
josephperrott, line 126: Will this work within our own repository with `ts-node`?

josephperrott (approved):
Overall LGTM,  just one thing I wanted to confirm before we merge.

Long term, I would prefer that we aren't doing `export *` in our `ng-dev/index.ts` for everything.  Though we could set up a api golden snapshot test as well if we want to prevent accidentally adding things.

ng-dev/pr/merge/strategies/autosquash-merge.ts
devversion: In our repo it would find the `.js` output built from Bazel. It would also still work with `ts-node` though.

devversion:
@josephperrott Agreed on the export star thing. I was thinking about it as well, but it felt too cumbersome being super fine-grained about what is needed, and I imagined it being painful if more stuff is being needed (in relation to existing stuff being exposed). -- I like the golden though! 👍

ng-dev/pr/merge/strategies/autosquash-merge.ts
devversion: Actually, I now understand what you were referring to. Let me check.

devversion: Fixed with the latest commit.

josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit eaffaad11e60fda33b286d0c2ba5ce3b9d4274f6.

PR #431, opened 2022-02-25
build: fix workspace stamping breaking snapshot publishing

devversion:
We usually would use the `ng-dev` tool directly for stamping, but that one would need to be
built with Bazel and building with Bazel inside the Bazel-invoked workspace status
script does not work. To workaround this we have a `ts-node`-based stamping script.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -86,7 +86,11 @@ build:remote --remote_accept_cached=true
 ################################
 
 # Releases should always be stamped with version control info
-build:release --workspace_status_command="yarn ng-dev release build-env-stamp --mode=release"
+# Note: We usually would use the `ng-dev` tool directly, but that one would need to be
+# built with Bazel and building with Bazel inside the Bazel-invoked workspace status
+# script does not work. To workaround this, we have a `ts-node`-based stamping script.
+# TODO: Consider using `ng-dev` directly if this is possible.
+build:release --workspace_status_command="yarn build-env-stamp --mode=release"
 build:release --stamp
 
 ####################################################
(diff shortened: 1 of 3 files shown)


.bazelrc
```diff
@@ -86,7 +86,10 @@ build:remote --remote_accept_cached=true
 ################################
 
 # Releases should always be stamped with version control info
-build:release --workspace_status_command="yarn ng-dev release build-env-stamp --mode=release"
+# Note: We usually would use the `ng-dev` tool directly, but that one would need to be
+# built with Bazel and building with Bazel inside the Bazel-invoked workspace status
+# cript does not work. To workaround this we have a `ts-node`-based stamping script.
+build:release --workspace_status_command="yarn build-env-stamp --mode=release"
```
josephperrott, line 92: Can we have a TODO that specifically notes that we would like to not have to have this workaround?

.bazelrc
```diff
@@ -86,7 +86,10 @@ build:remote --remote_accept_cached=true
 ################################
 
 # Releases should always be stamped with version control info
-build:release --workspace_status_command="yarn ng-dev release build-env-stamp --mode=release"
+# Note: We usually would use the `ng-dev` tool directly, but that one would need to be
+# built with Bazel and building with Bazel inside the Bazel-invoked workspace status
+# cript does not work. To workaround this we have a `ts-node`-based stamping script.
```
josephperrott, line 91:
```suggestion
# script does not work. To workaround this us a `ts-node`-based stamping script.
```

josephperrott (approved):
LGTM

.bazelrc
devversion: Added. I don't think there is a way to achieve this without any weird tricks, but good to have it on the radar 👍

PR #446, opened 2022-03-02
build: remove unused managed directory bazel setup for node modules

devversion:
As of RNJ v5, the node modules are no longer symlinked by default. Rather
a new separate install is maintained inside Bazel. This was a little
controversial in the past since it made debugging harder, and required
more IO locally/or even downloads depending on how Yarn is cached/configured
locally.

With the current setup, symlinks are no longer enabled and we can remove
the managed directory configuration.

https://github.com/bazelbuild/rules_nodejs/pull/3214

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,6 +1,5 @@
 workspace(
     name = "dev-infra",
-    managed_directories = {"@npm": ["node_modules"]},
 )
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_file")


devversion:
@josephperrott this is something I'm not 100% sure about yet. Debugging and all this is also quite important. I also like it being better cached/encapsulated in bazel though (which seems like the reason it's no longer the default).

cc. @gregmagolan if you have other arguments on why symlinks should be avoided.

gregmagolan (approved):

josephperrott:
I am completely on the unsure of the best way to go with it.

Having the node_modules symlinked more matches the node ecosystem typical experience, but that also means that we lose some of the "guarantees" we have with bazel about being the same environment on every run regardless of machine.

I wonder if we can do more investigating to see what this would do in other repositories than dev-infra since they more heavily try to modify things and such?

devversion:
@josephperrott I tried to play with this in the COMP repo. I don't think we will run into any issues with maintaining another node module install. Once `@npm//` is fetched, the timing/tracing seems almost identical. The only major downsides I can see:

* Potential slow install when node modules need to be fully re-installed again twice. The global Yarn cache is still active here though, so it's not too bad (eventually with Yarn berry this could be a noop)
* Debugging of Bazel code being a little more advanced (devs need to know of this second copy; that seems not too bad though as I think we are quite familiar with Bazel)

I'm still super on the fence, but maybe it's best keeping it as is for now and revisiting with Yarn berry in the future/or when `rules_nodejs` stops supporting symlinks. @gregmagolan any comment on whether that could happen?

josephperrott:
I think that if we don't have a specific target date yet for when its going to stop being supported then we should leave it for now.  As you noted if we were to move to yarn berry before this it would end up being a noop.  I think it makes sense to continue as is for now instead of adding another moving piece.

PR #449, opened 2022-03-04
fix(ng-dev/release): only match on a single character prefixed version for snapshotting

josephperrott:
Previously we matched on all characters before our semver matching in the glob used for finding
the latest version for the branch we are on.  However when the branch contains something like
`zonejs-1.2.3`, the `*` matcher would capture it, leading to an unexpected syntax.  Instead we should
allow for only a single letter, typically `v`, to prefix our version tags.

cc: @AndrewKushnir

### DIFF
--- a/ng-dev/release/stamping/env-stamp.ts
+++ b/ng-dev/release/stamping/env-stamp.ts
@@ -63,7 +63,10 @@ function getSCMVersions(
       const {stdout: rawVersion} = git.run([
         'describe',
         '--match',
-        '*[0-9]*.[0-9]*.[0-9]*',
+        // As git describe uses glob matchers we cannot the specific describe what we expect to see
+        // starting character we expect for our version string.  To ensure we can handle 'v'
+        // prefixed verstions we have the '?' wildcard character.
+        '?[0-9]*.[0-9]*.[0-9]*',
         '--abbrev=7',
         '--tags',
         'HEAD',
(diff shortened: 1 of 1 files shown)


ng-dev/release/stamping/env-stamp.ts
```diff
@@ -83,7 +86,8 @@ function getSCMVersions(
       const {version: experimentalVersion} = createExperimentalSemver(new SemVer(version));
       return {version, experimentalVersion};
     }
-  } catch {
+  } catch (e) {
+    console.error(e);
```
devversion, line 90: will that work with stamping? or will Bazel "swallow" the message? should we just throw as well?

devversion (approved):

ng-dev/release/stamping/env-stamp.ts
josephperrott: My instinct is to say that we should throw here, but I am not sure if I missing something with respect to what this could break if we throw.  Or rather what it could break unintentionally.

devversion: I think it would fail the build? (can be easily tested). If that"s the case, I'd propose going with that for better erroring/stamping

josephperrott: Yeah, removing the try catch results in a good error message even from within bazel.

```
ng-dev release build-env-stamp

Build the environment stamping information

Options:
  --version  Show version number                                                                               [boolean]
  --help     Show help                                                                                         [boolean]
  --mode     Whether the env-stamp should be built for a snapshot or release [required] [choices: "snapshot", "release"]

TypeError: Invalid Version: zone.js-0.11.5
```

devversion (approved):

PR #450, opened 2022-03-05
refactor(ng-dev): expose git and github client classes in ng-dev entry-point

devversion:
These classes will be needed for interacting with the release versioning APIs.

### DIFF
--- a/ng-dev/index.ts
+++ b/ng-dev/index.ts
@@ -22,6 +22,12 @@ export * from './release/versioning';
 // to the dev-infra log which is stored on failures.
 export * from './utils/console';
 
+// Exposes Git/Github client classes needed for interacting with some of the
+// release versioning APIs.
+export * from './utils/git/authenticated-git-client';
+export * from './utils/git/git-client';
+export * from './utils/git/github';
+
 // Additional exports for adding custom release pre-staging, post-build checks.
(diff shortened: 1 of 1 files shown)


ng-dev/index.ts
```diff
@@ -22,6 +22,12 @@ export * from './release/versioning';
 // to the dev-infra log which is stored on failures.
 export * from './utils/console';
 
+// Exposes Git/Github client classes needed for interacting with some of the
+// release versioning APIs.
+export * from './utils/git/authenticated-git-client';
```
josephperrott, line 27: Can change this to expose the specific symbols we want rather than all?

josephperrott (approved):
LGTM

ng-dev/index.ts
devversion: Yeah, could do. Just not sure if it's worth maintaining that list. It's already quite unclear what specific APIs e.g. the versioning API transitively requires 😓

devversion: Let me know, happy to go with either approach (star would be consistent with the file curently)

josephperrott: We can leave it as is, I am surprised we aren't able to just provide the specific ones here as for instance `./utils/git/authenticated-git-client` only has two exported symbols.

PR #451, opened 2022-03-05
feat(ng-dev): support skipping version stamp in release stamp env

devversion:
Some repositories like the `dev-infra` repository do not follow the
usual Angular SemVer versioning but still would like to use the shared
stamping script.

For these repositories, another option is added to the stamping command
that allows version info from being included. There needs to be a way
to disable this as otherwise stamping currently fails when it tries to
find/compute a SemVer version from the workspace `package.json`, causing
errors like:

```
TypeError: Invalid Version: 0.0.0-{SCM_HEAD_SHA}
```

### DIFF
--- a/ng-dev/release/stamping/env-stamp.ts
+++ b/ng-dev/release/stamping/env-stamp.ts
@@ -10,21 +10,14 @@ import {join} from 'path';
 import {SemVer} from 'semver';
 import {GitClient} from '../../utils/git/git-client';
 import {createExperimentalSemver} from '../../release/versioning/experimental-versions';
+import * as fs from 'fs';
 
 export type EnvStampMode = 'snapshot' | 'release';
 
-/**
- * Log the environment variables expected by bazel for stamping.
- *
- * See the section on stamping in docs / BAZEL.md
- *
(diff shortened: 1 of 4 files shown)


ng-dev/release/stamping/env-stamp.ts
```diff
@@ -51,40 +48,39 @@ function hasLocalChanges(git: GitClient) {
...
-      // prefixed verstions we have the '?' wildcard character.
-      '?[0-9]*.[0-9]*.[0-9]*',
-      '--abbrev=7',
-      '--tags',
-      'HEAD',
-    ]);
-    const {version} = new SemVer(rawVersion);
-    const {version: experimentalVersion} = createExperimentalSemver(version);
+  if (mode === 'release') {
+    const workspaceVersion = getVersionFromWorkspacePackageJson(git);
```
devversion, line 58: This diff here is quite a lot, but only the `mode === 'release'` case got updated, and I swaped the blocks for better readability. Feel free to ignore the rest.

crisbeto (approved):
LGTM

PR #455, opened 2022-03-08
build: update to typescript v4.6 to match with FW repo

devversion:
As we start to update TS in all other Angular repos, we should
make sure the shared tooling is also in sync with regards to the
TS version (https://github.com/angular/angular/commit/94bba76a4a9594a5eb90e581f407f1b70697e715)


crisbeto (approved):
LGTM

PR #456, opened 2022-03-08
refactor: no longer import `JSModuleInfo` provider through backwards-compat entry

devversion:
The `JSModuleInfo` provider is no longer part of the `build_bazel_rules_nodejs` repository
but rather part of the "rules_nodejs" fundamentals. We should no longer rely on the
backwards-compatibility exposed provider, but instead use the expected new import.


josephperrott (approved):
LGTM

PR #457, opened 2022-03-08
feat(ng-dev): format yaml files with prettier by default

devversion:
This should help making editing of the very few Yaml files (like the CircleCI config) easier 😄


josephperrott (approved):
LGTM

PR #458, opened 2022-03-09
build: enable node module symlinking for Bazel

devversion:
Enables symlinking of the node modules folder to keep a single
install of NPM/Yarn. More details can be found here:

https://github.com/angular/dev-infra/pull/446#issuecomment-1059820287

Symlinking previously was enabled by default, but as part of RNJ 5 this
has changed without any super-noticable changelog entry.


josephperrott (approved):
LGTM

PR #460, opened 2022-03-09
ci: fix yarn install timeout in macos job

devversion:
https://app.circleci.com/pipelines/github/angular/dev-infra/164070/workflows/cc6cd9a7-5e0f-450c-b839-ece856156796/jobs/165678

@josephperrott I'm proposing to just always set the timeout to 5min (windows takes 3min fresh as it seems), so we could use the shared command and avoid some complexity in the config (as macos 2min also is too short)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,7 @@ var_5: &save_cache
   save_cache:
     key: *cache_key
     paths:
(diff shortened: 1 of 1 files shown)


josephperrott (approved):
LGTM

PR #461, opened 2022-03-09
ci: fix cache key to also include `yarn.lock`

devversion:
Looks like the cache key did not contain the `yarn.lock`, which I
assumed it had when I re-enabled the Yarn cache in a previous commit.


josephperrott (approved):
LGTM

PR #467, opened 2022-03-15
ci: ignore integration test projects in renovate

devversion:
Renovate is never able to fully/properly update integration
test projects as they are technically invalid (due to e.g.
`file:../../whatever` version installs).

We should just skip these tests as they should rarely need
updates anyway, and since that makes Renovate happy and not
complain about broken updates/invalid artifacts.


josephperrott (approved):
LGTM

PR #472, opened 2022-03-17
fix(ng-dev): avoid runtime error when pull request does not have status/checks

devversion:
See individual commits


josephperrott (approved):
LGTM

PR #473, opened 2022-03-18
build: remove workaround for rules_nodejs#3106 which landed in 4.5.0

devversion:
https://github.com/bazelbuild/rules_nodejs/pull/3106


josephperrott (approved):
LGTM

PR #475, opened 2022-03-19
fix(bazel): use correct `LinkerPackageMappingInfo` field name

gkalpak:
Use the correct field name (`node_modules_roots` instead of `node_module_roots`) when trying to access data off of the [LinkerPackageMappingInfo][1] value of a dependency in `extract_js_module_output()`.

See [here][2] for an example of a failure with the current incorrect field name.

[1]: https://github.com/bazelbuild/rules_nodejs/blob/2a78e6d815ee36e99cf9cc6462419e16f11c057d/internal/linker/link_node_modules.bzl#L34
[2]: https://circleci.com/gh/angular/angular/1137996


devversion (approved):
Good catch. This property is named a little weirdly -- likely we should set up a test or something. I will make a note

PR #476, opened 2022-03-19
test: add basic tests to ensure `extract_types` and `extract_js_module_output` works

devversion:
Self-explanatory.

Note: I want to clean up the extract JS module rule as I find it being concerned with _too_ much, but that's
another story. This commit just adds a little test ensuring it does what it suggests + continues to work.

### DIFF
--- a/bazel/test/BUILD.bazel
+++ b/bazel/test/BUILD.bazel
@@ -1,0 +1,57 @@
+load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_test")
+load("@npm//@bazel/concatjs:index.bzl", "ts_library")
+load("//bazel:extract_types.bzl", "extract_types")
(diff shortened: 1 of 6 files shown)


bazel/test/BUILD.bazel
```diff
@@ -0,0 +1,57 @@
...
+)
+
+extract_js_module_output(
+    name = "extract_js_module_output_target",
+    testonly = True,
+    forward_linker_mappings = True,
+    include_declarations = True,
+    include_default_files = True,
+    include_external_npm_packages = True,
+    provider = "JSModuleInfo",
```
devversion, line 38: It's worth noting that we won't capture all possible code-paths here (e.g. all combinations of the target definitions), but that's out-of-scope and we are mostly interesting in the provider validation/analysis phase of Bazel

gkalpak (approved):
:+1:

PR #480, opened 2022-03-21
feat(bazel): api-golden test should leverage package exports information for finding entries

devversion:
See individual commits


bazel/api-golden/test/goldens/test_package/index.md
```diff
@@ -0,0 +1,12 @@
+## API Report File for "test-package"
+
+> Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
+
+```ts
+
+// @public (undocumented)
+export const fromTopLevel = true;
```
josephperrott, line 8: Shouldn't these be types?

devversion: It is, for a constant the actual literal value is the concrete type

josephperrott: I thought it might be that, can we updated one of the symbols to be something that will show as a type here for better illustration?

josephperrott (approved):

bazel/api-golden/test/goldens/test_package/index.md
devversion: good idea, also adding `export declare` since that is what `d.ts` files usually have when emitted by TSC

PR #484, opened 2022-03-24
fix(bazel): continue to support non-exports resolution in api-golden test

devversion:
This is needed since the CLI does not use exports yet. See:
https://github.com/angular/angular-cli/issues/22889


alan-agius4 (approved):

alan-agius4:
Thanks for this @devversion

PR #485, opened 2022-03-25
Create initial base of apps infrastructure and framework

josephperrott:
See individual commits.


apps/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
...
+        "//apps/functions:functions_files",
+    ],
+    entry_point = "@npm//:node_modules/firebase-tools/lib/bin/firebase.js",
+    templated_args = [
+        "--project",
+        "angular-caretaker",
+        "--config",
+        "$$(rlocation $(execpath :firebase.json))",
+        "emulators:start",
+    ],
```
devversion, line 21: That's great! do we know if this will work on Windows? Not that is a strict requirement here, but that would certainly be nice I think.

devversion: I guess it works since you used `copy_to_bin`? Nice --  I think we should try avoid using this unless really needed (refactoring some FW tests for this -- [see details here](https://angular-team.slack.com/archives/C02PARQNMC1/p1648137933069659)). I think in this case there might be no way around it though, but a comment might be good to mention e.g. Windows?

apps/functions/BUILD.bazel
```diff
@@ -0,0 +1,36 @@
...
+        "@npm//firebase-functions",
+    ],
+)
+
+esbuild(
+    name = "functions_compiled",
+    entry_points = [
+        "index.ts",
+    ],
+    format = "cjs",
```
devversion, line 30: Should we use `ESM` here? looks like firebase tools now supports this and long-term we want to use ESM everywhere. Doing it now would make the migration easier in the future then (CJS -> ESM is never easy)

https://github.com/firebase/firebase-tools/releases/tag/v9.15.0

apps/functions/githubWebhook/index.ts
```diff
@@ -0,0 +1,29 @@
+import * as functions from 'firebase-functions';
+import * as admin from 'firebase-admin';
+import {handlePullRequestEvent} from './pull-request';
+
+admin.initializeApp({...functions.firebaseConfig()});
+
+export const githubWebhook = functions
```
devversion, line 7: Nit: I think every function index should have a small little doc outlining the goal

apps/shared/models/pull-request.ts
```diff
@@ -0,0 +1,88 @@
...
+  node: string;
+  state: PullRequest['state'];
+  authorAssociation: PullRequest['author_association'];
+  body: string;
+  changedFiles: number;
+  closedAt: string | null;
+  commits: number;
+  createdAt: string;
+  draft: boolean;
+  labels: string[];
```
devversion, line 20: Can we denote these as actual references somehow? using e.g. branded strings? I think it's easy to confuse this with an actual label name right? same for other foreign "values/keys"

apps/shared/models/server-models.ts
```diff
@@ -0,0 +1,68 @@
...
+  converter: AdminFirestoreDataConverter<Model>;
+};
+type WithAdminFirestoreDataConverterAndGithub<Model, GithubModel> = Model & {
+  converter: AdminFirestoreDataConverter<Model> & {
+    fromGithub: (model: GithubModel) => Model;
+    getFirestoreRefForGithubModel: (model: GithubModel) => AdminDocumentReference;
+  };
+};
+
+export function forServer<
```
devversion, line 36: I like the architecture you planned for the models. It looks great, and actually is very easy to follow. I think we could make this perfect by having small little comments w/ TSDoc for `forApp`, `forServer` etc. wdyt?

apps/shared/models/team.ts
```diff
@@ -0,0 +1,34 @@
...
+      return {
+        name: model.name,
+        privacy: model.privacy,
+        slug: model.slug,
+      };
+    },
+  };
+}
+
+export function isTeam(userOrTeam: User | Team): userOrTeam is Team {
```
devversion, line 29: Nit: Suggestion to make it obvious this is for the github webhook payload identification (same for `isUser`)

```suggestion
export function isTeamPayload(userOrTeam: User | Team): userOrTeam is Team {
```

devversion (approved):
Nice, nice! 🤩 just some minor comments, but this looks fantastic to me

apps/shared/models/team.ts
josephperrott: I went with `is`<Type>FromGithub`, Its not actually the payload from a user or team event, as the object appears throughout the webhook payloads when the objects are present.

apps/shared/models/base.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Because it is less expensive to store strings than Firestore references, with no change in the
+ * value it provides, we have this local type to express this usage.
+ */
+export type FirestoreReference<T> = string;
```
devversion, line 5: Nit: Should we use a branded type so that we can make this type-safe. Can be a follow-up

```suggestion
export type FirestoreReference<T> = {
     // ensures this "Reference" string is not passed around as "string" accidentally.
  __eefirestoreReference: true;
};

export function getValueOfReference<T>(ref: FirestoreReference<T>): T {
  // we know this is a string.
  const value = ref as string;
  
  // do something
}
```

devversion (approved):

devversion:
@josephperrott I wanted to merge this but the fixup commits seem to be not squash-able.

apps/shared/models/base.ts
josephperrott: I played with this a bit and I think we should do it in a follow up as it makes the question of how we do some of the nested interactions a bit tricky and I think when we are in a place of more practical decisions being made it will be clearer.

josephperrott:
@devversion manually squashed and pushed up the new commits.

PR #488, opened 2022-03-26
fix(ng-dev): lock formatting dependencies to avoid formating failures

devversion:
Due to us not locking prettier, buildifier etc, downstream projects
might end up running into formatting failures when e.g. lock file
maintance is performed. The underlying formatting tools should only
update along with dev-infra, so that it's obvious when something
related to formatting changes.


crisbeto (approved):
LGTM

PR #489, opened 2022-03-26
feat(ng-dev/pullapprove): handle references to `author` in condition

devversion:
Necessary change for: https://github.com/angular/angular/pull/44866


josephperrott (approved):
LGTM

Nice, I had completely forgotten about wanting to do this in framework and needing to make these changes.

PR #491, opened 2022-03-28
feat(ng-dev/pullapprove): support `in` operator in condition evaluation

devversion:
A few conditions in framework currently rely on the `in` operator. We
should support this even though it potentially going away when we
re-work how global approvals are performed. This makes the verify command
a little more future-proof and currently unblocks the dev-infra update.

It's worth noting that our Python -> JS conversion is very limited but
we can always expand when needed, or start running actual Python. Right
now this captures the most common cases of conditions we have seen across
Angular.


alan-agius4 (approved):
LGTM

PR #502, opened 2022-04-07
Create account menu and login page

josephperrott:
See individual commits.

### DIFF
--- a/apps/prs/src/app/app.component.html
+++ b/apps/prs/src/app/app.component.html
@@ -1,4 +1,7 @@
-<mat-toolbar color="primary" class="app-header-bar"></mat-toolbar>
+<mat-toolbar color="primary" class="app-header-bar">
+  <span class="spacer"></span>
+  <account></account>
+</mat-toolbar>
 <mat-sidenav-container class="sidenav-container">
   <mat-sidenav #drawer class="sidenav" mode="side">
     <!-- TODO(josephperrott): Establish need and content for sidenav once appropriate. -->
--- a/apps/prs/src/app/app.component.scss
+++ b/apps/prs/src/app/app.component.scss
@@ -2,6 +2,11 @@
   position: sticky;
   top: 0;
   z-index: 1;
+  flex-direction: row;
+
+  .spacer {
+    flex: 1;
+  }
 }
 
 .sidenav {
--- a/apps/prs/src/app/login/login.component.scss
+++ b/apps/prs/src/app/login/login.component.scss
@@ -1,0 +1,6 @@
+:host {
+  .mat-card {
+    width: 500px;
+    margin: 150px auto 0;
+  }
+}
--- a/apps/shared/account/BUILD.bazel
+++ b/apps/shared/account/BUILD.bazel
@@ -1,0 +1,33 @@
+load("//tools:defaults.bzl", "ng_module")
+load("@io_bazel_rules_sass//:defs.bzl", "sass_binary")
+
+package(default_visibility = ["//visibility:private"])
+
+ng_module(
+    name = "account",
+    srcs = glob(
+        [
+            "*.ts",
+        ],
+        exclude = [
+            "*.spec.ts",
+        ],
+    ),
+    assets = [
+        "account.component.css",
+        "account.component.html",
+    ],
+    deps = [
+        "@npm//@angular/cdk",
+        "@npm//@angular/common",
+        "@npm//@angular/core",
+        "@npm//@angular/fire",
+        "@npm//@angular/material",
+        "@npm//rxjs",
+    ],
+)
+
+sass_binary(
+    name = "account-style",
+    src = "account.component.scss",
+)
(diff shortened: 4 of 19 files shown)


apps/shared/account/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
...
+    srcs = glob(
+        [
+            "*.ts",
+        ],
+        exclude = [
+            "*.spec.ts",
+        ],
+    ),
+    assets = [
+        "account.component.css",
```
devversion, line 17: Optional nit (if you agree): I think generated files should be specially-denoted (not feeling strongly)

```suggestion
        ":account.component.css",
```

apps/prs/src/app/app.component.scss
```diff
@@ -2,6 +2,11 @@
   position: sticky;
   top: 0;
   z-index: 1;
+  flex-direction: row;
```
devversion, line 5: This shouldn't be needed as `mat-toolbar` sets it automatically?

apps/prs/src/app/login/login.component.scss
```diff
@@ -0,0 +1,6 @@
+:host {
```
devversion, line 1: Nit: This extra layer could be removed or does it exist to increase specificity? (if so maybe a comment?)

apps/shared/account/account.component.html
```diff
@@ -0,0 +1,46 @@
...
+          <span class="provider-email">{{ account.googleInfo?.email || "Unknown" }}</span>
+        </div>
+        <div class="provider-row">
+          <img class="provider-icon" src="http://github.com/favicon.ico" />
+          <span class="provider-email">{{ account.githubInfo?.email || "Account Not Linked" }}</span>
+          <button *ngIf="account.githubInfo === null" mat-icon-button class="link-provider-account"
+            (click)="account.linkWithGithub()">
+            <mat-icon>add_link</mat-icon>
+          </button>
+          <button *ngIf="account.githubInfo" mat-icon-button class="link-provider-account"
```
devversion, line 26: Maybe add a button title/label for these icon-actions?

apps/prs/src/app/app.component.html
```diff
@@ -1,4 +1,7 @@
-<mat-toolbar color="primary" class="app-header-bar"></mat-toolbar>
+<mat-toolbar color="primary" class="app-header-bar">
+  <span class="spacer"></span>
+  <account></account>
```
devversion, line 3:
```suggestion
  <account-button></account-button>
```

apps/shared/account/account.component.html
```diff
@@ -0,0 +1,46 @@
+<img (click)="open()" referrerpolicy=“no-referrer” [src]="account.avatarUrl">
```
devversion, line 1: I guess we don't care a lot about high-a11y here, but should this be just a button instead?

apps/shared/account/account.component.html
```diff
@@ -0,0 +1,46 @@
...
+          <button *ngIf="account.githubInfo" mat-icon-button class="link-provider-account"
+            (click)="account.unlinkFromGithub()">
+            <mat-icon>link_off</mat-icon>
+          </button>
+        </div>
+      </div>
+    </ng-container>
+
+    <div class="projected-content">
+      <ng-content></ng-content>
```
devversion, line 35: Is there any information on what the projected content is used for? It's not quite clear to me (yet 😄)

apps/shared/account/account.service.ts
```diff
@@ -0,0 +1,81 @@
...
+  loggedInStateChange = new Subject<void>();
+
+  /** Whether a user is logged in currently. */
+  isLoggedIn!: boolean;
+
+  /** The Github account information for the current user, if a user is logged in and has linked their Github account. */
+  githubInfo!: UserInfo | null;
+
+  /** The Google account information for the current user, if a user is logged in. */
+  googleInfo!: UserInfo | null;
```
devversion, line 27: I think using nullable and then the exclamation is an anti-pattern, maybe be just assign null explicitly?

apps/shared/account/account.service.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  avatarUrl = DEFAULT_AVATAR_URL;
+  displayName: string | undefined;
+
+  constructor(private auth: Auth) {
+    this.auth.onAuthStateChanged((user) => {
+      this.avatarUrl = user?.photoURL || DEFAULT_AVATAR_URL;
```
devversion, line 34: For more explicitness we could use nullish coalescing, same below.

```suggestion
      this.avatarUrl = user?.photoURL ?? DEFAULT_AVATAR_URL;
```

devversion (approved):
LGTM, just some minor comments

apps/shared/account/BUILD.bazel
josephperrott: I think thats a good idea to have that distinction.

apps/shared/account/account.component.html
josephperrott: Yeah, it definitely should, good call.

apps/shared/account/account.service.ts
josephperrott: We actually don't want nullish coalescing in this case because if the photoURL we get is `''` we want to use the default one instead.

PR #510, opened 2022-04-11
feat(apps): create the base of the pull request table

josephperrott:
Create a table and base column definition for the pull request table to allow for the table columns to
be established incramentally.

### DIFF
--- a/apps/prs/src/app/pr-table/columns/base.ts
+++ b/apps/prs/src/app/pr-table/columns/base.ts
@@ -1,0 +1,14 @@
+import {Directive, ViewChild} from '@angular/core';
+import {MatCellDef, MatHeaderCellDef} from '@angular/material/table';
+
+// A Directive is used to satisfy the Angular compiler which wants a decorated class
+// even though its abstract.
+@Directive()
+export abstract class BaseColumn {
+  /** The cell definintion for a row. */
+  @ViewChild(MatCellDef, {static: true}) cell!: MatCellDef;
+  /** The header cell definintion for a row. */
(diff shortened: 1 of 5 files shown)


apps/prs/src/app/pr-table/columns/base.ts
```diff
@@ -0,0 +1,14 @@
...
+import {MatCellDef, MatHeaderCellDef} from '@angular/material/table';
+
+// A Directive is used to satisfy the Angular compiler which wants a decorated class
+// even though its abstract.
+@Directive()
+export abstract class BaseColumn {
+  /** The cell definintion for a row. */
+  @ViewChild(MatCellDef, {static: true}) cell!: MatCellDef;
+  /** The header cell definintion for a row. */
+  @ViewChild(MatHeaderCellDef, {static: true}) headerCell!: MatHeaderCellDef;
```
devversion, line 11: Are these intentionally static? https://angular.io/guide/static-query-migration#how-do-i-choose-which-static-flag-value-to-use-true-or-false

josephperrott: I think its safe to do so, but I suppose we don't specifically need to for this case.  Removed.

josephperrott: Actually, I just tested it and we have to use static because we want to get the template ref before its created to provide it to the table.

devversion: I'm a little confused on why the timing matters here since you only access the template refs in the after view init hook, but I don't know the specifics here and it shouldn't matter. 

I just raised it since it's not recommended to use `static` unless really needed.

PR #515, opened 2022-04-14
refactor(ng-dev): improve stamping for snapshot releases 

devversion:
See individual commits

### DIFF
--- a/ng-dev/release/stamping/env-stamp.ts
+++ b/ng-dev/release/stamping/env-stamp.ts
@@ -6,11 +6,12 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {join} from 'path';
-import {SemVer} from 'semver';
-import {GitClient} from '../../utils/git/git-client';
-import {createExperimentalSemver} from '../../release/versioning/experimental-versions';
 import * as fs from 'fs';
+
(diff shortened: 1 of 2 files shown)


ng-dev/release/stamping/env-stamp.ts
```diff
@@ -45,41 +46,37 @@ function hasLocalChanges(git: GitClient) {
...
- * In release mode, the version is based on the workspace version.
+ * In snapshot mode, we will include the current SHA along with the workspace version.
  */
 function getSCMVersions(
   git: GitClient,
   mode: EnvStampMode,
 ): {version: string; experimentalVersion: string} {
-  if (mode === 'release') {
-    const workspaceVersion = getVersionFromWorkspacePackageJson(git);
+  const workspaceVersion = getVersionFromWorkspacePackageJson(git).format();
```
josephperrott, line 64: nit: this could just be `version` and `experimentalVersion` since its always the workspace and is the only version we are working with.

josephperrott (approved):
LGTM

ng-dev/release/stamping/env-stamp.ts
devversion: 👍

PR #519, opened 2022-04-15
refactor(apps): use static helper methods for getting firestore references and converters

josephperrott:
Instead of creating a static converter with a mixin, the mixin instead sets methods to get the converter and DocumentReference using the class definition.

### DIFF
--- a/apps/shared/models/app-models.ts
+++ b/apps/shared/models/app-models.ts
@@ -1,9 +1,6 @@
 import {getDoc, getFirestore, doc, DocumentData} from 'firebase/firestore';
 import {BaseModel, Constructor, FirestoreReference, fromFirestoreReference} from './base';
 
-/** Marker symbol for noting that a model has been decorated for app use. */
-const decoratedForApp = Symbol('forApp');
-
 // Import all of the models for the module and decorate all of them for app use.
 import * as models from './index';
 Object.values(models).forEach(forApp);
@@ -18,7 +15,7 @@ function forApp<
   TBase extends Constructor<BaseModel<FirebaseModel>>,
 >(model: TBase) {
   const staticModel = model as unknown as typeof BaseModel;
-  staticModel.decoratedFor(decoratedForApp);
+  staticModel.markAsDecorated();
 
   /** The converter object for performing conversions in and out of Firestore. */
   const converter = {
--- a/apps/shared/models/app-models.ts
+++ b/apps/shared/models/app-models.ts
@@ -1,13 +1,13 @@
 import {getDoc, getFirestore, doc, DocumentData} from 'firebase/firestore';
 import {BaseModel, Constructor, FirestoreReference, fromFirestoreReference} from './base';
 
-// Import all of the models for the module and decorate all of them for App usage.
+/** Marker symbol for noting that a model has been decorated for app use. */
+const decoratedForApp = Symbol('forApp');
+
+// Import all of the models for the module and decorate all of them for app use.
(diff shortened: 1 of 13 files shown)


apps/shared/models/app-models.ts
```diff
@@ -1,51 +1,51 @@
...
+   * Class method to get the converter object, ensuring that the converter returned is always
+   * the converter from the specific class definition rather than a parent class.
+   */
+  model.prototype.getConverter = function () {
+    return converter;
+  };
+
+  /**
+   * Gets the model referenced by the provided FirestoreReference, returning the same reference
+   * as previously queried if the instance cache finds an entry.
```
devversion, line 40: Do we need the caching here? I wonder if it would be more safe to just not do that, and pass-through models where needed? Is there a concrete case you are thinking of (and should we put it in the comment if so?)

apps/shared/models/base.ts
```diff
@@ -1,22 +1,77 @@
...
+
+/**
+ * Because it is less expensive to store strings than Firestore references, with no change in the
+ * value it provides, we have this local type to express this usage.
+ */
+export type FirestoreReference<T> = {
+  // ensures this "Reference" string is not passed around as "string" accidentally.
+  __eefirestoreReference: true;
+  __eereferencedType: T;
+};
```
devversion, line 67: 🎉

apps/shared/models/app-models.ts
devversion: Especially asking because `forServer` does not have this. I think it might be easier without the caching, but you might be thinking of something.

devversion (approved):
Nice! Can you just confirm that:

The use of both models (from server and from app) in a "sub-project" is not working because the monkey-patching would override each other. Maybe add some safety mechanism where we throw if the properties are already defined? I think this would be a great heads-up if something like this could cause unexpected/confusing behavior.

apps/shared/models/app-models.ts
josephperrott: I went ahead and removed it, I think its something that might make sense to do in the future, but we can look at it when the time comes.

josephperrott:
@devversion Previously it failed because the types for both essentially can't exist together because the namespaces clobber each other.

I went ahead and added in a runtime check as well, which not only confirms that we only tried to decorate a model once, but also that a model is guaranteed to have been decorated.

apps/shared/models/base.ts
```diff
@@ -31,6 +36,21 @@ export abstract class BaseModel<T> {
   } {
     throw Error('This was not implemented');
   }
+
+  /** Set the decoratedFor value on the prototype of the class. */
+  static decoratedFor(decoratedFor: symbol) {
+    this.prototype.decoratedForFunc(this, decoratedFor);
```
devversion, line 42: Would it be better to just assign this as a static property, and not on a per-instance basis through the prototype? That would simplify the code. I can't think of a benefit with using prototype

devversion (approved):

apps/shared/models/base.ts
josephperrott: We want to have it as a class property so that we can check it during construction to make sure that some decorator has been set.

However, now that I looked closer at it, we don't need to leverage a prototype method to set the value on the prototype, we can just do it directly from the static method.

devversion: I see what you are saying. It's because we can't know the class constructor from the derived classes in the Model base class

PR #520, opened 2022-04-16
Create summary column

josephperrott:
See invididual commits.


apps/shared/models/pull-request.ts
```diff
@@ -33,25 +33,48 @@ export interface FirestorePullRequest {
...
+    this.createdAt = data.createdAt;
+    this.draft = data.draft;
+    this.maintainerCanModify = data.maintainerCanModify;
+    this.number = data.number;
+    this.requestedReviewers = data.requestedReviewers as any;
+    this.title = data.title;
+    this.milestone = data.milestone as any;
+    this.assignees = data.assignees as any;
+    this.commit = data.commit;
+    User.getByReference(data.user).then((u) => (this.user = u));
```
devversion, line 75: Do we need to await this? Otherwise user might be undefined shortly right?

devversion: Tsec complains about this. I believe it would be good supporting async population of a model. We control the instantiation anyway, so that would improve this even further.

devversion (approved):

apps/shared/models/pull-request.ts
josephperrott: I awaited it and will make the instantiation async in a follow up.

devversion (approved):

devversion:
This PR was merged into the repository by commit 8a7e038d773fbae483d5366d7f27bd403f072665.

PR #521, opened 2022-04-18
Create a status column

josephperrott:
See individual commits.


apps/prs/src/app/pr-table/columns/status/status.component.ts
```diff
@@ -0,0 +1,26 @@
...
+@Component({
+  selector: 'status-column',
+  templateUrl: './status.component.html',
+  styleUrls: ['./status.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class StatusColumn extends BaseColumn {
+  name = 'status';
+
+  getStatusIconForStatus(status: string) {
```
devversion, line 13: Could we use an octokit type here for the status values? 

Also nit: I think we should not throw errors without message. Obviously the stack trace could always help, or source-maps, but a little message is low-effort to add.

devversion (approved):

PR #522, opened 2022-04-18
Add additional webhooks for Github events

josephperrott:
See individual commits.

### DIFF
--- a/apps/functions/githubWebhook/status.ts
+++ b/apps/functions/githubWebhook/status.ts
@@ -1,0 +1,20 @@
+import {StatusEvent} from '@octokit/webhooks-types';
+import {firestore} from 'firebase-admin';
+// TODO(josephperrott): Remove usage of FirestoreReference and fromFirestoreReference.
+import {
+  Status,
+  FirestoreReference,
+  fromFirestoreReference,
+} from '../../shared/models/server-models';
+
+export async function handleStatusEvent(event: StatusEvent) {
+  const {getFirestoreRefForGithubModel, fromGithub} = Status.getGithubHelpers();
+  /** The pull request model for the pull request status data. */
+  const status = fromGithub(event);
+  /** FirestoreReference for the pull request status. */
+  const firestoreRef = getFirestoreRefForGithubModel(event) as FirestoreReference<Status>;
+  /** The Firestore document reference for the pull request status. */
+  const docRef = firestore().doc(fromFirestoreReference(firestoreRef));
+
+  await docRef.set(Status.getConverter().toFirestore(status));
+}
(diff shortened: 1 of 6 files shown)


apps/functions/githubWebhook/status.ts
```diff
@@ -0,0 +1,20 @@
+import {StatusEvent} from '@octokit/webhooks-types';
+import {firestore} from 'firebase-admin';
+// TODO(josephperrott): Remove usage of FirestoreReference and fromFirestoreReference.
```
devversion, line 3: OOC: Why is this something you want to remove? Is that mainly a todo because you want to have better helpers?

apps/functions/githubWebhook/status.ts
```diff
@@ -0,0 +1,20 @@
...
+export async function handleStatusEvent(event: StatusEvent) {
+  const {getFirestoreRefForGithubModel, fromGithub} = Status.getGithubHelpers();
+  /** The pull request model for the pull request status data. */
+  const status = fromGithub(event);
+  /** FirestoreReference for the pull request status. */
+  const firestoreRef = getFirestoreRefForGithubModel(event) as FirestoreReference<Status>;
+  /** The Firestore document reference for the pull request status. */
+  const docRef = firestore().doc(fromFirestoreReference(firestoreRef));
+
+  await docRef.set(status.data);
```
devversion, line 19: Should we be using `toFirestore` here?

apps/shared/models/status.ts
```diff
@@ -14,7 +14,9 @@ export class Status extends GithubBaseModel<FirestoreStatus> {
 
   static override githubHelpers: GithubHelperFunctions<Status, StatusEvent, FirestoreStatus> = {
     buildRefString(model: StatusEvent) {
-      return toFirestoreReference(`githubCommit/${model.sha}/status/${model.context}`);
+      return toFirestoreReference(
+        `githubCommit/${model.sha}/status/${model.context.replace(/[\:\s\/]/g, '_')}`,
```
devversion, line 18: Super nit: The escapes for colon and slash should not be needed.

devversion (approved):

apps/functions/githubWebhook/status.ts
josephperrott: Yeah, I think we should be able to do this without having to have these helpers used this way, but I couldn't for the life of me figure it out at the time.  I was planning to just come back to it.

apps/functions/githubWebhook/status.ts
josephperrott: Done

PR #523, opened 2022-04-18
Set up user account related things for user creation, login and authentication.

josephperrott:
See individual commits.

### DIFF
--- a/apps/account-functions/before-sign-in.ts
+++ b/apps/account-functions/before-sign-in.ts
@@ -1,0 +1,34 @@
+import {
+  Auth,
+  https,
+  UserRecord,
+  UserEventUpdateRequest,
+  AuthEventContext,
+} from 'gcip-cloud-functions';
+
+/** Validate accounts before sign in using google cloud before sigin in syncronous function. */
+
+export const beforeSignIn = new Auth()
+  .functions()
+  .beforeSignInHandler(
+    async (user: UserRecord, context: AuthEventContext): Promise<UserEventUpdateRequest> => {
+      /** The UserEventUpdate to save based on the signin results. */
+      const event: UserEventUpdateRequest = {};
+
+      // If a user is able to reach this without a login credential being present, throw an auth error.
+      // Note: This should not be possible, but it doesn't hurt to have this check in place.
+      if (context.credential === undefined) {
+        throw new https.HttpsError(
+          'unauthenticated',
+          `Cannot sign in as '${user.email}' without credential.`,
+        );
+      }
+
+      // When users sign in with github, we save the access token as a claim on the user object.
+      if (context.credential.providerId === 'github.com') {
+        event.customClaims = {...event.customClaims, githubToken: context.credential.accessToken};
+      }
+
+      return event;
+    },
+  );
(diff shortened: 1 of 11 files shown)


apps/shared/account/account.service.ts
```diff
@@ -27,9 +27,19 @@ export class AccountService {
...
+  /** The current accounts github token, if available. */
+  private githubToken: string | null = null;
 
   constructor(private auth: Auth) {
-    this.auth.onAuthStateChanged((user) => {
+    this.auth.onIdTokenChanged((user) => {
+      if (user === null) {
+        this.githubToken = null;
+      } else {
+        user.getIdTokenResult(false).then((value) => {
```
devversion, line 38: The force refresh parameter is optional right? Maybe omit it, or add a parameter comment for clarity?

apps/firebase.json
```diff
@@ -5,7 +5,7 @@
   },
   "functions": {
     "predeploy": "bazel build //apps/functions:functions_compiled",
-    "source": "../dist/bin/apps/functions"
+    "source": "../apps/functions/dist/functions"
```
devversion, line 8:
```suggestion
    "source": "./functions/dist/functions"
```

devversion: is a symlink also working on windows? likely it does at this point.

apps/account-functions/before-sign-in.ts
```diff
@@ -0,0 +1,34 @@
...
+      if (context.credential === undefined) {
+        throw new https.HttpsError(
+          'unauthenticated',
+          `Cannot sign in as '${user.email}' without credential.`,
+        );
+      }
+
+      // When users sign in with github, we save the access token as a claim on the user object.
+      if (context.credential.providerId === 'github.com') {
+        event.customClaims = {...event.customClaims, githubToken: context.credential.accessToken};
```
devversion, line 29: Reading through the custom claims documentation, it seems like this is an anti-pattern, to provide custom data like that. I also wanted to ask if we would rather want to perform Github actions through the backend instead?

Note: I'm good proceeding here, but wanted to mention it. Maybe worth considering because this seems like an architectural question

devversion (approved):

apps/firebase.json
josephperrott: I believe it should and will verify.

apps/account-functions/before-sign-in.ts
josephperrott: I think that its not an anti-pattern because this data is specific to the user and should not be accessed any other way, its similar to the the token for the firebase account being included in the claims as well.  But if you would prefer we store it another way I definitely can.

Because we want to have the github credential be for the individual user who does the action (for instance setting the target shows in Github as the user setting a label) I don't think it makes sense to have it proxied through our backend, but instead just done directly via the Github API.

devversion: Yeah, not having any objections here. I just wanted to bring it up since the quote from the docs sounded to me like the incorporation of an access token as custom claim is not recommended.

> Custom claims are only used to provide access control. They are not designed to store additional data (such as profile and other custom data

I interpreted "access control" to be more like assigning certain properties like `isAdmin`, and then in combination with security rules. Anyway though, I might be misinterpreting, and it shouldn't matter, so feel free to go ahead.

devversion (approved):

PR #534, opened 2022-04-22
fix(apps): avoid symlink that breaks github action cloning

devversion:
The symlink is invalid when a Github action is cloned because
Bazel did not run yet. We should avoid the symlink to keep our
actions working:

https://github.com/angular/components/runs/6127973017?check_suite_focus=true


josephperrott (approved):
LGTM

PR #538, opened 2022-04-26
fix(ng-dev): make `new-main-branch` command more flexible to local setup

devversion:
After we will perform the default branch rename, there are three scenarios for
a contributor:

1. They manually fetch `upstream/main`, run `yarn` and then the
   migration tool.
2. They will not sync with upstream (as `upstream/master` does not
   exist), run `yarn` and then the migration tool.
3. They will manually update everything locally and in their forks. This
bullet-point is just for completion (not something we bother about here).

For (1): The `master` branch and `main` branch will locally exist. We should
ask to delete the `main` branch switch back to `master`. Then continue with
the full migration (with all other changes to the local git config)

For (2): We should run the migration as is (e.g. fetch main and set it up locally).

### DIFF
--- a/ng-dev/misc/new-main-branch/cli.ts
+++ b/ng-dev/misc/new-main-branch/cli.ts
@@ -39,12 +39,12 @@ async function handler() {
   }
 
   if (hasLocalBranch(git, 'main')) {
-    console.warn(yellow('The new `main` branch is already fetched locally. In order to run'));
-    console.warn(yellow('this tool, the `main` branch needs to be non-existent locally.'));
-    console.warn('');
-    console.warn(yellow('The tool will re-fetch the `main` branch and configure it properly.'));
+    warn(yellow('The new `main` branch is already fetched locally. In order to run'));
+    warn(yellow('this tool, the `main` branch needs to be non-existent locally.'));
+    warn('');
+    warn(yellow('The tool will re-fetch the `main` branch and configure it properly.'));
 
-    if (!(await promptConfirm('Do you want to proceed and delete the `main` branch?'))) {
+    if (!(await promptConfirm('Do you want to proceed and delete the local `main` branch?'))) {
       error(red('Aborting..'));
       return;
     }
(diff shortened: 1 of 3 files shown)


ng-dev/misc/new-main-branch/cli.ts
```diff
@@ -33,19 +33,29 @@ async function handler() {
...
-
   if (!hasLocalBranch(git, 'master')) {
     error(red('Local repository does not have a local branch named `master`. Aborting..'));
     return;
   }
 
   if (hasLocalBranch(git, 'main')) {
-    error(red('Local repository already has a branch named `main`. Aborting..'));
-    return;
+    console.warn(yellow('The new `main` branch is already fetched locally. In order to run'));
```
clydin, line 42: nit: Looks like there is a `warn` function imported should that be used instead of `console.warn`?

ng-dev/misc/new-main-branch/cli.ts
```diff
@@ -33,19 +33,29 @@ async function handler() {
...
+    console.warn(yellow('this tool, the `main` branch needs to be non-existent locally.'));
+    console.warn('');
+    console.warn(yellow('The tool will re-fetch the `main` branch and configure it properly.'));
+
+    if (!(await promptConfirm('Do you want to proceed and delete the `main` branch?'))) {
```
clydin, line 47:
```suggestion
    if (!(await promptConfirm('Do you want to proceed and delete the local `main` branch?'))) {
```
might be good to explicitly mention that the _local_ branch will be deleted.

ng-dev/misc/new-main-branch/cli.ts
```diff
@@ -33,19 +33,29 @@ async function handler() {
...
+      error(red('Aborting..'));
+      return;
+    }
+
+    // If we are already on the `main` branch, we cannot delete it without
+    // checking out a different branch. We switch to `master` in such a case.
+    if (getCurrentBranch(git) === 'main') {
+      git.run(['checkout', 'master']);
```
clydin, line 55: Will there always still be a `master` branch at this time?

ng-dev/misc/new-main-branch/cli.ts
devversion: Good catch. done!

ng-dev/misc/new-main-branch/cli.ts
devversion: Yes, we checked above if the local old `master` branch still exists. We assume this to be still existent locally, and assert it early in the CLI command

clydin (approved):

PR #547, opened 2022-04-30
fix(ng-dev): release notes not capturing multiple note keywords from single commit

devversion:
Commits like the following were not captured properly because we only
assumed a single deprecation/breaking-change to be set on a commit.


josephperrott (approved):
LGTM

PR #571, opened 2022-05-21
feat(circleci-orb): initial setup for publishing dev-infra orb

devversion:
Initial setup for publishing an dev-infra orb.


circleci-orb/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
+ORB_NAME = "devversion/ng-dev-infra-private"
```
josephperrott, line 1:
```suggestion
ORB_NAME = "angular/dev-infra"
```

I think we can actually host this as a private orb which means that we can just name it dev-infra?

Or maybe it makes sense to still use `dev-infra-private`

josephperrott (approved):
LGTM

Nice, this will be a great base for us to create common orb actions with.

circleci-orb/BUILD.bazel
devversion: Yeah, I think this makes sense. I wasn't sure if I had access, but now I surely have, and it seems to work!

PR #574, opened 2022-05-23
build: update renovate to exclude `protobufjs` CLI deps from being updated

devversion:
We have some trickery in the Protobuf Bazel macro where we need to pull
some of the `protobufjs/cli` dependencies are direct dependencies. These
dependencies need to satsify the `devDependencies` ranges of
`protobufjs`. Renovate does not know about this and attempts to update
versions we should not update/cannot update currently.


josephperrott (approved):
LGTM


One nit: Can you add why we are including `espree` in the commit message?

devversion:
Should be more clear now

PR #576, opened 2022-05-23
refactor(ng-dev/utils): update spinner to have success and failure messages, better interop with logging

josephperrott:
Update the Spinner class to have class methods for success and failure message, automatically placing a
checkmark or x as appropriate for the method.

Additionally, updates the shared logging code to properly print a message while a Spinner is active, using
method on the spinner directly to print correctly instep with the spinner.

### DIFF
--- a/ng-dev/utils/console.ts
+++ b/ng-dev/utils/console.ts
@@ -6,11 +6,13 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {inspect} from 'util';
 import * as chalk from 'chalk';
 import {writeFileSync} from 'fs';
 import {prompt} from 'inquirer';
 import {join} from 'path';
 import {Arguments} from 'yargs';
+import {Spinner} from './spinner';
 
 import {GitClient} from './git/git-client';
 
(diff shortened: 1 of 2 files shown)


ng-dev/utils/console.ts
```diff
@@ -103,7 +105,11 @@ function buildLogLevelFunction(loadCommand: () => Function, level: LOG_LEVELS) {
  */
 function runConsoleCommand(loadCommand: () => Function, logLevel: LOG_LEVELS, ...text: unknown[]) {
   if (getLogLevel() >= logLevel) {
-    loadCommand()(...text);
+    if (Spinner.activeSpinner) {
+      Spinner.activeSpinner.info(text.map((val) => inspect(val, true, 3, true)).join(' '));
+    } else {
+      loadCommand()(...text);
```
devversion, line 111: I find it a little magical that any of the logging methods would interact with the spinner tbh. Can't we just print as always and reset the spinner to the lines after? e.g.

`/ Loading..` then we push a new message. We delete the current spinner line, call the original log method like before, and then re-start the spinner in a new line. Basically almost the same, but not doing anything with `inspect`, and without passing through the message to the spinner.

PR #578, opened 2022-05-24
feat(ng-dev/pr): allow for manual override of targeted branches

josephperrott:
Using the `manual-branch-targeting` user's can choose to manually select the branches they would like to target rather than relying on the branches defined by the label.  This allows for one-off manual overrides of the logic we rely on. As noted in the warning at runtime, this does remove the attempts we make to guarantee that all commits are valid for the targeted branches.

### DIFF
--- a/ng-dev/pr/common/validation/failures.ts
+++ b/ng-dev/pr/common/validation/failures.ts
@@ -130,4 +130,11 @@ export class PullRequestFailure {
       'breaking change notes (i.e. commits do not have a `BREAKING CHANGE: <..>` section).';
     return new this(message);
   }
+
+  static failedToManualSelectGithubTargetBranch(branch: string) {
+    const message =
+      `Pull Requests must merge into their targeted Github branch. If this branch (${branch}) ` +
+      'should not be included, please change the targeted branch via the Github UI.';
+    return new this(message);
+  }
 }
--- a/ng-dev/pr/merge/task.ts
+++ b/ng-dev/pr/merge/task.ts
@@ -6,9 +6,12 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {promptConfirm} from '../../utils/console';
+import {promptConfirm, red, yellow} from '../../utils/console';
 import {AuthenticatedGitClient} from '../../utils/git/authenticated-git-client';
 import {GitCommandError} from '../../utils/git/git-client';
+import * as semver from 'semver';
+import {prompt} from 'inquirer';
+import {warn} from '../../utils/console';
 
 import {PullRequestConfig} from '../config';
 import {PullRequestFailure} from '../common/validation/failures';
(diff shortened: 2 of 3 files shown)


ng-dev/pr/common/validation/failures.ts
```diff
@@ -130,4 +130,11 @@ export class PullRequestFailure {
       'breaking change notes (i.e. commits do not have a `BREAKING CHANGE: <..>` section).';
     return new this(message);
   }
+
+  static invalidBranchTargets(branch: string) {
```
devversion, line 134: We seem to have a similar error: `mismatchingTargetBranch`

ng-dev/pr/merge/task.ts
```diff
@@ -114,7 +124,16 @@ export class PullRequestMergeTask {
...
 
+    if (this.flags.forceManualBranches) {
+      const forceManualBranchesFailure = await this.setTargetedBranchesManually(pullRequest);
+      if (forceManualBranchesFailure) {
+        return forceManualBranchesFailure;
+      }
+    }
+
     if (
+      // In cases where manual branch targeting is used, the user is already beforming this confirmation.
```
devversion, line 135:
```suggestion
      // In cases where manual branch targeting is used, the user already confirmed.

File: ng-dev/pr/merge/task.ts [hunk 3]
```diff
@@ -167,4 +186,94 @@ export class PullRequestMergeTask {
...
+    } catch {
+      warn(
+        'Unable to determine the active LTS branches as a release config is not set for this repo.',
+      );
+    }
+
+    // Gather the current active release trains.
+    const {latest, next, releaseCandidate} = await fetchActiveReleaseTrains({
+      name,
+      nextBranchName: mainBranchName,
```
devversion, line 215: Even though it's the same, for correctness we should use getNextBranchName

ng-dev/pr/merge/task.ts
```diff
@@ -167,4 +186,94 @@ export class PullRequestMergeTask {
...
+    const {selectedBranches, confirm} = await prompt([
+      {
+        type: 'checkbox',
+        default: pullRequest.targetBranches,
+        choices: activeBranches.map(({branchName, version}) => {
+          return {
+            checked: pullRequest.targetBranches.includes(branchName),
+            value: branchName,
+            short: branchName,
+            name: `${branchName} (${version}) ${
```
devversion, line 241: real nit: the space should only be added when needed below

```suggestion
            name: `${branchName} (${version})${
```

ng-dev/pr/merge/task.ts
```diff
@@ -167,4 +186,94 @@ export class PullRequestMergeTask {
...
+      {
+        type: 'confirm',
+        default: false,
+        message:
+          red('!!!!!! WARNING !!!!!!!\n') +
+          yellow(
+            'Using manual branch selection disables protective checks provided by the merge ' +
+              'tooling. This means that the merge tooling will not prevent changes which are not ' +
+              'allowed for the targeted branches. Please proceed with caution.\n',
+          ) +
```
devversion, line 258: 👍

ng-dev/pr/merge/task.ts
```diff
@@ -167,4 +186,94 @@ export class PullRequestMergeTask {
...
+          'Are you sure you would like to proceed with the selected branches?',
+        name: 'confirm',
+      },
+    ]);
+
+    if (confirm === false) {
+      return {status: MergeStatus.USER_ABORTED};
+    }
+
+    // The Github Targeted branch must always be selected.  It is not currently possible to make a
```
devversion, line 268: Super nit: extra space

devversion (commented):
Overall lgtm, nice

PR #583, opened 2022-05-26
fix(ng-dev/pr): properly name the rebase command

josephperrott:
Properly sets the rebase subcommand so it is accessible in the tooling.


devversion (approved):

devversion:
@josephperrott I think discover conflicts also needs the same fixup. I didn't spot this in review

PR #585, opened 2022-05-28
feat(ng-dev): move release output build and checks to staging phase

devversion:
Instead of performing the build only after the staging PR has been
merged, we run the build and its output checks before staging so
that issues can be spotted early in the release process (avoiding
complicated/confusing situations where the bump commit already landed).

Also this will allow us to combine the current pre-staging and
post-build checks into a single step, allowing this to become a
configurable config option for custom pre-release pre-checks.


ng-dev/release/publish/actions.ts
```diff
@@ -563,50 +638,47 @@ export abstract class ReleaseAction {
...
-
-    // Combine the built packages with their NPM package information. This is helpful
-    // later for filtering out e.g. experimental packages.
-    const builtPackagesWithInfo = mergeBuiltPackagesWithInfo(
-      builtPackages,
-      releaseInfo.npmPackages,
-    );
+    // Ensure no commits have landed since we started the staging process. This would signify
+    // that the locally-built release packages are not reflecting what is on GitHub.
+    if (!(await this._isRevisionAheadOfBase(beforeStagingSha, versionBumpCommitSha, 1))) {
```
josephperrott, line 673: Can you comment why we only expect to only be 1 commit ahead? I want to make sure if we ever add more commits to the release process that its easy to understand how this plays into it (as well as to find it)

ng-dev/release/publish/actions.ts
```diff
@@ -563,50 +638,47 @@ export abstract class ReleaseAction {
...
+      error(red(`  ✘   Unexpected additional commits have landed while staging the release.`));
+      error(red('      Please revert the bump commit and retry, or cut a new version on top.'));
+      throw new FatalReleaseActionError();
```
josephperrott, line 676: Nit: We could list the unexpected commit shas if you want to

ng-dev/release/publish/actions/branch-off-next-branch.ts
```diff
@@ -44,30 +44,36 @@ export abstract class BranchOffNextBranchBaseAction extends ReleaseAction {
...
 
   override async perform() {
+    const nextBranchName = this.active.next.branchName;
     const compareVersionForReleaseNotes = await getReleaseNotesCompareVersionForNext(
       this.active,
       this.config,
     );
     const newVersion = await this._computeNewVersion();
     const newBranch = `${newVersion.major}.${newVersion.minor}.x`;
+    const beforeStagingSha = await this.getLatestCommitOfBranch(nextBranchName);
```
josephperrott, line 54: Shouldn't this value be created during the build method as its meant to validate what was used to actually build?

Particularly in this case as a whole new branch is being created.

ng-dev/release/publish/actions.ts
devversion: I agree, but it's probably not worth the restructuring/complications as this should happen rarely.

ng-dev/release/publish/actions/branch-off-next-branch.ts
devversion: I thought a lot about this, but we would basically be making multiple requests for the same information (getting the latest SHA of the upstream branch). Also considered just using the local `rev-parse HEAD` once we checked out the remote branch (to save the second API call -- but this is inconsistent and has other downsides -- happy to expand).

All actions require awareness of how the staging / publishing works anyway, and this branch-off action heavily relies on that information already, so to save some additional requests, and to avoid having to mock a duplicate second GitHub request, I decided to just call it `beforeStagingSha` and require it to be passed manually when staging is performed in an advanced way (like done here in the banch-off scenario)

josephperrott: Okay, sounds good

josephperrott (approved):
LGTM

PR #586, opened 2022-05-30
refactor(ng-dev): include branch name for feature-freeze train for cut RC option

### LINKED ISSUE #556: RC option in the release selector to mention a branch
When running a release command via ng-dev (`yarn ng-dev release publish`), one of the options during the RC phase is: "Cut a first release-candidate for the feature-freeze branch (v14.0.0-rc.0).". Unlike other options, the text doesn't contain a name of a particular branch that will be used.

It'd be great to include the branch info into the option text as well, so there is no ambiguity, for example: "Cut a first release-candidate (v14.0.0-rc.0) for the feature-freeze branch (14.0.x).".

devversion:
Fixes #556.

@AndrewKushnir I kept the version at the end of the message since I want the "to-be-cut" version always easily discoverable. Hope that makes sense.


AndrewKushnir (approved):
Thanks for the improvement @devversion 👍

josephperrott (approved):
LGTM

PR #594, opened 2022-06-03
refactor(ng-dev): do not print `undefined` when release prechecks fail

devversion:
We always print `undefined` upon release pre-check errors. The actual
pre-check error messaging is piped from the child process to
stdout/stderr, and therefore the the `spawn` utility-thrown error
is not necessary to be printed.


alan-agius4 (approved):
LGTM

PR #595, opened 2022-06-03
refactor(ng-dev): enable color for non-interactive child processes

devversion:
Since non-interactive child processes without `stdio: inherit` do not
have a TTY, chalk in the child processes is not rendering colors.

We expect to have color support in all environments and therefore
should force-enable colors, so that failures/messages from release
external commands (like `ng-dev release precheck`) are colored.

In both modes, either where we directly forward to `stdout`/`stderr`,
color support is expected to be available, and in `silent` mode, colors
are accepted as well because we either print the captured output again,
or store it in a file (which will then just have ANSI colors like the
ng-dev log currently has anyway).


crisbeto (approved):
LGTM

PR #610, opened 2022-06-10
fix(ng-dev/pr): yargs positional placeholder does not match positional name

devversion:
In the `CommandModule` below we named the positional `<pr>`, but our
option does not reflect this. This causes the PR number to be not passed
to the GraphQL GitHub request. This happens due to a recent refactoring
that intended to standardize `pr` and `pr-number` to just `pr`.


crisbeto (approved):
LGTM

PR #611, opened 2022-06-10
ESM Migration #2: Support ESM-written ng-dev configurations

devversion:
This is already being tested/used in https://github.com/angular/components/pull/25052

### DIFF
--- a/ng-dev/commit-message/validate-range/validate-range.ts
+++ b/ng-dev/commit-message/validate-range/validate-range.ts
@@ -21,7 +21,7 @@ const isNonFixup = (commit: Commit) => !commit.isFixup;
 const extractCommitHeader = (commit: Commit) => commit.header;
 
 /** Validate all commits in a provided git commit range. */
-export async function validateCommitRange(from: string, to: string) {
+export async function validateCommitRange(from: string, to: string): Promise<void> {
   /** A list of tuples of the commit header string and a list of error messages for the commit. */
   const errors: [commitHeader: string, errors: string[]][] = [];
 
@@ -33,7 +33,10 @@ export async function validateCommitRange(from: string, to: string) {
    * Whether all commits in the range are valid, commits are allowed to be fixup commits for other
    * commits in the provided commit range.
    */
-  const allCommitsInRangeValid = commits.every((commit, i) => {
+  let allCommitsInRangeValid = true;
+
+  for (let i = 0; i < commits.length; i++) {
+    const commit = commits[i];
     const options: ValidateCommitMessageOptions = {
       disallowSquash: true,
       nonFixupCommitHeaders: isNonFixup(commit)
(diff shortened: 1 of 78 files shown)


ng-dev/commit-message/validate-range/validate-range.ts
```diff
@@ -43,12 +46,14 @@ export async function validateCommitRange(from: string, to: string) {
...
-    const {valid, errors: localErrors} = validateCommitMessage(commit, options);
+
+    const {valid, errors: localErrors} = await validateCommitMessage(commit, options);
     if (localErrors.length) {
       errors.push([commit.header, localErrors]);
     }
-    return valid;
-  });
+
+    allCommitsInRangeValid = allCommitsInRangeValid && valid;
```
josephperrott, line 55: We don't need to continue checking after we find an invalid commit.  Once we find an invalid commit, we can set it to false and `break`

ng-dev/utils/git/authenticated-git-client.ts
```diff
@@ -141,17 +141,27 @@ export class AuthenticatedGitClient extends GitClient {
     }));
   }
 
+  /** The previously configured access token. */
+  private static _token: string | null = null;
   /** The singleton instance of the `AuthenticatedGitClient`. */
-  private static _authenticatedInstance: AuthenticatedGitClient;
+  private static _authenticatedInstance: AuthenticatedGitClient | null = null;
```
josephperrott, line 147: This should become a `Promise<AuthenticatedGitClient>` so that we don't have to create a new promise everytime right?

ng-dev/utils/git/git-client.ts
```diff
@@ -224,9 +223,11 @@ export class GitClient {
    * Static method to get the singleton instance of the `GitClient`, creating it
    * if it has not yet been created.
    */
-  static get(): GitClient {
+  static async get(): Promise<GitClient> {
     if (!this._unauthenticatedInstance) {
-      GitClient._unauthenticatedInstance = new GitClient();
+      GitClient._unauthenticatedInstance = new GitClient(
```
josephperrott, line 228: Same comment as the authenticated client

josephperrott (approved):
LGTM overall

ng-dev/commit-message/validate-range/validate-range.ts
devversion: The way the code was written, it seemed like we actually would want to collect as much findings as possible, instead of printing them sequentially one by one after fixing

ng-dev/utils/git/authenticated-git-client.ts
devversion: I would need to think more about this. I considered it initially, but it required dealing with Promise chaining directly, which felt unexpected

devversion: Creating the short-lived promise is not necessarily a problem here

ng-dev/commit-message/validate-range/validate-range.ts
devversion: I will keep it like that because otherwise keeping track of the errors that way wouldn't make sense as well. Happy to clean up in a follow-up.

ng-dev/utils/git/authenticated-git-client.ts
devversion: Done!

ng-dev/commit-message/validate-range/validate-range.ts
josephperrott: Makes sense, its probably best that we collect all of the findings.  From the perspective of how it was previously written `Array.every` exits as soon as it encounters a `false`.  But as you said tracking all of the errors is probably what we would really like.

devversion: Yeah, agreed. I realized `every` would exit early, but then it felt like everything around it was built to track more errors. could have been a simple `.find` otherwise

PR #615, opened 2022-06-15
fix(ng-dev): yarnpkg lockfile package cannot be bundled with ESM

devversion:
ESBuild bundles CommonJS packages, but some cannot be due to use of
CJS-only features. The yarn lockfile is one package that needs to be
kept external for it to leverage CJS features through NodeJS built-in
ESM -> CJS interop.


josephperrott (approved):
LGTM

devversion:
Related issue: https://github.com/evanw/esbuild/issues/1874

PR #617, opened 2022-06-16
Various ESM migration fixes

devversion:


bazel/spec-bundling/index.bzl
```diff
@@ -61,23 +65,34 @@ def spec_bundle(
...
 
-    # Browser tests (Karma) need named AMD modules to load.
-    # TODO(devversion): consider updating `@bazel/concatjs` to support loading JS files directly.
-    esbuild_rule = esbuild_amd if is_browser_test else esbuild_esm_bundle
-    amd_name = "%s/%s/%s" % (workspace_name, package_name, name + "_spec") if is_browser_test else None
+    esbuild_rule = None
+
+    if is_browser_test:
+        esbuild_rule = esbuild_amd
+        amd_name = "%s/%s/%s" % (workspace_name, package_name, name + "_spec") if is_browser_test else None
```
crisbeto, line 72: Is the `is_browser_test` check here necessary? This whole block is inside of a `if is_browser_test:`.

crisbeto (approved):
LGTM with one nit.

bazel/spec-bundling/index.bzl
devversion: good catch

PR #619, opened 2022-06-16
build: improve stability of concurrent integration tests running locally

devversion:
We have some non-RBE tests (to test platform-specific behavior), and
even within the sandbox, the shared Yarn cache is accesed. This can
result in errors as the folder might be accessed concurrently
(even though it should not be writable with Bazels sandbox)

https://app.circleci.com/pipelines/github/angular/dev-infra/306734/workflows/8ee6418f-cac4-489c-9d21-4baf6bf6b35b/jobs/309267/steps


crisbeto (approved):
LGTM

PR #629, opened 2022-06-20
build: fix remote execution not enabled on linux builds

devversion:
Looks like the path is incorrect and we didn't run with RBE.


bazel/api-golden/test/BUILD.bazel
```diff
@@ -1,4 +1,17 @@
...
+
+copy_to_bin(
+    name = "test_package_golden_dir",
+    srcs = glob(["goldens/test_package/**"]),
+)
+
+copy_to_bin(
+    name = "pkg_no_exports_field_golden_dir",
+    srcs = glob(["goldens/pkg_no_exports_field/**"]),
+)
```
devversion, line 14: This is all necessary now because RBE does not support tree artifact source directories

PR #632, opened 2022-06-21
refactor(ng-dev/release): indicate that a major stable release is first cut as `@next`

devversion:
When we previously cut v14, it was not quite obvious to the caretaker
that the cut-stable action will publish to `@next` first (as it was
requested/required by the release checklist).

cc. @alxhub

### DIFF
--- a/ng-dev/release/publish/actions/cut-stable.ts
+++ b/ng-dev/release/publish/actions/cut-stable.ts
@@ -19,16 +19,19 @@ import {ExternalCommands} from '../external-commands.js';
  */
 export class CutStableAction extends ReleaseAction {
   private _newVersion = this._computeNewVersion();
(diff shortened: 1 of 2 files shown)


ng-dev/release/publish/actions/cut-stable.ts
```diff
@@ -19,16 +19,19 @@ import {ExternalCommands} from '../external-commands.js';
  */
 export class CutStableAction extends ReleaseAction {
   private _newVersion = this._computeNewVersion();
+  private _isNewMajor = this.active.releaseCandidate!.isMajor;
```
josephperrott, line 22:
```suggestion
  private _isNewMajor = this.active.releaseCandidate?.isMajor;
```


We can't assert that releaseCandidate will be defined right?

josephperrott (approved):
LGTM

just one quick question

PR #642, opened 2022-06-27
fix(ng-dev): do not match dots automatically in pullapprove

devversion:
Pullapprove does not match dot files automatically, using `wcmatch`. We
should adjust our verify logic to reflect that.

This will cause issues in framework, but expected..


josephperrott (approved):
LGTM

PR #644, opened 2022-06-27
feat(ng-dev): validate all licenses as being allowed for our Angular projects

josephperrott:
A tool for validating that all packages provide allowed licenses in our projects.


ng-dev/misc/cli.ts
```diff
@@ -10,6 +10,7 @@ import {Argv} from 'yargs';
 import {BuildAndLinkCommandModule} from './build-and-link/cli.js';
 import {NewMainBranchCommandModule} from './new-main-branch/cli.js';
 import {UpdateYarnCommandModule} from './update-yarn/cli.js';
+import { ValidateLicensesModule } from './validate-licenses/cli.js';
```
devversion, line 13: Nit: Inconsistent spacing here (likely lint will fail)

ng-dev/misc/BUILD.bazel
```diff
@@ -11,5 +11,9 @@ ts_library(
         "@npm//@types/node",
         "@npm//@types/yargs",
         "@npm//chalk",
+        "@npm//license-checker",
+        "@npm//@types/license-checker",
+        "@npm//spdx-satisfies",
+        "@npm//@types/spdx-satisfies",
```
devversion, line 17: Can we have a fine-grained target for the new sub-command? We should start doing that incrementally now.

ng-dev/misc/validate-licenses/cli.ts
```diff
@@ -0,0 +1,51 @@
...
+export interface Options {}
+
+/** Yargs command builder for the command. */
+function builder(argv: Argv): Argv<Options> {
+  return argv;
+}
+
+/** Yargs command handler for the command. */
+async function handler({ }: Arguments<Options>) {
+  await checkAllLicenses(determineRepoBaseDirFromCwd()).then(({ valid, maxPkgNameLength, packages }) => {
```
devversion, line 25: Why not use `async/await` instead?

ng-dev/misc/validate-licenses/cli.ts
```diff
@@ -0,0 +1,51 @@
...
+      process.exitCode = 1;
+    });
+}
+
+/** CLI command module. */
+export const ValidateLicensesModule: CommandModule<{}, Options> = {
+  builder,
+  handler,
+  command: 'validate-licenses',
+  describe: 'Validate the licenses for all dependencies for the repo',
```
devversion, line 50:
```suggestion
  describe: 'Validate the licenses for all dependencies in the project',
```

ng-dev/misc/validate-licenses/validate.ts
```diff
@@ -0,0 +1,115 @@
...
+  allowed: boolean;
+}
+
+export interface LicenseCheckResult { valid: boolean, packages: ExpandedModuleInfo[], maxPkgNameLength: number };
+
+
+export async function checkAllLicenses(start: string): Promise<LicenseCheckResult> {
+  return new Promise((resolve, reject) => {
+    let maxPkgNameLength = 0;
+    init({ start }, (err: Error, pkgInfoObject: ModuleInfos) => {
```
devversion, line 70: It seems like this tool relies on the local `node_modules` to be installed, compared to e.g. analyzing the lock file and consulting the registry (which I would find nicer -- but also this runs rarely I think).

What would happen if the local node modules are not installed? should we have a good check for that?

ng-dev/misc/validate-licenses/validate.ts
```diff
@@ -0,0 +1,115 @@
...
+         * 
+         * Note: Typically a package will only have one license, but support for multiple license
+         *       is necessary for full support.
+         */
+        const licenses = Array.isArray(pkg.licenses) ? pkg.licenses : [pkg.licenses!];
+
+        return {
+          ...pkg,
+          name,
+          allowed: licenses.reduce((invalid, license) => invalid || assertAllowedLicense(license), false),
```
devversion, line 90: Is that a simple `licenses.some`?

ng-dev/misc/validate-licenses/validate.ts
```diff
@@ -0,0 +1,115 @@
...
+        valid: packages.every(pkg => pkg.allowed),
+        packages,
+        maxPkgNameLength,
+      });
+    });
+  });
+}
+
+
+const allowedLicensesString = allowedLicenses.join(' OR ');
```
devversion, line 104:
```suggestion
const allowedLicensesSpdxExpression = allowedLicenses.join(' OR ');
```

ng-dev/misc/validate-licenses/validate.ts
josephperrott: Realistically if node_modules is not installed neither is `ng-dev`, but I am unsure if we should just rely on that fact.

ng-dev/BUILD.bazel
```diff
@@ -9,6 +9,11 @@ NG_DEV_EXTERNALS = [
     # Package uses `__filename` and `__dirname` and cannot be bundled in ESM. We do not
     # intend to provide interop globals for this as it could hide other significant issues.
     "@yarnpkg/lockfile",
+    # SPDX uses dynamic imports that work at runtime, but esbuild complains about.
+    "spdx-ranges",
+    "spdx-license-ids",
+    "spdx-license-ids/deprecated",
+    "spdx-exceptions",
```
devversion, line 16: You should be able to just add `spdx-satisfies` instead of manual transitive deps. Do you also recall how ESBuild complained OOC.

package.json
```diff
@@ -28,20 +28,27 @@
...
     "@types/selenium-webdriver": "^4.0.18",
     "@types/send": "^0.17.1",
+    "@types/spdx-satisfies": "^0.1.0",
     "@types/tmp": "^0.2.1",
     "@types/uuid": "^8.3.1",
     "@types/ws": "8.5.3",
     "@types/yargs": "^17.0.0",
     "browser-sync": "^2.27.7",
     "clang-format": "1.8.0",
+    "license-checker": "^25.0.1",
```
devversion, line 42: All of the deps above are dev only dependencies and will be bundled by `ng-dev`.

ng-dev/BUILD.bazel
josephperrott: I tried, but it continues to complain about them.

The error is 
```
✘ [ERROR] Could not resolve "spdx-license-ids"

    node_modules/spdx-correct/index.js:17:29:
      17 │ var spdxLicenseIds = require('spdx-license-ids')
         ╵                              ~~~~~~~~~~~~~~~~~~
  You can mark the path "spdx-license-ids" as external to exclude it from the bundle, which will remove this error. You can also surround this "require" call with a try/catch block to handle this failure at run-time instead of bundle-time.
```

Same thing for all of the other ones.

devversion: Seems to be because e.g. https://unpkg.com/browse/spdx-license-ids@3.0.11/ is just exposing a JSON file. We would need to tell ESBuild to also load JSON files (which is an option IIRC).

josephperrott: From looking at the docs, it looks like the json file loader is included by default.

devversion: Yeah, but we aren't setting it up. https://github.com/angular/dev-infra/blob/main/bazel/esbuild/index.bzl#L31

I think we can add `.json` here (and in other places)

josephperrott: Ahh that did the trick.

bazel/esbuild/index.bzl
```diff
@@ -28,7 +28,7 @@ def esbuild_esm_bundle(name, **kwargs):
     """
 
     args = dict(
-        resolveExtensions = [".mjs", ".js"],
+        resolveExtensions = [".json", ".mjs", ".js"],
```
devversion, line 31: It should be the last one. We want to prioritize JS in this rule.

devversion (approved):

bazel/esbuild/index.bzl
devversion: Order matters here

devversion (approved):

PR #647, opened 2022-06-28
fix(github-actions): rework token revoke mechanism to not rely on `post` run

devversion:
Currently Github actions acquire tokens from the secrets for an app
installation. The actions can operate on repositories using this token.

Later, as a Github action (separate step) we attempt to revoke the
token. This does not work at all, and never did, because the previously
used installation Github token is not known in the post step, so the
post step always failed.

We rework this to always revoke the token as part of the Node process
where we acquired the installation token.


alan-agius4 (approved):
LGTM

PR #652, opened 2022-06-28
feat(ng-dev/pr): provide instructions for pushing back to the remote branch for pr checkout

josephperrott:
Provide, as a convenience, the command for pushing the locally checked out branch back to the PR on github.


devversion (approved):
Nice! this will close #384

PR #663, opened 2022-06-29
build: skip building of bottleneck `re2` native dependency

devversion:
The `re2` package is an optional dependency of `firebase-tools`
(transitively). It's super slow to build locally and also doesn't
seem to be cached locally, nor on CI. We should disable it as it
slows down the install by 1min.

Can be removed in the future with (worth checking) https://github.com/uhop/node-re2/issues/125

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,6 +150,11 @@
     "yaml": "^2.0.0",
     "zone.js": "^0.11.4"
   },
+  "dependenciesMeta": {
+    "re2": {
+      "built": false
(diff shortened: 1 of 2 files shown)


josephperrott (approved):
LGTM

We should do this in the other repos as well.

devversion:
@josephperrott As it seems, this only affects Yarn berry, because Yarn 1.x is able to just download the prebuilt binaries.

PR #665, opened 2022-06-29
refactor(ng-dev): do not capture Angular robot in release notes

devversion:
We now use the Angular robot for managing dependencies, and likely it
will be used more in the future. The bot commits should not end up
polluting release notes.


josephperrott (approved):
LGTM

PR #668, opened 2022-06-29
feat(github-actions): create org-file-sync action

josephperrott:
Create an org-file-sync action which allows for automatically synchronizing organization level files
from the `.github` directory with target repositories automatically.

The angular-robot Github app will automatically push changes to keep the registered files in sync.


github-actions/org-file-sync/src/main.ts
```diff
@@ -0,0 +1,123 @@
...
+  return github.rest.repos.getContent({owner: context.repo.owner, repo, path}).then(
+    (response) => {
+      if ((response.data as {content?: string}).content !== undefined) {
+        return response.data as File;
+      }
+      return NotFound;
+    },
+    (reason: RequestError) => {
+      if (reason.status === 404) {
+        return NotFound;
```
devversion, line 49: Could we just use an explicit `null` instead of dealing with a custom symbol?

github-actions/org-file-sync/src/main.ts
```diff
@@ -0,0 +1,123 @@
...
+ * Update the provided repo to ensure the provided golden file contents are used for the files
+ * with the same path in the repo.
+ */
+async function updateRepoWithFiles(github: Octokit, repo: string, goldenFiles: Files) {
+  /**
+   * Promises resolved when an update is completed.
+   * Note: An array is used to allow for awaiting everything via `Promise.all`
+   */
+  let updates: Promise<unknown>[] = [];
+  /** The current files, or lack of files, for syncronizing in provided repo. */
```
devversion, line 66:
```suggestion
  /** The current files, or lack of files, for synchronizing in provided repo. */
```

github-actions/org-file-sync/src/main.ts
```diff
@@ -0,0 +1,123 @@
...
+  const repoFiles = await getFilesForRepo(github, repo);
+
+  goldenFiles.forEach((goldenFile, path) => {
+    // If the golden file does not exist, we have nothing to syncronize.
+    if (goldenFile === NotFound) {
+      return;
+    }
+    /** The provided repository's File for the path. */
+    const repoFile = repoFiles.get(path) || NotFound;
+    /** The sha of the last time the file was updated in the provided repo. */
```
devversion, line 76: Nit: Potentially use consistent terminology here and refer to it as `target` repositories.

```suggestion
    /** The SHA of the last time the file was updated in the target repo. */
```

.prettierignore
```diff
@@ -8,6 +8,7 @@ github-actions/lock-closed/main.js
 github-actions/commit-message-based-labels/main.js
 github-actions/slash-commands/main.js
 github-actions/post-approval-changes/main.js
+github-actions/org-file-sync/main.js
```
devversion, line 11: Can we put this as a `local-actions`, or are we expecting this to run in other repositories as well? I guess I don't know a concrete use-case yet, but seems useful 😄

github-actions/org-file-sync/action.yml
```diff
@@ -0,0 +1,17 @@
+name: 'Organization File Sync'
+description: 'Syncronize files throughout the Github Organization'
+author: 'Angular'
+inputs:
+  angular-robot-key:
```
devversion, line 5: One important question I think: We now use Renovate with the `repo` and `workflow` access token for `angular-robot`. With the robot account being used in such cases (with write access to default branches), we are exposing more permissions to the Renovate processes. Should we have a separate Renovate account instead?

devversion (approved):

PR #671, opened 2022-06-30
refactor: reference other scripts through relative paths

devversion:
In order to resolve the build worker script, or the commit message
filter script we should simply compute an absolute path relative
to the current script path. This is better compared to using the CJS
require resolution (going away anyway at some point with ESM), because
it makes our `dev-infra/npm_package` work without necessarily being
installed inside the `node_modules` directory. This allows us to
conveniently run the tool inside the dev-infra repo (dogfooding) and
using it for testing from `HEAD` in other repositories (without risking
to have a mismatch from the dev-infra version actually in the node
modules).

### DIFF
--- a/ng-dev/pr/merge/strategies/autosquash-merge.ts
+++ b/ng-dev/pr/merge/strategies/autosquash-merge.ts
@@ -6,6 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {dirname, join} from 'path';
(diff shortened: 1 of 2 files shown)


josephperrott (approved):
LGTM

devversion:
This PR was merged into the repository by commit 050669a27b26f501c1b6e2a5680032927fba9d81.

PR #673, opened 2022-06-30
build: speed up local `ng-dev` and fix release build command

devversion:
Similar to the `ts-node` fixes we landed in all repositories, we need to
use Node directly to not break the build-worker NodeJS fork invocation.

Also we disable type-checking when `yarn ng-dev` runs, for speed.

### DIFF
--- a/tools/local-dev.sh
+++ b/tools/local-dev.sh
@@ -17,6 +17,9 @@ bazelCommand=${BAZEL:-"yarn bazel"}
   ${bazelCommand} build :npm_package
 )
 
+export TS_NODE_TRANSPILE_ONLY=1
+export TS_NODE_PROJECT=${PWD}/.ng-dev/tsconfig.json
+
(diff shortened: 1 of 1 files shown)


tools/local-dev.sh
```diff
@@ -17,6 +17,9 @@ bazelCommand=${BAZEL:-"yarn bazel"}
   ${bazelCommand} build :npm_package
 )
 
+export TS_NODE_TRANSPILE_ONLY=1
```
josephperrott, line 20: We don't have to export this right? Wouldn't have pollute the environment for all subsequent runs?

devversion: This script is never sourced, so it should only be local to this subshell.

josephperrott: Okay, just wanted to be sure.

josephperrott (approved):
LGTM

PR #705, opened 2022-07-11
feat(github-actions): skip angular-robot in postapproval action

devversion:
The Angular Robot is used for Renovate, and when Renovate pushes a small
update / or rebases upon conflicts, we should not need another explicit
review.

All actions performed by the `angular-robot` are managed by Googlers, so
this account should be able to skip the post approval checks.

### DIFF
--- a/github-actions/post-approval-changes/lib/main.ts
+++ b/github-actions/post-approval-changes/lib/main.ts
@@ -31,6 +31,8 @@ const googlers = [
   'zarend',
 ];
 
+const googleOwnedRobots = ['angular-robot'];
+
 async function main() {
   let installationClient: Octokit | null = null;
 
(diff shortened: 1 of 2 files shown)


josephperrott (approved):
LGTM

github-actions/post-approval-changes/lib/main.ts
```diff
@@ -57,6 +59,10 @@ async function runPostApprovalChangesAction(client: Octokit): Promise<void> {
     return;
   }
 
+  if (googleOwnedRobots.includes(pr.user.login)) {
+    core.info('PR author is a robot owned by Google. Post approval changes are allowed.');
```
gkalpak, line 63: Did you mean to return here or something?

devversion: No, just logging. Joke, you're right. good catch 🤦 too many drive-by things today!

PR #712, opened 2022-07-12
feat(bazel): expose individual `.map` files for optimized bundles

devversion:
Currently we do not declare individual file targets/pre-declared outputs
for map files, but we do for `.js` files. This is inconsistent and the
individual `map` files are useful to be "reference-able" when e.g. a
JS mapping size test is configured.

Enables source maps for compressed app bundles.

To achieve this we introduce a new rule for filtering outputs. This
rule will become helpful in the future as we often have scenarios where
`$(rootpath X)` cannot be used due to multiple outputs, even though e.g.
a genrule might only be interested in a single file for constructing a
command.


bazel/private/manifest_path.bzl
```diff
@@ -0,0 +1,21 @@
+"""Starlark helpers for Bazel manifest paths."""
+
+def get_manifest_path(ctx, file):
+    """Computes the manifest path for the given `File`.
+
+    For example: `../some/file` becomes `<wksp_name>/some/file` or
```
josephperrott, line 6: I think this would become `some/file` right?

josephperrott (approved):
LGTM

PR #713, opened 2022-07-12
fix(ng-dev): do not throw if there is no diff between `g3` and `main`

devversion:
It seems like multimatch previously did not throw when it received an
empty file path. We should never need to call multimatch with an empty
path, avoiding the runtime error. To fix this, we just do not
incorrectly try to derive diff information from an empty line if there
is no diff (this was a brittle logic).

### DIFF
--- a/ng-dev/caretaker/check/g3.ts
+++ b/ng-dev/caretaker/check/g3.ts
@@ -94,11 +94,18 @@ export class G3Module extends BaseModule<G3StatsData | void> {
     );
 
     // Get the numstat information between main and g3
-    this.git
+    const numStatDiff = this.git
       .run(['diff', `${g3Ref}...${mainRef}`, '--numstat'])
       .stdout // Remove the extra space after git's output.
-      .trim()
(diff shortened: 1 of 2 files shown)


ng-dev/caretaker/check/g3.ts
```diff
@@ -94,11 +94,18 @@ export class G3Module extends BaseModule<G3StatsData | void> {
...
+    const numStatDiff = this.git
       .run(['diff', `${g3Ref}...${mainRef}`, '--numstat'])
       .stdout // Remove the extra space after git's output.
-      .trim()
-      // Split each line of git output into array
+      .trim();
+
+    // If there is no diff, we can return early.
+    if (numStatDiff === '') {
+      return stats;
```
devversion, line 104: To clarify: Previously we were splitting on an potentially empty string, resulting in `[""]`. We then tried to determine the addition, deletions and file path from this line. This is simply incorrect and a bug.

jessicajaniuk (approved):

PR #718, opened 2022-07-14
fix(ng-dev/release): wait for final branch-off PR before finishing release action

devversion:
This should make it clear to the caretaker (even though we have logging
for it already in the release tool and merge tool) that the final
branch-off PR needs to be merged..


josephperrott (approved):
LGTM

Though hopefully this is rendered unnecessary by the tooling handling merges.

PR #728, opened 2022-07-19
refactor(ng-dev): clean-up merge tool error handling in preparation for fixes

devversion:
Cleans up the merge tool error handling by not passing through failures
and having uncaught errors at the same time. This results in confusion
and easily results in errors accidentally not being caught/handled
correctly.

Instead, we always use error try/catch and throwing similar to how its
done in the release tool. This is more safe, predictable and readable.

This is in preparation of fixing some of the force skip behavior of
the merge tool. This does currently not seem to work and broke at
some point..

### DIFF
--- a/ng-dev/pr/common/validation/pull-request-failure.ts
+++ b/ng-dev/pr/common/validation/pull-request-failure.ts
@@ -10,16 +10,17 @@ import {breakingChangeLabel} from '../../config/index.js';
 import {TargetLabel} from '../targeting/target-label.js';
 
 /**
- * Class that can be used to describe pull request failures. A failure
- * is described through a human-readable message and a flag indicating
- * whether it is non-fatal or not.
+ * Class that can be used to describe pull request failures.
+ *
+ * A failure is described by a human-readable message and a flag indicating
+ * whether the failure can be forcibly ignore or not.
  */
 export class PullRequestFailure {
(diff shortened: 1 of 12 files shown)


ng-dev/pr/common/validation/pull-request-failure.ts
```diff
@@ -10,16 +10,17 @@ import {breakingChangeLabel} from '../../config/index.js';
...
+ * A failure is described by a human-readable message and a flag indicating
+ * whether the failure can be forcibly ignore or not.
  */
 export class PullRequestFailure {
   constructor(
     /** Human-readable message for the failure */
     public message: string,
     /** Whether the failure is non-fatal and can be forcibly ignored. */
-    public nonFatal = false,
+    public canBeIgnoredForcibly = false,
```
josephperrott, line 23: What about just `canBeIgnored`, `canBeIgnoredForcibly` doesn't quite sound right to me.  Another option might be `canBeIgnoredWithForced`?

josephperrott (approved):
LGTM

ng-dev/pr/common/validation/pull-request-failure.ts
devversion: Changed to `canBeIgnoredNonFatal`

PR #729, opened 2022-07-20
feat(apps): create credential service server

josephperrott:
Create the credential service server. When logged in with an ng-dev token, allows for a user to request via websocket, a temporary Github credential.

### DIFF
--- a/apps/credential-service/BUILD.bazel
+++ b/apps/credential-service/BUILD.bazel
@@ -1,0 +1,29 @@
+load("//tools:defaults.bzl", "esbuild_esm_bundle")
+load("@build_bazel_rules_nodejs//:index.bzl", "copy_to_bin", "nodejs_binary")
+
+nodejs_binary(
+    name = "serve",
+    data = [
+        ":bin_files",
+    ],
+    entry_point = ":credential-service.js",
+)
+
+copy_to_bin(
+    name = "bin_files",
+    srcs = [
+        "package.json",
+    ],
+)
+
+esbuild_esm_bundle(
+    name = "credential-service",
+    entry_point = "//apps/credential-service/lib:server.ts",
+    target = "node16",
+    visibility = [
+        "//apps/credential-service:__subpackages__",
+    ],
+    deps = [
+        "//apps/credential-service/lib",
+    ],
+)
--- a/apps/credential-service/lib/server.ts
+++ b/apps/credential-service/lib/server.ts
@@ -1,0 +1,149 @@
+#!/usr/bin/env node
+
+import {createServer, IncomingMessage} from 'http';
+import {WebSocketServer, WebSocket} from 'ws';
+import {Octokit} from '@octokit/rest';
+import {createAppAuth} from '@octokit/auth-app';
+import {Duplex} from 'stream';
+import admin, {AppOptions} from 'firebase-admin';
+
+/** The temporary access token and a convience method for revoking it. */
+interface AccessTokenAndRevocation {
+  token: string;
+  revokeToken: () => void;
+}
+
+/** Regex for matching authorization header uses. */
+const authorizationRegex = new RegExp(/Bearer (.*)/);
+/** The length of time in ms between heartbeat checks. */
+const heartBeatIntervalLength = 5000;
+/** The port to bind the server to */
+const PORT = process.env.PORT!;
+/** The ID of the Github app used to generating tokens. */
+const GITHUB_APP_ID = process.env.GITHUB_APP_ID!;
+/** The PEM key of the Github app used to generating tokens. */
+const GITHUB_APP_PEM = process.env.GITHUB_APP_PEM;
+/** The firebase confgiuration for the firebase application being used for authentication. */
+const FIREBASE_APP_CONFIG = JSON.parse(process.env.FIREBASE_APP_CONFIG!) as AppOptions;
+
+// Initialize the Firebase application.
+admin.initializeApp(FIREBASE_APP_CONFIG);
+
+/** Generate a temporary access token with access to the requested repository. */
+export async function generateAccessToken(
+  owner: string,
+  repo: string,
+): Promise<AccessTokenAndRevocation> {
+  /** The github client used for generating the token. */
+  const github = new Octokit({
+    authStrategy: createAppAuth,
+    auth: {appId: GITHUB_APP_ID, privateKey: GITHUB_APP_PEM},
+  });
+  /** The specific installation id for the provided repository. */
+  const {id: installation_id} = (await github.apps.getRepoInstallation({owner, repo})).data;
+  /** A temporary github access token. */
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -113,6 +116,8 @@
     "cli-progress": "^3.7.0",
     "conventional-commits-parser": "^3.2.1",
     "ejs": "^3.1.6",
+    "express": "^4.18.1",
+    "express-ws": "^5.0.2",
     "firebase": "9.9.0",
```
devversion, line 121: Do we use express actually?

apps/credential-service/lib/server.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+
+/**
+ * WebSocket handler to generate temporary Github access token.
+ *
+ * The access token is automatically revoked when the websocket is closed.
+ */
+async function wsHandler(ws: WebSocket, req: IncomingMessage) {
+  /** Whether the websocket heartbeat check is still alive. */
+  let hasHeartbeat: boolean;
```
devversion, line 60:
```suggestion
  let receivedHeartbeatResponse: boolean;
```

apps/credential-service/lib/server.ts
```diff
@@ -0,0 +1,149 @@
...
+  } catch (e) {
+    console.error('Unable to verified authorized user');
+    console.error(e);
+    socket.write('HTTP/1.1 401 Unauthorized\r\n\r\n');
+    socket.destroy();
+    return;
+  }
+
+  /** The repository name and owner for the request. */
+  const {repo, owner} = req.headers as {repo: string; owner: string};
```
devversion, line 127: can we make the headers more scoped to the custom use-case. Like: NgDevTargetRepoName and so on?

apps/credential-service/lib/server.ts
```diff
@@ -0,0 +1,149 @@
...
+
+/** The http web server. */
+const server = createServer();
+/** The websocket server to handle websocket requests. */
+const wss = new WebSocketServer({noServer: true});
+
+wss.on('connection', wsHandler);
+server.on('upgrade', upgradeHandler);
+server.on('listening', () => console.log('Credential Service startup complete, listening'));
+server.listen(parseInt(PORT, 10));
```
devversion, line 149:
```suggestion
server.listen(Number(PORT));
```

devversion (commented):
Should there be a docker file?

package.json
josephperrott: Bad rebase caused it to stick around.

josephperrott:
Updated the commit message originally, but then Github decided to not update the title and summary for the PR when I pushed again.  Its corrected now.

apps/credential-service/BUILD.bazel
```diff
@@ -0,0 +1,29 @@
...
+
+copy_to_bin(
+    name = "bin_files",
+    srcs = [
+        "package.json",
+    ],
+)
+
+esbuild_esm_bundle(
+    name = "credential-service",
```
devversion, line 20: ooc: how do you intend to deploy the script?

apps/credential-service/lib/server.ts
```diff
@@ -0,0 +1,160 @@
...
+  };
+}
+
+/**
+ * WebSocket handler to generate temporary Github access token.
+ *
+ * The access token is automatically revoked when the websocket is closed.
+ */
+async function wsHandler(ws: WebSocket, req: IncomingMessage) {
+  /** Whether the websocket heartbeat check is still alive. */
```
devversion, line 64: nit: comment is not reflecting what it actually represents

apps/credential-service/lib/server.ts
```diff
@@ -0,0 +1,160 @@
...
+ * Handle upgrade requests before the websocket is used. Enforces authentication mechanisms
+ * and ensuring the required data is present in the request.
+ */
+async function upgradeHandler(req: IncomingMessage, socket: Duplex, head: Buffer) {
+  try {
+    if (!authorizationRegex.test(req.headers.authorization!)) {
+      throw Error('Missing or invalid authorization header syntax');
+    }
+    /** The NgDev token from the user to be verified. */
+    const [_, ngDevToken] = req.headers.authorization!.match(/Bearer (.*)/)!;
```
devversion, line 119: You have a regex for that above

apps/credential-service/lib/server.ts
```diff
@@ -0,0 +1,160 @@
...
+  ws.send(token);
+}
+
+/**
+ * Handle upgrade requests before the websocket is used. Enforces authentication mechanisms
+ * and ensuring the required data is present in the request.
+ */
+async function upgradeHandler(req: IncomingMessage, socket: Duplex, head: Buffer) {
+  try {
+    if (!authorizationRegex.test(req.headers.authorization!)) {
```
devversion, line 115: Super nit (while I leave actual comments): Is the exclamation needed? generally a good thing to avoid

devversion (approved):
LGTM, just some minor comments

apps/credential-service/BUILD.bazel
josephperrott: My plan is to handle it in a follow up, the plan is to run the service in a container.  I was looking at `rules_docker` but its set to be replaced so I am still sorting out specifics.

apps/credential-service/lib/server.ts
josephperrott: Switched it to a `|| ''`, which ends up functioning the same for `Regex.test`

apps/credential-service/BUILD.bazel
devversion: Sounds good. was mostly asking out of curiosity given the checked in script

apps/credential-service/lib/server.ts
```diff
@@ -112,11 +112,14 @@ async function wsHandler(ws: WebSocket, req: IncomingMessage) {
...
+    if (!authorizationRegex.test(req.headers.authorization || '')) {
       throw Error('Missing or invalid authorization header syntax');
     }
-    /** The NgDev token from the user to be verified. */
-    const [_, ngDevToken] = req.headers.authorization!.match(/Bearer (.*)/)!;
+    /**
+     * The NgDev token from the user to be verified.
+     * We use a non-null assertion as the regex was already tested above.
+     */
+    const [_, ngDevToken] = authorizationRegex.exec(req.headers.authorization!)!;
```
devversion, line 122: just FYI the whole non-nullability could be trivially inferred by TypeScript if you made an explicit check above

devversion (approved):
LGTM!

apps/credential-service/lib/server.ts
devversion: Also lint is failing

josephperrott:
This PR was merged into the repository by commit 1609543e7d76ccf66fbbab1bfa8ae4c2767adb80.

PR #730, opened 2022-07-20
fix(ng-dev): force prompt should only ignore the failed validation

devversion:
Currently when a validation check fails, the caretaker is prompted to
continue forcibly if the failure was non-fatal.

This seemed nice at first glance, but there is a real issue here. The
force merge run might hide actual validations which might have not been
hit before.

To fix this, we introduce an incremental validation system where we run
validations from a config (all by default) and upon failure, if
non-fatal the individual validation can be ignored, re-attempting the
merge.

Also as part of this the validation logic has been refactored/reworked
to individual files with some more complicated wrapper logic that
ensures we properly skip and report validation failures in a safe way.


devversion:
Note: This also enables us to skip the "changes allow for target label" validation, fixing issues like we had last time.

ng-dev/pr/merge/merge-pull-request.ts
```diff
@@ -36,17 +37,18 @@ export async function mergePullRequest(prNumber: number, flags: PullRequestMerge
...
   }
 
   /** Performs the merge and returns whether it was successful or not. */
-  async function performMerge(ignoreFatalErrors: boolean): Promise<boolean> {
+  async function performMerge(
+    validationConfig: PullRequestValidationConfig = new PullRequestValidationConfig(),
+  ): Promise<boolean> {
     try {
-      await tool.merge(prNumber, /* force */ ignoreFatalErrors);
+      await tool.merge(prNumber, validationConfig);
```
josephperrott, line 51: Rather than recursively calling on an overall failure, would it make more sense to have the force prompt occur from within the validation?

In the `PullRequestValidation` abstract class you could expand it to:

```ts
export abstract class PullRequestValidation {
  abstract canBeForced: boolean;

  constructor(
    protected name: keyof PullRequestValidationConfig,
    protected _createError: PullRequestValidationErrorCreateFn,
  ) {}

  abstract internalAssert(pullRequest: PullRequestFromGithub): Promise<void>;

  async assert(pullRequest: PullRequestFromGithub) {
    try {
      this.internalAssert(pullRequest);
    } catch (e) {
      if (this.canBeForced && await prompt('Continue anyway?')) {
        return;
      }
      throw e;
    }
  }
}

```

Then the forcing logic becomes localize to the specific validation, and we can continue on directly from there.

josephperrott (commented):
I love the direction of this but had a question about how we are going about it.

ng-dev/pr/merge/merge-pull-request.ts
devversion: Done. I love it! that made this much more simple and even more readable.

One note: Instead of making the internal assert method typed in the base class, I'm leaving that to the consumer because every validation has different parameters and typing this with a generic (and combining the args with the validation config) is not worth the effort.

josephperrott (approved):
LGTM

devversion:
This PR was merged into the repository by commit 0f1cacea776de61d0ed7e187a11e1be9b0a1fe0a.

PR #733, opened 2022-07-22
Create `--dry-run` for `pr merge`

josephperrott:
See individual commits.

### DIFF
--- a/ng-dev/pr/merge/merge-tool.ts
+++ b/ng-dev/pr/merge/merge-tool.ts
@@ -64,6 +64,7 @@ export class MergeTool {
    * @param validationConfig Pull request validation config. Can be modified to skip
    *   certain non-fatal validations.
    */
+
   async merge(prNumber: number, validationConfig: PullRequestValidationConfig): Promise<void> {
     if (this.git.hasUncommittedChanges()) {
       throw new FatalMergeToolError(
@@ -112,15 +113,6 @@ export class MergeTool {
       await this.updatePullRequestTargetedBranchesFromPrompt(pullRequest);
     }
 
-    if (
-      // In cases where manual branch targeting is used, the user already confirmed.
-      !this.flags.forceManualBranches &&
-      this.flags.branchPrompt &&
-      !(await Prompt.confirm(getTargetedBranchesConfirmationPromptMessage(pullRequest)))
-    ) {
-      throw new UserAbortedMergeToolError();
-    }
-
     // If the pull request has a caretaker note applied, raise awareness by prompting
     // the caretaker. The caretaker can then decide to proceed or abort the merge.
     if (
@@ -143,6 +135,18 @@ export class MergeTool {
     try {
       // Run preparations for the merge (e.g. fetching branches).
       await strategy.prepare(pullRequest);
+
+      // Check for conflicts between the pull request and target branches.
+      await strategy.check(pullRequest);
+
+      if (
+        // In cases where manual branch targeting is used, the user already confirmed.
+        !this.flags.forceManualBranches &&
+        this.flags.branchPrompt &&
+        !(await Prompt.confirm(getTargetedBranchesConfirmationPromptMessage(pullRequest)))
+      ) {
+        throw new UserAbortedMergeToolError();
+      }
 
       // Perform the merge and pass-through potential failures.
       await strategy.merge(pullRequest);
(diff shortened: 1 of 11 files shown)


ng-dev/pr/merge/merge-tool.ts
```diff
@@ -143,8 +135,26 @@ export class MergeTool {
       // Run preparations for the merge (e.g. fetching branches).
       await strategy.prepare(pullRequest);
 
+      // Check for conflicts between the pull request and target branches.
+      await strategy.check(pullRequest);
+
+      if (this.flags.dryRun) {
+        Log.info(green(`  ✓  Dry run sucessful, pull request merges as expected.`));
```
devversion, line 142: In the prompt IIRC we print the target branches. If this is true, we should move this out and print it in a single place, even with dry run mode

devversion: Might be worth always stating that the PR merges cleanly, and just exiting early with dry-mode. That way it would be clear before the prompt that the mergability has already been checked

ng-dev/pr/merge/strategies/api-merge.ts
```diff
@@ -55,14 +54,19 @@ export class GithubApiMergeStrategy extends MergeStrategy {
       throw PullRequestFailure.unsatisfiedBaseSha();
     }
 
-    const method = this._getMergeActionFromPullRequest(pullRequest);
     const cherryPickTargetBranches = targetBranches.filter((b) => b !== githubTargetBranch);
```
devversion, line 57: We should avoid duplicating this logic and have a private class function for it

ng-dev/pr/merge/strategies/autosquash-merge.ts
```diff
@@ -30,9 +30,8 @@ export class AutosquashMergeStrategy extends MergeStrategy {
    * @throws {PullRequestFailure} A pull request failure if the the pull request could not
    *   be merged and the pull request is misconfigured.
    */
-  override async merge(pullRequest: PullRequest): Promise<void> {
-    const {prNumber, targetBranches, requiredBaseSha, needsCommitMessageFixup, githubTargetBranch} =
-      pullRequest;
+  override async check(pullRequest: PullRequest): Promise<void> {
+    const {prNumber, targetBranches, requiredBaseSha, needsCommitMessageFixup} = pullRequest;
     // In case a required base is specified for this pull request, check if the pull
```
devversion, line 35: We seem to do much more than checking here, that seems unfortunate. Also the merge function below depends on the side effects from check before. Should we just check consistently outside of strategies similar to how we do it in the API merge?

ng-dev/pr/merge/merge-tool.ts
josephperrott: I updated the message as suggested, though they will not see the prompt because we `return` out of the method before we continue on to it.

ng-dev/pr/merge/strategies/autosquash-merge.ts
josephperrott: I extracted this check concept into the base merge strategy class, since checking if we can merge a pull request shouldn't actually be based on the type of merge.

Instead the merge strategies individually can just perform the actual merge.

ng-dev/pr/merge/pull-request.ts
```diff
@@ -106,16 +112,34 @@ export async function loadAndValidatePullRequest(
...
     labels,
     requiredBaseSha,
     githubTargetBranch,
     needsCommitMessageFixup,
+    commitCount,
+    baseSha,
+    revisionRange,
     hasCaretakerNote,
     targetBranches: target.branches,
+    cherryPickTargetBranches: target.branches.filter((b) => b !== githubTargetBranch),
```
devversion, line 142: Since this is now only used in the API merge, we can keep defining this in the API strategy there. Mostly to avoid duplicating the target branches in two places as part of the `PullRequest` interface

ng-dev/pr/merge/merge-tool.ts
```diff
@@ -144,8 +136,26 @@ export class MergeTool {
       // Run preparations for the merge (e.g. fetching branches).
       await strategy.prepare(pullRequest);
 
+      // Check for conflicts between the pull request and target branches.
+      await strategy.check(pullRequest);
+
```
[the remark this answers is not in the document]
josephperrott: Good call, added it in.

devversion:
This PR was merged into the repository by commit 418ef131adfa37f2fced5df9d042d3e497073f75.

devversion:
FYI: I just improved the logging a little bit, and also avoided the second `git fetch`. This was just some clean-up so I went ahead and merged the PR since it looked great aside from the second `git fetch`.

PR #738, opened 2022-07-24
feat(github-actions): implement GitHub Action for creating a PR from local changes

gkalpak:
Implement a new GitHub Action, `create-pr-for-changes` that can check for changes in the working copy and create a pull request. It can skip creating a pull request if there are no local changes or if a pull request already exists for the same changes and branches (based on some heuristic - see `github-actions/create-pr-for-changes/lib/main.ts` for details).

This action can be useful for periodically updating checked-in files that are auto-generated based on data from external sources (such as events stored in a database).

##
FYI, I have tested the new action with angular/angular#45588 (at commit [c3e545f][1]) with a [slight modification][2]: using the GitHub Action provided token instead of an `angular-robot-key`, since the Angular Robot is not installed on my fork.
Here are a couple of runs:
- [A run][3], which resulted in PR gkalpak/angular#12.
- [Another run][4], which did not create a new PR, since the changes were identical to PR gkalpak/angular#12.

[1]: https://github.com/angular/angular/pull/45588/commits/c3e545f6a0f61907f06e0cc85d3a1f0d40909015
[2]: https://github.com/gkalpak/dev-infra/commit/9b6e75075b398f95aee5d7a6e43a86ea7aab0f35#diff-0416a58f993e853f85fbb6602354b3380396182b1ea7318423457032076568f3R41
[3]: https://github.com/gkalpak/angular/runs/7488852623?check_suite_focus=true
[4]: https://github.com/gkalpak/angular/runs/7488873829?check_suite_focus=true


github-actions/create-pr-for-changes/lib/main.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    // Create a branch name based on the provided branch-prefix and the contents of the changed
+    // files.
+    //
+    // NOTE:
+    //   Hashing the contents of the changed files is a heuristic to uniquely-ish represent a
+    //   particular set of changes. It is not 100% accurate (and could result in both false
+    //   positives and false negatives), but should be good enough for our purposes.
+    const branchPrefix = core.getInput('branch-prefix', {required: true});
+    const branchName = `${branchPrefix}-${hashFiles(touchedFiles)}`;
```
josephperrott, line 72: Why not derive the branch name using the commit sha?

`${branchPrefix}-${git.run(['rev-parse', 'HEAD'])}`

We could then rely on Git to create something that is even more likely to be unique.

github-actions/create-pr-for-changes/lib/main.ts
```diff
@@ -0,0 +1,204 @@
...
+      .getInput('pr-labels', {required: false})
+      .split('\n')
+      .map((l) => l.trim())
+      .filter((l) => l !== '');
+    const commitMessage = `${prTitle}\n\n${prBody}`;
+
+    git.run(['checkout', '-b', branchName]);
+    git.run(['add', '--all']);
+    git.run(['commit', '-m', commitMessage, '--no-verify']);
+    git.run(['push', '--force', git.getRepoGitUrl(), `HEAD:refs/heads/${branchName}`]);
```
josephperrott, line 135: This shouldn't require a force if its a new branch.

github-actions/create-pr-for-changes/lib/main.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    core.info('Committed changes.');
```
josephperrott, line 137:
```suggestion
    core.info('Committed changes and pushed to Github.');
```

github-actions/create-pr-for-changes/lib/main.ts
```diff
@@ -0,0 +1,204 @@
...
+        labels: prLabels,
+      });
+
+      core.info(`Added ${prLabels.length} label(s) to PR: '${prLabels.join("', '")}'`);
+    }
+
+    // Comment on superseded PRs.
+    if (supersededPrs.length > 0) {
+      for (const pr of supersededPrs) {
+        await git.github.issues.createComment({
```
josephperrott, line 169: This should be closing the PR right not just noting it superceded, or at least we should support that as an option.

github-actions/create-pr-for-changes/lib/main.ts
josephperrott, line 135: I would prefer these pull commits were placed in the accounts fork.

github-actions/create-pr-for-changes/lib/main.ts
```diff
@@ -0,0 +1,204 @@
...
+          body: `Superseded by PR #${createdPr.number}.`,
+        });
+
+        // TODO(gkalpak):
+        //   Do we want to add/remove labels on superseded PRs?
+        //   For example, we could add `state: blocked` and remove `action: merge[-assistance]` by
+        //   default (or we could make these configurable).
+      }
+
+      core.info('Commented on superseded PRs.');
```
josephperrott, line 182:
```suggestion
      core.info('Commented on ${supersededPrs.length} superseded PRs.');
```

github-actions/create-pr-for-changes/lib/main.ts
gkalpak: This is not just about being unique, but about uniquely identifying a set of changes (regardless of other changes that might have happened to the repo).

Imagine that we use this action to modify file `foo.txt`. I can have two separate SHAs that both apply the same change to file `foo.txt` but are different because their parent SHA is different (due to the unrelated file `bar.txt` being modified between the two). In such a case, we want to recognize the both commits apply the same change, even though their SHAs are different.

Does that make sense?

github-actions/create-pr-for-changes/lib/main.ts
gkalpak: I used `--force`, because it is possible that the branch exists on the remote (for example, from a previous invocation of the action).

But I think it would be better to clean-up matching branches that do not correspond to open PRs (which will also make `--force` redundant).

github-actions/create-pr-for-changes/lib/main.ts
gkalpak: My reasoning for not closing the PRs directly is to account for situations like the following:
- PR #1 is created which has some changes.
- PR #1 passes CI checks, looks good for merging, but is not yet merged.
- On the next run, PR #2 is created which has more changes than PR #1.
- It turns out that the changes in PR #2 are not valid and cause failures.
- At this point, we might want to get PR #1 in (which includes useful updates), while we address the issues that cause PR #2 to fail.

All superseded PRs are referenced in the superseding PR's commit message and will be automatically closed once the superseding PR is merged.

##
I personally like this workflow, but I understand it might be confusing for people that are not familiar with how the action works. Happy to change this to close the superseded PRs instead (or support both options).

WDYT?

josephperrott: That makes sense, I think I will need to see it in action before I can really judge it. What you laid out makes sense, I think I worry about how deep of superseding we could end up with.  Lets go with this and if we determine that its better to close we couldn't be able to do that.

github-actions/create-pr-for-changes/lib/main.ts
josephperrott: Yeah, that makes sense. Will this hold for an action that doesn't always change all of the same files though?  For instance if you have an action that updates all of the generated files, but some of them are unchanged.

josephperrott (approved):
LGTM

Just the one thing to resolve and be sure about.

github-actions/create-pr-for-changes/lib/main.ts
gkalpak: The current hashing looks at the files modified by the action run and their content. If there are two different runs that result in the same files being touched and the contents of those touched files are the same on both runs, then the two runs will be considered the same.

If you have two runs that result in a different set of files being touched (or if the contents of the touched files has changed between the two runs), then they will be considered different.

This can result in "misses" (i.e. two runs that should be identified as the same being considered different), but that is not a big problem: It just means that we might get two PRs that do the same thing.

PR #739, opened 2022-07-25
build: split up tooling into `@angular/build-tooling` and `@angular/ng-dev`

devversion:
This will allow us to easily update `ng-dev` without necessarily needing
to update the Bazel setup. This is always more difficult e.g. when an
LTS branch needs release tool updates but the LTS branch is e.g.
multiple majors behind of `main` with regards to e.g. `rules_nodejs`.


.circleci/config.yml
```diff
@@ -192,8 +192,8 @@ jobs:
       - setup_bazel_config
       - yarn_install
       - run:
-          name: Build npm package
-          command: yarn bazel build //:npm_package --config=release
+          name: Build npm packages
```
josephperrott, line 195: Maybe in a follow up, we should consider having this build everything with something like a `releasable_pkg` tag or something like that.

.circleci/publish_to_github.sh
```diff
@@ -18,11 +18,16 @@ if [ -z ${DEV_INFRA_BUILDS_TOKEN} ]; then
...
-  packageName="dev-infra-private"
-  packageRepo="dev-infra-private-builds"
-  buildDir="$(pwd)/dist/bin/npm_package"
-  buildVersion=$(node -pe "require('./package.json').version")
+  packageName="${1}"
+  packageRepo="${2}"
+  packageDistDir="${3}"
+
+  buildVersion=$(cd ${packageDistDir} && node -pe "require('./package.json').version")
   branchName=${CIRCLE_BRANCH:-'master'}
```
josephperrott, line 31:
```suggestion
  branchName=${CIRCLE_BRANCH:-'main'}
```

.circleci/publish_to_github.sh
```diff
@@ -83,5 +88,5 @@ publishPackage() {
   fi
 }
 
-# Run the publish package function
-publishPackage
+publishPackage "@angular/build-tooling" "dev-infra-private-build-tooling-builds" "dist/bin/npm_package"
```
josephperrott, line 91: Should we change this to script taking in the parameters? and then call the specifics from the config.yml file?

BUILD.bazel
```diff
@@ -38,21 +38,18 @@ package_group(
 
 pkg_npm(
     name = "npm_package",
-    package_name = "@angular/dev-infra-private",
+    package_name = "@angular/build-tooling",
```
josephperrott, line 41: We need to get signoff from @jelbourn, @alxhub and @dgp1130 on naming this.

Also, would it make sense to `_` prefix?  `@angular/_build-tooling`

package.json
```diff
@@ -1,9 +1,6 @@
 {
-  "name": "@angular/dev-infra-private",
+  "name": "@angular/build-tooling",
```
josephperrott, line 2: Should we create this package in its own package.json file?

BUILD.bazel
devversion: I thought we were good on this, remembering our discussion in the sync:

1. These names are never going on NPM
2. The repositories exposing these internal packages are explicitly named to contain `-private-`
3. If e.g. CLI ever intends to use this, we can just change the name, but for now we use this locally through snapshot repos and want it to be short and concise

package.json
devversion: I considered it, but it will be annoying as we would need to repeat dependencies etc.

A more invasive approach would be to restructure the whole repo:

- src/ng-dev
- src/build-tooling/...

.circleci/config.yml
devversion: I don't think that would be worth doing, since it's just two packages.

.circleci/publish_to_github.sh
devversion: That seems like a nice idea, will do

BUILD.bazel
josephperrott: Yeah, that was the feedback @jelbourn just gave me as well.

package.json
josephperrott: Okay, something to probably keep in mind in the future, but lets leave it for now.

josephperrott (approved):
LGTM

devversion:
This PR was merged into the repository by commit e6d1537069baaf49bf6ef72c306087b0a9e39b20.

PR #752, opened 2022-08-03
fix(apps): misc fixes from testing

josephperrott:
Properly revoke the installation token, use a static port of 8080, extend the heartbeat interval
length to 10 seconds, improve request header names.

### DIFF
--- a/apps/credential-service/lib/server.ts
+++ b/apps/credential-service/lib/server.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-import {createServer, IncomingMessage} from 'http';
+import {createServer, IncomingHttpHeaders, IncomingMessage} from 'http';
 import {WebSocketServer, WebSocket} from 'ws';
 import {Octokit} from '@octokit/rest';
 import {createAppAuth} from '@octokit/auth-app';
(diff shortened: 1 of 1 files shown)


apps/credential-service/lib/server.ts
```diff
@@ -14,13 +14,17 @@ interface AccessTokenAndRevocation {
...
 
+interface RequestParameterHeaders extends IncomingHttpHeaders {
+  ng_repo_name: string;
+  ng_repo_owner: string;
+}
+
 /** Regex for matching authorization header uses. */
 const authorizationRegex = new RegExp(/Bearer (.*)/);
 /** The length of time in ms between heartbeat checks. */
-const heartBeatIntervalLength = 5000;
```
devversion, line 20: What's the exact reasoning for the increase, to have it captured?

devversion (approved):

apps/credential-service/lib/server.ts
josephperrott: Added into the comment, essentially we are accounting for local processing where we block the thread because JS is single threaded.

apps/credential-service/lib/server.ts
```diff
@@ -21,7 +21,12 @@ interface RequestParameterHeaders extends IncomingHttpHeaders {
 
 /** Regex for matching authorization header uses. */
 const authorizationRegex = new RegExp(/Bearer (.*)/);
-/** The length of time in ms between heartbeat checks. */
+/**
+ * The length of time in ms between heartbeat checks.
+ *
+ * 10 seconds is used as during the local PR validation process, the javascript thread
+ * becomes blocked, and can cause a timeout.
+ */
```
devversion, line 29: Ohh, that is pretty interesting. Couldn't we offload the heartbeat somehow? It seems pretty brittle that we need a large enough timeout here (especially if we add stuff in the future / use it for other tools).

josephperrott: Yeah, I think it makes sense in a follow up to do once we have everything in place as part of larger testing.

devversion: Sounds good

PR #754, opened 2022-08-04
ci: allow any command for Renovate post-upgrade tasks

gkalpak:
The [allowedPostUpgradeCommands][1] pattern list is used to determine whether a [post-upgrade task][2] command is allowed to be run by Renovate. Using a regex to try and limit the type of commands that can be run is not a viable strategy.

This commit allows any command to be run in a post-upgrade task, which saves us the trouble of having to update the allowed commands pattern list every time we want to add a new post-upgrade task. Since adding/modifying Renovate post-upgrade tasks are subject to the regular PR review workflow, unwanted commands will be caught there.

##
_INFO: [Internal discussion][3]_

[1]: https://docs.renovatebot.com/self-hosted-configuration/#allowedpostupgradecommands
[2]: https://docs.renovatebot.com/configuration-options/#postupgradetasks
[3]: https://angular-team.slack.com/archives/C029MM6RS4B/p1659605197808499

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -18,10 +18,10 @@ http_archive(
 
 http_archive(
     name = "io_bazel_rules_sass",
-    sha256 = "fcd963afb0744685ee798f991942ff02196ae71ccbeedd8efaac17040dd99989",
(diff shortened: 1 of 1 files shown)


devversion (approved):

gkalpak:
New PR: #782

PR #757, opened 2022-08-04
create serve script for credential-service and use base64 for environment variable

josephperrott:
See individual commits.

### DIFF
--- a/apps/credential-service/deploy.sh
+++ b/apps/credential-service/deploy.sh
@@ -1,7 +1,7 @@
 if [ -z $(which gcloud) ]; then
   echo "###################################################################"
   echo "# Failed:                                                         #"
-  echo "#  gcloud must be installed in order to deploy this service        #"
+  echo "#  gcloud must be installed in order to deploy this service       #"
   echo "###################################################################"
   exit 1;
 fi
--- a/apps/credential-service/deploy.sh
+++ b/apps/credential-service/deploy.sh
@@ -1,0 +1,9 @@
+if [ -z $(which gcloud) ]; then
+  echo "###################################################################"
+  echo "# Failed:                                                         #"
+  echo "#  gcloud must be installed in order to deploy this service        #"
+  echo "###################################################################"
+  exit 1;
+fi
+
+gcloud run deploy credential-service --platform=managed --region=us-central1 --source $(dirname "$0")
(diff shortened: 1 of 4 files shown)


apps/credential-service/lib/server.ts
```diff
@@ -35,7 +35,7 @@ assert(process.env.GITHUB_APP_ID, 'GITHUB_APP_ID is not defined in the environme
 const GITHUB_APP_ID = process.env.GITHUB_APP_ID;
 /** The PEM key of the Github app used to generating tokens. */
 assert(process.env.GITHUB_APP_PEM, 'GITHUB_APP_PEM is not defined in the environment');
-const GITHUB_APP_PEM = process.env.GITHUB_APP_PEM;
+const GITHUB_APP_PEM = Buffer.from(process.env.GITHUB_APP_PEM, 'base64').toString('utf-8');
```
devversion, line 38: Any reason to encode the secret as base64 in the environment? Please clarify in code/or here, and also rename the environment variable to e.g. `GITHUB_APP_PRIVATE_KEY_BASE64`

apps/credential-service/deploy.sh
```diff
@@ -0,0 +1,9 @@
+if [ -z $(which gcloud) ]; then
+  echo "###################################################################"
+  echo "# Failed:                                                         #"
+  echo "#  gcloud must be installed in order to deploy this service        #"
```
devversion, line 4:
```suggestion
  echo "#  gcloud must be installed in order to deploy this service       #"
```

apps/credential-service/lib/server.ts
josephperrott: Done

apps/credential-service/lib/server.ts
```diff
@@ -33,9 +33,12 @@ const PORT = 8080;
...
 assert(process.env.GITHUB_APP_ID, 'GITHUB_APP_ID is not defined in the environment');
 const GITHUB_APP_ID = process.env.GITHUB_APP_ID;
-/** The PEM key of the Github app used to generating tokens. */
-assert(process.env.GITHUB_APP_PEM, 'GITHUB_APP_PEM is not defined in the environment');
-const GITHUB_APP_PEM = process.env.GITHUB_APP_PEM;
+/**
+ * The PEM key of the Github app used to generating tokens, the value is stored in the environment
+ * variable in Base64 to make it easier to manage it being a multiline string.
+ */
+assert(process.env.GITHUB_APP_PEM_BASE64, 'GITHUB_APP_PEM is not defined in the environment');
```
devversion, line 40:
```suggestion
assert(process.env.GITHUB_APP_PEM_BASE64, '`GITHUB_APP_PEM_BASE64` is not defined in the environment');
```

devversion (approved):

PR #764, opened 2022-08-10
Move ng-service enablement to .ng-dev github config, enable it for this repo. 

josephperrott:


ng-dev/utils/ng-dev-service.ts
```diff
@@ -30,49 +32,56 @@ const firebaseConfig = {
...
 
-        Log.error('  ✘  You must be logged in to run this command\n');
-        Log.log('Log in by running the following command:');
-        Log.log('  yarn ng-dev auth login');
-        argv.exit(1, new Error('The user is not logged in'));
-      })
+          Log.error('  ✘  You must be logged in to run this command\n');
+          Log.log('Log in by running the following command:');
+          Log.log('  yarn ng-dev auth login');
+          argv.exit(1, new Error('The user is not logged in'));
```
devversion, line 83: I don't think we use `argv.exit` anywhere else. Can we just throw the error like elsewhere and Yargs will pick it up?

devversion (approved):

devversion (approved):

ng-dev/utils/ng-dev-service.ts
josephperrott: Done.  I will come back to how we handle exiting as we currently are very inconsistent in a follow up.

.ng-dev/github.mts
```diff
@@ -5,4 +5,5 @@ export const github: GithubConfig = {
   owner: 'angular',
   name: 'dev-infra',
   mainBranchName: 'main',
+  useNgDevService: true,
```
devversion, line 8: I wonder if we should name this

```suggestion
  useNgDevAuthService: true,
```

devversion: nothing blocking, but just as an idea

PR #765, opened 2022-08-10
feat(ng-dev): create system for registering functions to be called on command completion

josephperrott:
Create a callback system to call functions after a command completes.

### DIFF
--- a/ng-dev/utils/yargs.ts
+++ b/ng-dev/utils/yargs.ts
@@ -1,5 +1,4 @@
 import yargs, {Arguments, Argv} from 'yargs';
-import {Log} from './logging.js';
 
 // A function to be called when the command completes.
 type CompletedFn = (err: Error | null) => Promise<void> | void;
@@ -12,34 +11,24 @@ export function registerCompletedFunction(fn: CompletedFn) {
   completedFunctions.push(fn);
 }
 
-/** Error to be thrown when yargs completes without running a command. */
-export class YargsError extends Error {
-  constructor() {
-    super('Error is parse or validation of command');
-  }
-}
-
 /**
  * Run the yargs process, as configured by the supplied function, calling a set of completion
  * functions after the command completes.
  */
-export function runParserWithCompletedFunctions(applyConfiguration: (argv: Argv) => Argv) {
-  applyConfiguration(yargs([])).parse(
-    process.argv.slice(2),
-    async (err: Error | null, _: Arguments, output: string) => {
-      if (err && [undefined, 0].includes(process.exitCode)) {
-        process.exitCode = 1;
-      }
-      if (err) {
-        Log.debug(err);
-      }
-      if (output) {
-        err = err || new YargsError();
-        Log.log(output);
-      }
-      for (const completedFunc of completedFunctions) {
-        await completedFunc(err);
-      }
-    },
-  );
+export async function runParserWithCompletedFunctions(applyConfiguration: (argv: Argv) => Argv) {
+  let err: Error | null = null;
+  try {
+    await applyConfiguration(yargs(process.argv.slice(2)))
+      .exitProcess(false)
+      .parse();
+  } catch (e) {
+    err = e as Error;
+    if ([undefined, 0].includes(process.exitCode)) {
+      process.exitCode = 1;
+    }
+  } finally {
+    for (const completedFunc of completedFunctions) {
+      await completedFunc(err);
+    }
+  }
 }
--- a/ng-dev/utils/yargs.ts
+++ b/ng-dev/utils/yargs.ts
@@ -1,4 +1,5 @@
 import yargs, {Arguments, Argv} from 'yargs';
+import {Log} from './logging.js';
 
 // A function to be called when the command completes.
 type CompletedFn = (err: Error | null) => Promise<void> | void;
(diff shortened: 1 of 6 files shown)


ng-dev/utils/yargs.ts
```diff
@@ -0,0 +1,26 @@
...
+  completedFunctions.push(fn);
+}
+
+/**
+ * Run the yargs process, as configured by the supplied function, calling a set of completion
+ * functions after the command completes.
+ */
+export function runParserWithCompletedFunctions(applyConfiguration: (argv: Argv) => Argv) {
+  applyConfiguration(yargs(process.argv.slice(2)))
+    .exitProcess(false)
```
devversion, line 20: From the docs:

>.exitProcess(false) should not be used when [.command()](https://yargs.js.org/docs/#command) is called with a handler returning a promise, as it would lead to a duplicated error message when this promise rejects

is that relevant here

ng-dev/utils/yargs.ts
```diff
@@ -0,0 +1,26 @@
...
+    .parse(process.argv.slice(2), async (err: Error | null) => {
```
devversion, line 21: In the docs there is also an `output` parameter. It sounds like this one contains necessary information that would otherwise be printed? I think we need to be careful about not causing unexpected scenarios here.

Have you considered just keeping the register approach (I like that), but then just subscribing to `process.on('exit'` or something?

Also what about the exit code if `err` is not `null`?

ng-dev/utils/yargs.ts
josephperrott: Yes, it does in some situations.  Removed.

ng-dev/utils/yargs.ts
josephperrott: `output` contains the string that has been printed to the console throughout the command, so we don't need its information.

The problem using `process.on('exit'` is that if something prevents the `exit` from occurring then we don't every run the completed functions.  We want to tie the completed functions to the completion of the command specifically.

The error would have been intentionally thrown, but by providing it to the completed functions, functions can act upon the error that was thrown.

devversion: I don't know if Yarn could capture our log messages as it doesn't patch e.g. console.log.

Quote from the docs

> output: any text that would have been output by yargs to the terminal, had a callback not been provided

Sounds like this is something different? 

Makes sense with the explanation on `process.exit`. Good point.

josephperrott: I see what you are saying, I think I tested that and then ended up switching things around.  Updated.

ng-dev/utils/yargs.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Run the yargs process, as configured by the supplied function, calling a set of completion
+ * functions after the command completes.
+ */
+export function runParserWithCompletedFunctions(applyConfiguration: (argv: Argv) => Argv) {
+  applyConfiguration(yargs([])).parse(
+    process.argv.slice(2),
+    async (err: Error | null, _: Arguments, output: string) => {
+      if (output) {
+        err = new YargsError();
+        Log.log(output);
```
devversion, line 32: Sorry for being a little picky here, but Yargs is very imprecise on `.parse` as it seems. My thinking (correct me if I'm wrong) is that we should:

* Always log `output` if there is something (it doesn't need to be for an error technically)
* If there is an `err`, we print it? (not sure about that -- what did Yarn do by default?)
* If there is an `err`, we set the `process.exitCode` to `1` (because Yarn no longer exits on errors right?)

josephperrott: I updated again, let me know how this looks to you.

devversion: Looks close, subjectively. I would be interested in what you actually think is the right thing to do here. Only thing that looks suspicious too me right now is that we build an err if there is an output (does output always indicate an error?)

josephperrott: Output always indicates that Yargs did not run the handler for a command, whether than be cause of a a `--help` flag or because of a validation error.

Since calling with `--help` is not an error, we are creating an error to represent that the command did not run.

devversion: If that's the case, then this seems reasonable. From the documentation it doesn't seem to indicate that strictly the handler wasn't invoked though.

Don't want to overcomplicate this, feel free to merge.

ng-dev/utils/yargs.ts
devversion: I read more up and it seems like the most idiomatic solution would really be to just use a top level async/await and try/catch/finally.

https://github.com/yargs/yargs/pull/1823

Let's go with this 😆 but we can clean this up as it seems. I would like the more obvious/predictable top level await more

ng-dev/utils/yargs.ts
devversion: I read more up and it seems like the most idiomatic solution would really be to just use a top level async/await and try/catch/finally.

https://github.com/yargs/yargs/pull/1823

I would like the more obvious/predictable top level await more, but non blocking

josephperrott: Okay, I moved to the top level await method, let me know what you think.

ng-dev/utils/yargs.ts
```diff
@@ -0,0 +1,31 @@
...
+ * functions after the command completes.
+ */
+export async function runParserWithCompletedFunctions(applyConfiguration: (argv: Argv) => Argv) {
+  let err: Error | null = null;
+  try {
+    await applyConfiguration(yargs(process.argv.slice(2)))
+      .exitProcess(false)
+      .parse();
+  } catch (e) {
+    err = e as Error;
```
devversion, line 25: Thank you, that looks awesome now. Final
thing: Do you want to set the exit code if undefined here? or would you rather re-throw the error here and let cli.mts handle the error scenario?

devversion (approved):

PR #766, opened 2022-08-11
fix(ng-dev): properly encode URLs for GitHub queries in caretaker check

devversion:
The URLs constructed by `caretaker/github.ts` were incorrectly escaped
and ended up e.g. not escaping a colon in URLs.


pkozlowski-opensource (approved):
LGTM

PR #772, opened 2022-08-16
feat(ng-dev/ts-circular-dependencies): support ignoring type only imports/exports for circular dependency checks

josephperrott:
By setting a configuration, imports/exports which use the `type` keyword to indicate they are type-only imports/exports
are ignored with respect to circular dependency checks.

This circular dependency would be considered a circular dependency by the tool:

**File 1:**
```ts
import type {symbol} from './file-2';

export const coolSymbol: boolean = true;
...
```

**File 2:**
```ts
import {coolSymbol} from './file-1';

export const symbol: Symbol =  Symbol();
...
```

By enabling `ignoreTypeOnlyChecks`, the `type` import is ignored, resulting in no circular dependency.

### DIFF
--- a/ng-dev/ts-circular-dependencies/analyzer.ts
+++ b/ng-dev/ts-circular-dependencies/analyzer.ts
@@ -40,7 +40,7 @@ export class Analyzer {
 
   constructor(
     public resolveModuleFn?: ModuleResolver,
-    {ignoreTypeOnlyChecks}: CircularDependenciesParserOptions = {},
+    ignoreTypeOnlyChecks: boolean = false,
     public extensions: string[] = DEFAULT_EXTENSIONS,
   ) {
     this._ignoreTypeOnlyChecks = !!ignoreTypeOnlyChecks;
--- a/ng-dev/ts-circular-dependencies/analyzer.ts
+++ b/ng-dev/ts-circular-dependencies/analyzer.ts
@@ -9,6 +9,7 @@
 import {readFileSync} from 'fs';
 import {dirname, join, resolve} from 'path';
 import ts from 'typescript';
+import {CircularDependenciesParserOptions} from './config.js';
 
 import {getFileStatus} from './file_system.js';
 import {getModuleReferences} from './parser.js';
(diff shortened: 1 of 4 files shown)


alxhub (approved):

ng-dev/ts-circular-dependencies/analyzer.ts
```diff
@@ -32,13 +33,18 @@ const DEFAULT_EXTENSIONS = ['ts', 'js', 'd.ts'];
...
   private _sourceFileCache = new Map<string, ts.SourceFile>();
 
+  private _ignoreTypeOnlyChecks: boolean;
+
   unresolvedModules = new Set<string>();
   unresolvedFiles = new Map<string, string[]>();
 
   constructor(
     public resolveModuleFn?: ModuleResolver,
+    {ignoreTypeOnlyChecks}: CircularDependenciesParserOptions = {},
```
devversion, line 43: Nit: Can we just pass through the flag explicitly? The analyzer should just take a boolean and we do the config handling in the CLI index, like with other options.

devversion (approved):

ng-dev/ts-circular-dependencies/analyzer.ts
josephperrott: I can do that, I switched them into two interfaces so that the interface loaded here is just the things the parser needs, I could generalize it to the analyzer potentially as well.


Thoughts?

devversion: I would prefer just passing it through manually as a parameter, but this is just a nit anyway, and I'm good  as is.

ng-dev/ts-circular-dependencies/analyzer.ts
```diff
@@ -32,13 +33,18 @@ const DEFAULT_EXTENSIONS = ['ts', 'js', 'd.ts'];
...
   unresolvedModules = new Set<string>();
   unresolvedFiles = new Map<string, string[]>();
 
   constructor(
     public resolveModuleFn?: ModuleResolver,
+    ignoreTypeOnlyChecks: boolean = false,
     public extensions: string[] = DEFAULT_EXTENSIONS,
-  ) {}
+  ) {
+    this._ignoreTypeOnlyChecks = !!ignoreTypeOnlyChecks;
```
devversion, line 46: Now you can also spare the extra assignment and member declaration by just marking it e.g. as private in the constructor (that was mostly why I preferred the simple parameter)

devversion (approved):

PR #773, opened 2022-08-17
fix(ng-dev): use token to determine if the AuthenticatedGitClient has been configured

josephperrott:
Previously we allowed for the AuthenticatedGitClient to be configured multiple times before
it `get` was called, however this is incorrect as only one configuration is to be allowed.


devversion (approved):

josephperrott:
This PR was merged into the repository by commit 9323859dd14fab851865ca6725d60f207c3545a7.

PR #774, opened 2022-08-17
Properly handle setting github authentication with respect to ng-dev auth service

josephperrott:
See individual commits.

### DIFF
--- a/ng-dev/utils/git/github-yargs.ts
+++ b/ng-dev/utils/git/github-yargs.ts
@@ -24,29 +24,30 @@ export function addGithubTokenOption<T>(argv: Argv<T>) {
         default: '',
         defaultDescription: '<LOCAL_TOKEN>',
         description: 'Github token. If not set, token is retrieved from the environment variables.',
-        coerce: (token: string | null) => {
-          if (token === null) {
-            return '';
-          }
-          const githubToken = token || findGithubTokenInEnvironment();
-          if (!githubToken) {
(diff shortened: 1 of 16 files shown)


ng-dev/utils/git/github-yargs.ts
```diff
@@ -24,29 +24,30 @@ export function addGithubTokenOption<T>(argv: Argv<T>) {
...
 /**
- * Finds a non-explicitly provided Github token in the local environment.
- * The function looks for `GITHUB_TOKEN` or `TOKEN` in the environment variables.
+ * If the github token is able to be determined, either by being provided as a parameter or being
+ * present in the environment, it is used to set the configuration for the AuthenticatedGitClient.
+ * Otherwise, an error is thrown.
+ *
+ * By returning void, we enable the usage of this function as a coerce function for yargs which is
+ * present for users but does not become present in the type structure of the args values at
+ * runtime.
```
devversion, line 41: Optional nit: The last block reads a little confusingly to me (but that might be just me). Also wondering if it's actually needed and can be just removed with an explicit `: void` return type.

devversion (approved):

ng-dev/utils/git/github-yargs.ts
josephperrott: Updated the comment and return signature.  Let me know if this is clearer.  I think it's valuable for us to have so we remember why it is intentionally just a side effectful function.

PR #776, opened 2022-08-17
feat(ng-dev): automatically remove credentials generated by ng-dev auth service after use

josephperrott:
After use, the credentials generated by the ng-dev auth service are automatically requested
to be removed from any git credential helper which is currently active.


devversion (approved):
Nice!! 

Would love to see the same for our `--github-token` flag, but I assume we need to be careful not wiping actual user tokens. (We could annotate our username with something special like `ng-dev-x-github-token`)

PR #778, opened 2022-08-18
feat(github-actions): automatically add `comp: docs` label for pulls that contain docs changes

josephperrott:
For pull requests which contain commits with the `docs` type, automatically add the `comp: docs`
label to the pull request.

### DIFF
--- a/github-actions/commit-message-based-labels/lib/main.ts
+++ b/github-actions/commit-message-based-labels/lib/main.ts
@@ -20,8 +20,12 @@ class CommitMessageBasedLabelManager {
   static run = async () => {
     const token = await getAuthTokenFor(ANGULAR_ROBOT);
     const git = new Octokit({auth: token});
-    const inst = new this(git);
-    await inst.run();
+    try {
+      const inst = new this(git);
+      await inst.run();
+    } finally {
+      await revokeActiveInstallationToken(git);
+    }
   };
 
   /** Labels currently applied to the PR. */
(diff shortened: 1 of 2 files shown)


github-actions/commit-message-based-labels/lib/main.ts
```diff
@@ -11,66 +12,107 @@ const supportedLabels = [
...
+        }
+        if (!hasCommit && hasLabel) {
+          await this.removeLabel(label);
+        }
+      }
+
+      // Add 'comp: docs' label for changes which contain a docs commit.
+      if (!this.labels.has(compDocsLabel)) {
+        for (const commit of this.commits) {
+          if (commit.type === COMMIT_TYPES['docs'].name) {
```
devversion, line 59: Any reason for the docs being accessed through an index signature? It would be more type-safe using plain property access expressions

github-actions/commit-message-based-labels/lib/main.ts
```diff
@@ -11,66 +12,107 @@ const supportedLabels = [
...
 
-  try {
+class CommitMessageBasedLabelManager {
+  /** Run the commit message based labelling process. */
+  static run = async () => {
     const token = await getAuthTokenFor(ANGULAR_ROBOT);
-    installationClient = new Octokit({auth: token});
+    const git = new Octokit({auth: token});
+    const inst = new this(git);
+    await inst.run();
```
devversion, line 24: If something fails in here, we don't revoke the token. That was the whole point of the "new pattern" of the `main` function that introduced in all actions. Any reason to diverge here?

devversion (approved):
LGTM overall, two minor comments

github-actions/commit-message-based-labels/lib/main.ts
josephperrott: Because the type of `COMMIT_TYPES` is `{[key: string}: CommitType}` so it can't be regular property access from typescripts perspective.

github-actions/commit-message-based-labels/lib/main.ts
josephperrott: Ahh good, call, updated.

github-actions/commit-message-based-labels/lib/main.ts
devversion: It would be possible to make it a constant though. Especially since we don't support custom commit types in `ng-dev`. I'm good with this as is, can be a future clean-up, but it's worth noting that currently this is "unsafe"

devversion: In the worst case we change the key from `docs` and we wouldn't even get a type error, but a runtime error because we read `.name` from `undefined`.

PR #782, opened 2022-08-19
ci: allow any command for Renovate post-upgrade tasks

gkalpak:
The [allowedPostUpgradeCommands][1] pattern list is used to determine whether a [post-upgrade task][2] command is allowed to be run by Renovate. Having to explicitly allow each command creates unnecessary friction.

This commit allows any command to be run in a post-upgrade task, which saves us the trouble of having to update the allowed commands pattern list every time we want to add a new post-upgrade task.
Since adding/modifying Renovate post-upgrade tasks is subject to the regular PR review workflow, unwanted commands will be caught there.

##
_INFO: [Internal discussion][3]_

[1]: https://docs.renovatebot.com/self-hosted-configuration/#allowedpostupgradecommands
[2]: https://docs.renovatebot.com/configuration-options/#postupgradetasks
[3]: https://angular-team.slack.com/archives/C029MM6RS4B/p1659605197808499

### DIFF
--- a/.github/ng-renovate/runner-config.js
+++ b/.github/ng-renovate/runner-config.js
@@ -4,10 +4,7 @@ module.exports = {
   platform: 'github',
   forkMode: true,
   onboarding: false,
-  allowedPostUpgradeCommands: [
-    // See https://github.com/angular/angular/pull/47040.
-    '^yarn --cwd=aio/tools/examples/shared run sync-deps$',
-  ],
+  allowedPostUpgradeCommands: ['.'],
   repositories: [
     'angular/angular',
     'angular/dev-infra',


.github/ng-renovate/runner-config.js
```diff
@@ -4,10 +4,7 @@ module.exports = {
   platform: 'github',
   forkMode: true,
   onboarding: false,
-  allowedPostUpgradeCommands: [
-    // See https://github.com/angular/angular/pull/47040.
-    '^yarn --cwd=aio/tools/examples/shared run sync-deps$',
-  ],
+  allowedPostUpgradeCommands: ['.'],
```
devversion, line 7: Do we want to use `.*` or something? I don't know if Renovate expects a full match or if a partial is fine.

gkalpak: No, I don't think Renovate expects a full match. For example, we were using ['^yarn run '][1] before and it worked fine afaict.

[1]: https://github.com/angular/dev-infra/blob/d2f9d2f493ec99f89a1db2777e2b27c1f5b278e9/.github/ng-renovate/runner-config.js#L7

devversion (approved):

.github/ng-renovate/runner-config.js
devversion: makes sense!

PR #789, opened 2022-08-22
fix(apps): ensure credential service login only works with `@google.com` mail

devversion:
Currently we just test if @google.com is part of the email. With custom
domains it looks like this could be abused to something like:

`paul@google.com.my-domain.net`


josephperrott (approved):
LGTM


Interesting, my understanding is that it will be normalized by the firebase auth service, so it would end up providing us with not `google.com.something.something`, but this definitely won't hurt and isn't based on that assumption so 👍

PR #795, opened 2022-08-24
fix(bazel/browsers): exclude a log file that chromium writes to on linux causing cache misses

kormide:
@devversion I think this is the source of cache misses I'm seeing in https://github.com/angular/angular/pull/47161. In unsandboxed execution (which is the case for targets run under RBE, right?) chromium will write to this log file each run in under the `external` dir causing remote cache misses on successive builds.

### DIFF
--- a/bazel/browsers/chromium/chromium.bzl
+++ b/bazel/browsers/chromium/chromium.bzl
@@ -21,6 +21,11 @@ def define_chromium_repositories():
         named_files = {
             "CHROMIUM": "chrome-linux/chrome",
         },
+        exclude_patterns = [
+            # Exclude a log file that chromium writes to each run, causing remote cache
+            # misses in downstream targets.
+            "chrome-linux/chrome_debug.log",
+        ],
     )
 
     browser_archive(


josephperrott (approved):
LGTM

kormide:
As a potential workaround, in my protractor config file I tried passing the flags `--disable-logging` and `--disable-and-delete-previous-log` however they had no effect. Perhaps those flags don't work on the current version of chromium supported.

Passing `--log-level=3` helped to keep this file empty, but it still gets created when it doesn't exist after a fresh bazel run..

bazel/browsers/chromium/chromium.bzl
```diff
@@ -21,6 +21,11 @@ def define_chromium_repositories():
         named_files = {
             "CHROMIUM": "chrome-linux/chrome",
         },
+        exclude_patterns = [
+            # Exclude a log file that chromium writes to each run, causing remote cache
+            # misses in downstream targets.
+            "chrome-linux/chrome_debug.log",
```
devversion, line 27: Do we need a similar thing for e.g. Mac or Windows?

kormide: I'm not sure. I'll test both and report back.

kormide: @devversion It doesn't appear to cause cache misses on Windows.

On OSX, it's a bit more complicated. Using bazel-provided chromium in unsandboxed execution fails early because it can't build runfiles due to all of the files with spaces in them. If I exclude all of those files, then the runfiles can be built but chromium fails, probably because it needs some of those files.

It's not clear to me why it works with sandboxed execution on OSX but fails when run outside of the sandbox. My understanding was that using `--nobuild_runfile_links` fixed the runfiles with spaces issue but that doesn't seem to work for unsandboxed execution.

PR #796, opened 2022-08-25
fix(ng-dev): order results of forks when searching for user owned forks

josephperrott:
It appears that Github has made a change, or has introduced a bug in which unsorted results of forks for a repository with OWNER affiliation are returned as empty.  Once ordering is provided the results show again as expected.  While this should not be necessary, it is an innocuous change for us that is fine to remain indefinitely.


gkalpak (approved):

devversion (approved):
Super weird bug, can confirm the fix works.

@josephperrott can you capture the GitHub ticket outcome here?

josephperrott:
@devversion will update once I have a response.

josephperrott:
It appears that whatever change was made that caused the issue was rolled back and the requests now return as expected.

PR #798, opened 2022-08-25
feat(ng-dev): add logging and messaging to ng-dev auth login

josephperrott:
Add additional logging and information about the log in process for ng-dev auth login.


ng-dev/auth/shared/ng-dev-token.ts
```diff
@@ -107,16 +107,12 @@ async function saveTokenToFileSystem(data: NgDevUserWithToken) {
 
 /** Retrieve the token from the file system. */
 async function retrieveTokenFromFileSystem(): Promise<NgDevUserWithToken | null> {
-  try {
-    if (!(await stat(tokenPath))) {
-      return null;
-    }
-  } catch {
-    return null;
+  if (await hasEvidenceOfPreviousLogin()) {
```
devversion, line 110: Super nit: I think it would be better to call the function `hasTokenStoreFile`. Only considering the name it doesn't seem to guarantee that the `tokenPath` exists.

ng-dev/auth/shared/firebase.ts
```diff
@@ -17,6 +19,24 @@ export async function loginToFirebase() {
...
+    // Only present intial information about usage of login when it appears that the user has
+    // not logged into the service in the past.
+    if (!(await hasEvidenceOfPreviousLogin())) {
+      Log.warn(Array(80).fill('#').join(''));
+      Log.warn('The ng-dev auth service uses Google OAuth credentials to log in and create a');
+      Log.warn('short lived credential used for authenticating with the ng-dev service.');
+      Log.warn('');
+      Log.warn('In addition to logging in using Google credentials, upon first login you will be');
+      Log.warn('prompted to associate your Github account to your login, allowing the service to');
+      Log.warn('perform requests on your your behalf.');
```
devversion, line 31: Nit: You may want to use `Log.info` for messages that should not be yellow. Just in case you want to make some phrases more prominent.

devversion (approved):
LGTM, nice!

ng-dev/auth/shared/firebase.ts
josephperrott: I wanted this to stand out since its the explanation of what is going to happen on login. Do you think it shouldn't be?

devversion: I just wanted to see if you just wanted to make a subet of these lines yellow. Now all are yellow

josephperrott: I think having the whole block yellow is good in this specific case.

devversion: sgtm!

PR #801, opened 2022-08-25
Update typings to allow invocation of ng-dev with ts-node again

josephperrott:
See individual commits.


tools/ts_proto/index.bzl
```diff
@@ -8,26 +8,24 @@ def generate_ts_proto_module(name, protofile, visibility = None):
...
             "$(execpath %s)" % protofile,
+            "js",
         ],
         stdout = "generated_" + js_file,
     )
 
     npm_package_bin(
         name = "generate_ts_" + name,
-        tool = "@ts_proto_npm//protobufjs-cli/bin:pbts",
+        tool = "//tools/ts_proto:bundler",
```
devversion, line 23: What's the context for all of these changes? 

1. ProtobufJS CLI is intentionally kept in a separate install as it's only needed during generation and not for the normal development flow.
2. Doesn't protobuf CLI work correctly anymore? Is the result of an update?

tools/ts_proto/bundle-proto.js
```diff
@@ -0,0 +1,46 @@
...
+          process.stderr.write(err.name);
+          process.stderr.write(err.message);
+          process.stderr.write(err.stack);
+          return resolve(1);
+          
+        }
+
+        output = output.replace(
+          'import * as $protobuf from "protobufjs/minimal";',
+          'import $protobuf from "protobufjs";',
```
devversion, line 19: Doesn't ESBuild handle this properly? Has this come up recently — or did it never work and we just didn't notice?

tools/ts_proto/index.bzl
josephperrott: By using this bundler script, we can leverage the protobufjs-cli bundler, but modify the import statement used so that this works with ESM.  Currently protobufjs-cli does not have an option for generating ESM.

tools/ts_proto/bundle-proto.js
devversion: It seems brittle to do it the current way with a custom bundler. There are alternative Protobuf generation tools, or we could just have a temporary Bazel rule in between like `expand_template` to substitute things.

Just want to avoid having to clean this up in the future – if we can just move to something better. Also we should link to an issue so that ProtobufJS-cli can support ESM properly.

tools/ts_proto/index.bzl
devversion: Makes sense. I replied in another comment on this.

tools/ts_proto/bundle-proto.js
josephperrott: I looked into the code more and found that we can do it another way.  PTAL

devversion (approved):
awesome! Nice work 👍

PR #820, opened 2022-09-14
feat(create-pr-for-changes): allow specifying a base branch

gkalpak:
Previously, the `create-pr-for-changes` GitHub action would use the branch that the workflow was triggered on as the base branch. This prevented the action from being used on other branches as part of a [scheduled workflow][1], since scheduled workflows are only triggered on the default branch of a repository.

This commit adds the ability to specify a different base branch. If not specified, this defaults to the branch the workflow was triggered on, as before.

[1]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

### DIFF
--- a/github-actions/create-pr-for-changes/lib/main.ts
+++ b/github-actions/create-pr-for-changes/lib/main.ts
@@ -48,7 +48,7 @@ async function main(): Promise<void> {
 
     // Unless configured otherwise, clean up obsolete branches.
     const branchPrefix = `${core.getInput('branch-prefix', {required: true})}-`;
-    const cleanUpBranches = core.getInput('clean-up-branches', {required: false}) === 'true';
+    const cleanUpBranches = core.getInput('clean-up-branches', {required: true}) === 'true';
     const forkRepo = await git.getForkOfAuthenticatedUser();
 
     if (cleanUpBranches) {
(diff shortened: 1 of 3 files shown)


github-actions/create-pr-for-changes/lib/main.ts
```diff
@@ -48,7 +48,7 @@ async function main(): Promise<void> {
 
     // Unless configured otherwise, clean up obsolete branches.
     const branchPrefix = `${core.getInput('branch-prefix', {required: true})}-`;
-    const cleanUpBranches = core.getInput('clean-up-branches', {required: false}) === 'true';
+    const cleanUpBranches = core.getInput('clean-up-branches', {required: true}) === 'true';
```
devversion, line 51: do we need to reflect that in the YAML or are these separate things?

devversion (approved):

josephperrott (approved):
LGTM

github-actions/create-pr-for-changes/lib/main.ts
gkalpak: Reflect what?

devversion: The fact that you now specify `required: true` in TS— but in YAML, `clean-up-branches` has `required: false`.

devversion: I think I see that these are two separate things though:

* In YAML: Required means that the user needs to specify this.
* Via `core.getInput`, required means that the input needs to be defined (e.g. maybe through a default)

gkalpak: Yeah, exactly. The ones that are `required: false` in the YAML file have default values, so they are expected to be present in the TS file (thus `{required: true}`).

devversion: Makes sense, thx for clarifying

PR #825, opened 2022-09-15
fix(github-actions): use GitHub PAT instead of GitHub App key in `create-pr-for-changes`

gkalpak:
Using the GitHub App key for Angular Robot does not retrieving a fork to create PRs from. Switching to using a GitHub Personal Access Token in order to be able to create PRs from a fork and not pollute the upstream branches.


github-actions/create-pr-for-changes/lib/main.ts
```diff
@@ -6,7 +6,6 @@ import {GithubConfig, setConfig} from '../../../ng-dev/utils/config.js';
 import {AuthenticatedGitClient} from '../../../ng-dev/utils/git/authenticated-git-client.js';
 import {GithubRepo} from '../../../ng-dev/utils/git/github.js';
 import {getRepositoryGitUrl} from '../../../ng-dev/utils/git/github-urls.js';
-import {ANGULAR_ROBOT, getAuthTokenFor, revokeActiveInstallationToken} from '../../utils.js';
```
devversion, line 9: Nit: Can you update the BUILD file to remove the dependency?

devversion (approved):

PR #826, opened 2022-09-16
feat(bazel/remote-execution): expose exec_properties constant to enable networking on remote exec platform

kormide:
@devversion


devversion (approved):
This is exactly what I had in mind. Thanks!

PR #828, opened 2022-09-17
fix(ng-dev/release): avoid accidentally incorporating unexpected changes in release build

devversion:
There is an interesting scenario where the caretaker may make unexpected non-changelog changes as part of the short prompt where we allow manual polishing to the changelog.

If the caretaker modified other files there, they will NOT be part of the release cutting PR (and also shouldn't be) - but currently the changes are not discarded/or detected and WILL influence the release output building. This could cause subtle differences and confusion because the release output is different than what we committed (because the release cut commits will not incorporate the other non-changelog changes).

The fix is to properly detect other modifications and abort properly. Other changes should be made beforehand, go through review process and should not be part of the release cut commit.

### DIFF
--- a/ng-dev/release/publish/actions.ts
+++ b/ng-dev/release/publish/actions.ts
@@ -195,11 +195,20 @@ export abstract class ReleaseAction {
 
     // Commit message for the release point.
     const commitMessage = getCommitMessageForRelease(newVersion);
+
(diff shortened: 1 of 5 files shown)


clydin (approved):

josephperrott (approved):
LGTM

PR #831, opened 2022-09-19
feat(ng-dev): support `target: feature` in merge tooling

josephperrott:
Support the usage of a target label for feature branches `target: feature`.


devversion (approved):
Nice nice!

PR #832, opened 2022-09-20
Various CircleCI orb packaging improvements and rebase command fixes

devversion:


devversion:
@josephperrott just because the commit is a little obfuscated, here is a little overview of changes:

1. Split up the rebase script into multiple TS files - to keep it more readable
2. Fixes the script on Windows. CLI repo requires this - the use of `spawnSync` is needed and proper line ending splitting.
3. Cuts the Orb at `1.0.1` (I experimented a bit - this is the version I ended up with)
4. Reworked the packaging to work with Bash command line length limits. Using stdin redirect here-docs syntax works around it. Windows Git Bash is still odd- but I don't expect a lot of Windows commands being shared.

josephperrott (approved):
LGTM

PR #837, opened 2022-09-22
refactor(ng-dev): centralize the configuration/definition of labels that are controlled by tooling

josephperrott:
Centralize the way that we configure tooling managed labels in preparation for label restructuring and allowing for general expansion of managed labels.

### DIFF
--- a/ng-dev/pr/config/index.ts
+++ b/ng-dev/pr/config/index.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {Commit} from '../../commit-message/parse.js';
 import {ConfigValidationError, GithubConfig, NgDevConfig} from '../../utils/config.js';
 
 /**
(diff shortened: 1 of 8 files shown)


ng-dev/pr/config/index.ts
```diff
@@ -91,8 +92,29 @@ export function assertValidPullRequestConfig<T extends NgDevConfig>(
...
+  /* The label string. */
+  label: string;
+  /** A matching function, if the label is automatically applied by our github action, otherwise false. */
+  commitCheck: ((c: Commit) => boolean) | false;
+}
 
-/** Label for pull requests containing a deprecation. */
-export const deprecationLabel = 'flag: deprecation';
+/** Set of labels which are known to tooling, and in some cases are managed by tooling. */
+export const ToolingPullRequestLabels = {
```
devversion, line 103: Nice! I love how flexible this is.

Do you think we can move this somewhere else outside of `/config/index` because that is meant to be always for the ng-dev configuration part, like e.g. `release/config`. How about `config/constants.ts`?

devversion (approved):

ng-dev/pr/config/index.ts
josephperrott: What do you think of `ng-dev/pr/common/labels.ts`?

devversion (approved):

PR #838, opened 2022-09-23
fix(circleci-orb): support empty input parameters when e.g. running in cronjob

devversion:
CircleCI cronjobs do not have the base revision or revision pipeline variables, so they may be empty. To avoid using empty variables which would then cause errors in our shared command- we support the empty string fallback. The command will be a noop anyway since we check for these variables in the actual script.

See: https://app.circleci.com/pipelines/github/angular/components/44462.


josephperrott (approved):
LGTM

PR #840, opened 2022-09-23
feat(ng-dev): include all labels from the team's label schema

josephperrott:
Include all labels from the agreed label schema for management via tooling.

PR #841, opened 2022-09-23
feat(github-actions): create action to syncronize labels in angular repositories

josephperrott:
This PR builds upon #840, leveraging the added labels to synchronize them within the action.

### DIFF
--- a/github-actions/labels-sync/action.yml
+++ b/github-actions/labels-sync/action.yml
@@ -1,0 +1,16 @@
+name: 'Labels Sync'
+description: 'Synchronize labels in repositories throughout the organization'
+author: 'Angular'
+inputs:
+  angular-robot-key:
+    description: 'The private key for the Angular Robot Github app.'
+    required: true
+  repos:
+    description: |
+      The repositories in which to keep labels in sync. The organization name is derived from
+      the context in where the action runs.
+    required: true
+
+runs:
+  using: 'node16'
+  main: 'main.js'
--- a/github-actions/labels-sync/src/main.ts
+++ b/github-actions/labels-sync/src/main.ts
@@ -1,0 +1,86 @@
+import * as core from '@actions/core';
+import {context} from '@actions/github';
+import {Octokit, RestEndpointMethodTypes} from '@octokit/rest';
+import {allLabels, Label} from '../../../ng-dev/pr/common/labels.js';
+import {getAuthTokenFor, ANGULAR_ROBOT, revokeActiveInstallationToken} from '../../utils.js';
+
+/** The type for a Github label returned from the Github API.  */
+type GithubLabel =
+  RestEndpointMethodTypes['issues']['listLabelsForRepo']['response']['data'][number];
+
+/** Synchronize the provided managed labels with the given repository. */
+async function syncLabelsInRepo(github: Octokit, repoName: string, managedLabels: Label[]) {
+  core.startGroup(`Syncing ${repoName}`);
+  /** The current repository name and owner for usage in the Github API. */
+  const repo = {repo: repoName, owner: context.repo.owner};
+
+  core.debug(`Requesting labels`);
+  /** The list of current labels from Github for the repository. */
+  const repoLabels = await github.paginate(github.issues.listLabelsForRepo, context.repo);
+  core.debug(`Retrieved ${repoLabels.length} from Github`);
+
+  // For each label in the list of managed labels, ensure that it exists and is in sync.
+  // NOTE: Not all labels in repositories are managed. Labels which are not included or managed in
+  // our tooling definitions and configurations are ignored entirely by tooling.
+  for (const {description, label} of managedLabels) {
+    await core.group(`Syncing label: ${label}`, async () => {
+      /** The label from Github if a match is found. */
+      const matchedLabel = repoLabels.find(({name}: GithubLabel) => name === label);
+
+      // When no matched label is found, Github doesn't currently have the label we intend to sync,
+      // we create the label via the API directly.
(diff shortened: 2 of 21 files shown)


github-actions/labels-sync/action.yml
```diff
@@ -0,0 +1,14 @@
+name: 'Labels Sync'
+description: 'Synchronize labels in repositories throughout the organization'
+author: 'Angular'
+inputs:
+  angular-robot-key:
+    description: 'The private key for the Angular Robot Github app.'
+    required: true
+  repos:
+    description: 'The repositories in which to keep labels in sync.'
```
devversion, line 9:
```suggestion
    description: |
      The repositories in which to keep labels in sync. Organization name is derived from
      the context in where the action runs.
```

github-actions/labels-sync/src/main.ts
```diff
@@ -0,0 +1,84 @@
...
+  /** The current repository name and owner for usage in the Github API. */
+  const repo = {repo: repoName, owner: context.repo.owner};
+
+  core.debug(`Requesting labels`);
+  /** The list of current labels from Github for the repository. */
+  const repoLabels = await github.paginate(github.issues.listLabelsForRepo, context.repo);
+  core.debug(`Retrieved ${repoLabels.length} from Github`);
+
+  // For each label in the list of managed labels, ensure that it exists and is in sync.
+  for (const {description, label} of managedLabels) {
```
devversion, line 23: What should happen with additional per-repo specific labels? Maybe we should have a comment or description somewhere stating that currently these are not cleaned up/removed?

ng-dev/pr/common/labels/action.ts
```diff
@@ -0,0 +1,23 @@
+import {createTypedObject, Label} from './base.js';
+
+interface ActionLabel extends Label {}
+
+export const actionLabels = createTypedObject<ActionLabel>()({
```
devversion, line 5: Why not just assert the labels like this?

```suggestion
export const actionLabels: Record<string, ActionLabel> = {
```

ng-dev/pr/common/labels/action.ts
```diff
@@ -0,0 +1,23 @@
...
+  ACTION_MERGE: {
+    description: 'The PR is ready for merge by the caretaker',
```
devversion, line 7:
```suggestion
    description: 'The PR author is ready for merge by the caretaker',
```

I wonder if this should be done that way. It's always unclear if people can apply it before when they are ready (as the author) etc. You know this 😄

ng-dev/pr/common/labels/action.ts
```diff
@@ -0,0 +1,23 @@
...
+    label: 'action: merge',
+  },
+  ACTION_CLEANUP: {
+    description:
+      'The PR is in need of cleanup, either due to needing a rebase or in response to comments from a review.',
```
devversion, line 12: Super nit: Note that some descriptions end with a dot, while others not.

ng-dev/pr/common/labels/feature.ts
```diff
@@ -0,0 +1,24 @@
...
+    description: 'Label used to distinguish requests in voting phase from other issues',
+  },
+  FEATURE_UNDER_CONSIDERATION: {
+    label: 'feature: under consideration',
+    description: 'Label used to distinguish features which are in our list for consideration',
+  },
+  FEATURE_INSUFFICIENT_VOTES: {
+    label: 'feature: insufficient votes',
+    description:
+      'Label to add when the `close-when-no-sufficient-votes` is set to false and there are no sufficient number of votes or comments from unique authors',
```
devversion, line 22: Is `close-when-no-sufficient-votes` relevant to our contributors? I think the description should not mention this.

ng-dev/pr/common/labels/managed.ts
```diff
@@ -0,0 +1,30 @@
...
+interface ManagedLabel extends Label {
+  /** A matching function, if the label is automatically applied by our github action, otherwise false. */
+  commitCheck: ((c: Commit) => boolean) | false;
+}
+
+export const managedLabels = createTypedObject<ManagedLabel>()({
+  DETECTED_BREAKING_CHANGE: {
+    description: 'PR contains a commit with a breaking change',
+    label: 'flag: breaking change',
+    commitCheck: (c: Commit) => c.breakingChanges.length !== 0,
```
devversion, line 13: Nice 🎉

devversion (approved):

ng-dev/pr/common/labels/action.ts
josephperrott: Having this style of typed object allows us to have specific keys while still enforcing that all of the values are of a specific type.

devversion: But in either case you need to explicitly add extra code here. So using an explicit type assertion (like in my suggestion) enforces that the object conforms to the `Record` type too?

ng-dev/pr/common/labels/action.ts
josephperrott: This was the text used in the description in the managed labels proposal.

Since we use the merge label currently to define the merge queue, I think we should leave it as is for now and if we find we need to update it in the future we can.

github-actions/labels-sync/src/main.ts
josephperrott: Added more comment to expand on it.

ng-dev/pr/common/labels/action.ts
devversion: Ahh, I didn't catch that you want the keys to explicit in the type. Makes sense

PR #847, opened 2022-10-07
fix(bazel/browsers): exclude chromium debug logs on remaining platforms

kormide:
@devversion

### DIFF
--- a/bazel/browsers/chromium/chromium.bzl
+++ b/bazel/browsers/chromium/chromium.bzl
@@ -22,8 +22,7 @@ def define_chromium_repositories():
             "CHROMIUM": "chrome-linux/chrome",
         },
         exclude_patterns = [
-            # Exclude a log file that chromium writes to each run, causing remote cache
-            # misses in downstream targets.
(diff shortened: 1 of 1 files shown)


bazel/browsers/chromium/chromium.bzl
```diff
@@ -41,6 +40,10 @@ def define_chromium_repositories():
         named_files = {
             "CHROMIUM": "chrome-mac/Chromium.app/Contents/MacOS/Chromium",
         },
+        exclude_patterns = [
+            # Exclude a log file that chromium writes to each run, causing remote cache misses.
+            "chrome-mac/Chromium.app/Contents/Frameworks/Chromium Framework.framework/Versions/*/chrome_debug.log",
```
kormide, line 45: Log gets output under `Versions/Current` and `Versions/99.0.4759.0`.

kormide: I don't actually have an `x64` mac to test this on, but I'm assuming the path is the same as the `arm64`.

devversion (approved):
Thank you!! this is awesome

PR #848, opened 2022-10-10
feat(github-actions): introduce action to identify PRs affecting Google

devversion:
This is the external part to the internal service. The internal service will complete the commit statuses for PRs needing internal presubmits.

This action exists as we do not auto-generate presubmits for PRs.


github-actions/google-internal-tests/lib/main.ts
```diff
@@ -0,0 +1,79 @@
...
+    return;
+  }
+  if (context.eventName !== 'pull_request_target' || context.payload.action !== 'synchronize') {
+    throw new Error(
+      'Expected Google Internal Tests action to be triggered for ' +
+        '`pull_request_target.synchronize` events.',
+    );
+  }
+
+  const githubToken = core.getInput('github-token', {required: true});
```
josephperrott, line 18: Can we use the angular robot instead of the Github Actions token?

github-actions/google-internal-tests/lib/main.ts
```diff
@@ -0,0 +1,79 @@
...
+      } as const)
+    : ({
+        state: 'success',
+        description: 'Does not affect Google.',
+      } as const);
+
+  await github.repos.createCommitStatus({
+    ...context.repo,
+    sha: prHeadSHA,
+    state: status.state,
```
josephperrott, line 58: If we use `target_url` as the key for the status object, then you could just use `...status` here.

github-actions/google-internal-tests/lib/main.ts
devversion: Resolved offline. Using the github action token we can limit permissions better

github-actions/google-internal-tests/lib/main.ts
```diff
@@ -0,0 +1,79 @@
...
+        // internal presubmit status service then.
+        description: `Waiting for Google Internal Tests. ${
+          runTestGuideURL ? `@Googlers: See Details for instructions -->` : ''
+        }`.trim(),
+        url: runTestGuideURL,
+      } as const)
+    : ({
+        state: 'success',
+        description: 'Does not affect Google.',
+      } as const);
```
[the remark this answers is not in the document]
devversion: I like it- done.

github-actions/google-internal-tests/lib/main.ts
devversion: good idea. done!

josephperrott (approved):
LGTM

PR #849, opened 2022-10-11
fix(github-actions): allow for other pull request event actions

devversion:
We also want to trigger for `opened` and `reopened` events.


josephperrott (approved):
LGTM

PR #853, opened 2022-10-12
feat(github-action): do not override g3 status if pointing to CL

devversion:


github-actions/google-internal-tests/lib/main.ts
```diff
@@ -15,17 +20,32 @@ async function main() {
...
   const alwaysExternalFiles = constructPatterns(alwaysExternalFilesRaw);
 
   const github = new Octokit({auth: githubToken});
-  const prNum = context.payload.pull_request!.number;
-  const prHeadSHA = context.payload.pull_request!.head!.sha;
+  const existingGoogleStatus = await findExistingTestStatus(github, prHeadSHA);
+
+  // If there is an existing status already pointing to an internal CL, we do not override
+  // the status. This can happen when e.g. a presubmit-tested PR is closed and reopened.
+  if (existingGoogleStatus && existingGoogleStatus.target_url?.startsWith('http://cl/')) {
```
devversion, line 46: This is the thing I'm a little unhappy about. It relies on the URL we populate from internal. At the same time though this whole action already relies on the context name being `google-internal-tests`- so maybe fine 🤷

github-actions/google-internal-tests/action.yml
```diff
@@ -7,6 +7,10 @@ inputs:
   github-token:
     required: true
     description: GitHub Token used for creating the commit statuses.
+  sync-branch:
```
josephperrott, line 10: IMHO I don't think we should have this as a setting, it should just always be `main`.  Particularly because if we allow another branch to be used, then we need to leverage our branch targeting logic from ng-dev.  `main` is the one exception as it must be targeted via the github UI to be included.

github-actions/google-internal-tests/lib/main.ts
josephperrott: I view this action as being part of the internal tool (or rather they together make up the tool), so I think its fine to rely on implementation details of the internal piece.

github-actions/google-internal-tests/lib/main.ts
```diff
@@ -80,6 +100,22 @@ function constructPatterns(rawPatterns: string): minimatch.IMinimatch[] {
   return patterns;
 }
 
+async function findExistingTestStatus(
+  github: Octokit,
+  prHeadSHA: string,
+): Promise<GithubStatus | null> {
+  const existingStatuses = await github.paginate(
+    github.repos.getCombinedStatusForRef,
```
josephperrott, line 108: Can this method even paginate?

josephperrott (commented):
As far as an implementation goes for not overrideing the g3 status, LGTM

Should we instead consider just gating when this action runs to only be `open`, `reopen` and `synchronize`?  This would accomplish the same result as the action would only run on what should be considered "fresh" commits.

github-actions/google-internal-tests/lib/main.ts
devversion: Yes, looks like it: https://octokit.github.io/rest.js/v19#:~:text=Get%20the%20combined%20status%20for%20a%20specific%20reference

github-actions/google-internal-tests/action.yml
devversion: Updated.

josephperrott (approved):
LGTM

Discussed the complexity vs precision tradeoffs offline.

PR #859, opened 2022-10-13
build: enable org wide action to sync labels

josephperrott:
Initially only set up the dev-infra repo confirm the sync works as expected.

### DIFF
--- a/.github/workflows/org-wide-actions.yml
+++ b/.github/workflows/org-wide-actions.yml
@@ -1,0 +1,22 @@
+name: Angular Org Actions
+
+on:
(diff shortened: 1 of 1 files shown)


devversion (approved):
LGTM, I assume we could limit it runs to e.g. every 3 hours given the manual dispatch too

josephperrott:
@devversion  Just to make sure I am thinking of this correctly: `'0 0 * * *'` is 1 time per day at 0:00 right?

devversion:
It is- but the comment doesn't reflect that, and I just look there

PR #861, opened 2022-10-13
refactor(ng-dev/pr): update standard label descriptions to be <= 100 characters

josephperrott:
Github only allows for descriptions to be 100 characters, and as such, standard labels must comply with this requirement.


devversion (approved):
🚀

PR #865, opened 2022-10-13
Refactor validations

josephperrott:
Part of the intent here is to separate our pull request validation from pull request merging so we can perform validations outside of actual merges.

### DIFF
--- a/ng-dev/pr/common/validation/validate-pull-request.ts
+++ b/ng-dev/pr/common/validation/validate-pull-request.ts
@@ -19,6 +19,7 @@ import {passingCiValidation} from './assert-passing-ci.js';
 import {pendingStateValidation} from './assert-pending.js';
 import {signedClaValidation} from './assert-signed-cla.js';
 import {PullRequestValidationConfig} from './validation-config.js';
+import {PullRequestValidationFailure} from './validation-failure.js';
 
 /**
  * Asserts that the given pull request is valid. Certain non-fatal validations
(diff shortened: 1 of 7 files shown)


ng-dev/pr/merge/pull-request.ts
```diff
@@ -46,6 +47,8 @@ export interface PullRequest {
   baseSha: string;
   /** Git revision range that matches the pull request commits. */
   revisionRange: string;
+
+  validationFailures: PullRequestValidationFailure[];
```
devversion, line 51: Add a comment for this field too, and should we actually do this now? Couldn't we continue asserting directly instead of storing the failures here (in the loadAndValidate function)

ng-dev/pr/merge/merge-tool.ts
```diff
@@ -86,6 +87,24 @@ export class MergeTool {
...
     const pullRequest = await loadAndValidatePullRequest(this, prNumber, validationConfig);
 
+    if (pullRequest.validationFailures.length > 0) {
+      Log.error(`Pull request did not pass one or more validation checks. Error:`);
+
+      for (const failure of pullRequest.validationFailures) {
+        Log.error(` -> ${bold(failure.message)}`);
+      }
+      Log.info();
+      if (pullRequest.validationFailures.find((failure) => !failure.canBeForceIgnored)) {
```
devversion, line 97:
```suggestion
      if (pullRequest.validationFailures.some((failure) => !failure.canBeForceIgnored)) {
```

ng-dev/pr/merge/merge-tool.ts
```diff
@@ -86,6 +87,24 @@ export class MergeTool {
...
+        Log.info(yellow(`All discovered validations are non-fatal and can be forcibly ignored.`));
+
+        if (await Prompt.confirm('Do you want to forcibly ignore these validation failures?')) {
+          return;
+        }
+      }
+
+      throw pullRequest.validationFailures[0];
```
devversion, line 105: oh this feels incorrect. could we just throw fatal merge tool error then?

ng-dev/pr/common/validation/validate-pull-request.ts
```diff
@@ -35,22 +36,38 @@ export async function assertValidPullRequest(
   ngDevConfig: NgDevConfig<{pullRequest: PullRequestConfig; github: GithubConfig}>,
```
devversion, line 36: the jsdoc above needs to be updated

devversion (commented):
Nice improvement. For what else do you anticipate this being useful?

ng-dev/pr/merge/merge-tool.ts
josephperrott: Created new FatalPullRequestValidationFailureError class for this usage.  This also aligns better as it is now defined in the same place as the other pull request fatal errors.

devversion (approved):

PR #867, opened 2022-10-14
fix(ng-dev): update caretaker check to support canonical config for g3 sync

devversion:
See individual commits


crisbeto (approved):
LGTM

PR #878, opened 2022-10-25
build: keep `branchNameStrict` set to `true`

devversion:
Renovate had a breaking change in v33 and reverted in v34. Since we were already on v33- we want to keep the strict branch names given that we already had new PRs sent out.

See https://github.com/renovatebot/renovate/releases/tag/34.0.0 for details.


devversion:
Related to #876

alan-agius4 (approved):

PR #879, opened 2022-10-25
refactor(github-actions): always set google internal tests status

devversion:
Currently the action is a noop if a PR targets a branch other than the sync branch. This works- but in practice the old legacy Github Robot status does not differentiate between branches and enforces that `google-internal-tests` is also present even for e.g. PRs only targeting the patch branch.


josephperrott (approved):
LGTM

PR #880, opened 2022-10-25
build: do not use test-results grather command in platform CI jobs

devversion:
Currently we end up building all of ng-dev in the Windows/MacOS job-even though we only are interested in some fine-grained selected targets. Also the ng-dev gather build fails on Windows likely due to no full windows Bazel setup. 

It's not worth investigating- our windows-specific tests run as expected and pass.


josephperrott (approved):
LGTM

PR #881, opened 2022-10-25
feat(ng-dev/caretaker): add status urls to caretaker check

josephperrott:
Add the status urls to the caretaker check for failing services to allow easier detail checks of the status for caretakers.


devversion (approved):

pkozlowski-opensource (approved):
LGTM, thnx @josephperrott !

PR #885, opened 2022-10-26
feat(github-actions): create a local branch manager github action

josephperrott:
Create the local branch manager action which can recieve requests for a branch manager check for any of the managed repositories in the angular organization.


github-actions/utils.ts
```diff
@@ -30,13 +30,16 @@ async function getJwtAuthedAppClient([appId, inputKey]: GithubAppMetadata) {
...
+  app: GithubAppMetadata,
+  useOrgInstallation: boolean = false,
+): Promise<string> {
   const github = await getJwtAuthedAppClient(app);
 
   const {id: installationId} = (
-    await github.apps.getRepoInstallation({
-      ...context.repo,
-    })
+    await (useOrgInstallation
```
devversion, line 40: Wouldn't GitHub even with an org installation automatically be able to provide a token for the current context repo?

.github/local-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,160 @@
...
+    context: 'Branch Manager',
+  });
+}
+
+/** The token for the angular robot to perform actions. */
+const token = await getAuthTokenFor(ANGULAR_ROBOT, true);
+/** The repository name for the pull request. */
+const repo = core.getInput('repo', {required: true, trimWhitespace: true});
+/** The owner of the repository for the pull request. */
+const owner = core.getInput('owner', {required: true, trimWhitespace: true});
```
devversion, line 149: Should we just retrieve owner/name of the repo through context? Like it's also done for the `getAuthTokenFor`. Right now we have this setting but it's not used consistently- and not sure if needed

.github/local-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,160 @@
...
+        state: 'failure',
+      };
+    }
+  })();
+
+  await git.github.repos.createCommitStatus({
+    ...repo,
+    ...statusInfo,
+    sha: pullRequest.headSha,
+    context: 'Branch Manager',
```
devversion, line 140: Serious question: Do we want to keep naming it branch manager? or should we make it more clear to users? e.g. `mergeability`?

.github/local-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,160 @@
...
+      };
+    } catch (e) {
+      // As the merge strategy class will express the failures during checks, any thrown error is a
+      // failure for our merge check.
+      let description: string;
+      if (e instanceof MergeConflictsFatalError) {
+        core.info('Merge conflict found');
+        description = `Unable to merge into ${e.failedBranches.join(
+          ', ',
+        )} please update changes or PR target`;
```
devversion, line 123: Note: GitHub doesn't show large descriptions and just cuts them off- without even showing the full description on hover (at least on my end)

.github/local-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,160 @@
...
+      } else {
+        core.info('Unknown error found when checking merge');
```
devversion, line 125: We should print the error here.

.github/local-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,160 @@
...
+  // and then move to the directory it is cloned into.
+  chdir('/tmp');
+  console.log(
+    spawnSync('git', [
+      'clone',
+      `https://github.com/${repo.owner}/${repo.repo}.git`,
+      './branch-manager-repo',
+    ]).output.toString(),
+  );
+  chdir('/tmp/branch-manager-repo');
```
devversion, line 46: Why not always use the actual target repository (like with `pull_request_target):

* This will be safer
* In a real merge- the caretaker also acts from the main branch
* We could just use the `actions/checkout` as usually done with GitHub actions

josephperrott: The idea is that this is a "service" which the other repos are able to call to perform this check, rather than the action being expressed in every repo.  This allows for us to call it in many different ways, via action directly, via API if we need in the future, etc.

> This will be safer
Since we just want to do a static analysis of the git diffs essentially, I don't know that it will be safer, because we are not actually performing any action on the code that is checked out, so regardless of the code change being analyzed, the changes are not executed.

>In a real merge- the caretaker also acts from the main branch
This will be done from the main branch of the repository in question, however its worth noting the caretaker does not have to merge from the main branch.

>We could just use the actions/checkout as usually done with GitHub actions
actions/checkout is controlled at the repo calling it, rather than the source code of the action itself, so we would have to "manage" this configuration at all of the call sites.

.github/local-actions/branch-manager/lib/main.ts
josephperrott: Hmm okay, I will shorten this and put in a cutoff to make sure we don't cause an API failure, looks like the limit is 140 characters

.github/local-actions/branch-manager/lib/main.ts
josephperrott: Because this always is running in the dev-infra repo, the context is not the "correct" repo and owner

github-actions/utils.ts
josephperrott: See above comments.

.github/local-actions/branch-manager/lib/main.ts
josephperrott: While I really want to keep branch manager because it allows me to have the follow up with assistant to the branch manager, you are right, the context name probably will be better as something other than branch manager.

We can go with mergability.

.github/local-actions/branch-manager/lib/main.ts
devversion: To be honest (disclaimer: not feeling super strongly): It sounds a little unnecessary to have this in the dev-infra repo if it could just be a shared single action workflow file per repository (compared to having workflows per repo and one here). Also it would simplify the action, and you could still keep the workflow_dispatch even on a per repository basis, right?

josephperrott: The reason I think its worth while is because we have to have a separate workflow that exists in each repo which will inform when to run these checks, allowing us to fan our multiple pr checks to run simultaneously.  Rather than having a complex matrix usage (which also comes with github autocanceling things) to communicate which PRs should be checked.  We can just direct all of them here.

devversion: Sorry, I just need to understand this a little better. Some questions:

1. I don't fully understand what you want to fan out. If it's multiple validations- couldn't the "action code" not create multiple individual statuses without needing an actual matrix?
2. Is the fan-out worth? (need to likely understand better what you are thinking of).

josephperrott: >1
The action code can definitely apply statuses to multiple PRs from one action run, but because we need to perform checks using the file system, it is required to be serial

2>
Whenever code is pushed to the main branch of the repository, we will need to recheck all of the PRs which target that branch, for main for instance this is ~all outstanding PRs.  We would like to parallelize these validation checks, and since it requires actually running git commands in the filesystem, we cannot do them in parallel on one VM.  We instead need to fan them out into different runners.  We can do this using an outputted list from one job as an input for a matrix in another, but its very brittle, and if anything in the matrix fails, all others are cancelled.

devversion: I see, thanks. The second bullet point cleared this up.

.github/local-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,161 @@
...
+
+interface CommmitStatus {
+  state: 'pending' | 'error' | 'failure' | 'success';
+  description: string;
+}
+
+async function main(repo: {owner: string; repo: string}, token: string, pr: number) {
+  if (isNaN(pr)) {
+    core.setFailed('The provided pr value was not a number');
+    return;
```
devversion, line 33: Nit: it is odd that the validation happens here while the type contract already explicitly is set to `number`

.github/local-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,161 @@
...
+  // and then move to the directory it is cloned into.
+  chdir('/tmp');
+  console.log(
+    spawnSync('git', [
+      'clone',
+      `https://github.com/${repo.owner}/${repo.repo}.git`,
+      './branch-manager-repo',
+    ]).output.toString(),
+  );
+  chdir('/tmp/branch-manager-repo');
```
devversion, line 46: Optional nit: Could make the repository name a reusable constant

.github/local-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,161 @@
...
+  /** Whether any fatal validation failures were discovered. */
+  let hasFatalFailures = false;
+  /** The status information to be pushed as a status to the pull request. */
+  let statusInfo: CommmitStatus = await (async () => {
+    // Log validation failures and check for any fatal failures.
+    if (pullRequest.validationFailures.length !== 0) {
+      core.info(`Found ${pullRequest.validationFailures.length} failing validation(s)`);
+      for (const failure of pullRequest.validationFailures) {
+        hasFatalFailures = !failure.canBeForceIgnored || hasFatalFailures;
+        await core.group('Validation failures', async () => {
```
devversion, line 85: I'm not familar with `.group`. Does this need to be outside the for loop?

.github/local-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  // Because we want to perform this check in the targetted repository, we first need to check out the repo
+  // and then move to the directory it is cloned into.
+  chdir('/tmp');
+  console.log(
+    spawnSync('git', [
+      'clone',
+      `https://github.com/${repo.owner}/${repo.repo}.git`,
+      './branch-manager-repo',
+    ]).output.toString(),
```
devversion, line 44: We can clone with `--depth=1` for increased speed right? I assume the problem might be that we miss branches, but those could be fetched explicitly. Still much faster than fetching everything

.github/local-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,161 @@
...
+        state: 'failure',
+      };
+    }
+  })();
+
+  await git.github.repos.createCommitStatus({
+    ...repo,
+    state: statusInfo.state,
+    // Status descriptions are limited to 140 characters.
+    description: statusInfo.description.substring(0, 139),
```
devversion, line 139: FWIW: I was more referring to the visual description limit. Some stuff might not be readable.

.github/local-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,161 @@
...
+    sha: pullRequest.headSha,
+    context: 'mergeability',
+  });
+}
+
+/** The token for the angular robot to perform actions. */
+const token = await getAuthTokenFor(ANGULAR_ROBOT, true);
```
devversion, line 146: Since the action only runs for a specific repo/PR, could it just use a repo-specific installation token, instead of introducing the org-wide token mechanism (we didn't need until this PR)

devversion (commented):
Overall LGTM.

.github/local-actions/branch-manager/lib/main.ts
josephperrott: The repo specific token will not have permission to set the status on other repos within the organization.

.github/local-actions/branch-manager/lib/main.ts
josephperrott: `NaN` is a valid value for the type `number`, we could perform the check before calling it. But Its still "correct" to check it here.

.github/local-actions/branch-manager/lib/main.ts
josephperrott: The `.group` method allows you to execute a callback and occurs between a `.groupStart` and `.groupEnd` call in logging.

I think in situations like this where we are just listing out values, its better readability than the alternative.

.github/local-actions/branch-manager/lib/main.ts
josephperrott: We should be fine to do this as the `prepare` step of the `MergeStrategy` explicitly fetches the necessary branches.

.github/local-actions/branch-manager/lib/main.ts
josephperrott: Ahh gotcha, I think its something we just have to deal with because we can't be completely sure of the exact branch names that will show here.

.github/local-actions/branch-manager/lib/main.ts
devversion: Sorry, I might miss something. Couldn't you just request an installation token for the target repository? (provided via the action inputs)

josephperrott: Yes, I don't know why I didn't think to do that.... will do.

.github/local-actions/branch-manager/lib/main.ts
devversion: interesting. I expected number to not include NaN. Seems like this makes it very context-dependent.

josephperrott: While typically anywhere you use a number you can use `NaN` right? it just propagates out all over.  So `2 * NaN` is `NaN`.

So it "functions" correctly as its not an error, but its definitely not ideal.

devversion: Yes, but many applications perform an initial check- like here, but then down the line they only assume correct number operations (hence the context dependency and TS proposal for fine-grained separation here)

josephperrott: I will move it outside of the function

devversion (approved):

PR #888, opened 2022-10-28
feat(github-actions): create an assistant to the branch manager

josephperrott:
Create a github action which creates workflow dispatches when appropriate for pull requests.  Triggers checks for merge ready prs when they are updated, or when the main branch is pushed to.


.github/workflows/assistant-to-the-branch-manager.yml
```diff
@@ -0,0 +1,26 @@
+name: Assistant to the Branch Manager
+
+# Because the branch manager does not work for dev-infra, due to not using target branches, we don't
+# actually use any automatic triggers.
+on:
+  repository_dispatch:
```
devversion, line 6: is this mostly for testing then?

.github/workflows/assistant-to-the-branch-manager.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  assistant_to_the_branch_manager:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # tag=v3.0.2
+        with:
+          # Setting `persist-credentials: false` prevents the github-action account from being the
+          # account that is attempted to be used for authentication, instead the remote is set to
+          # an authenticated URL.
```
devversion, line 22: Maybe it's just my language barrier- but this reads a little confusingly. Would it be fine just saying that credentials are removed after checkout?

github-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,106 @@
...
+    }),
+  );
+}
+
+/** The authorization token for the Github app. */
+let token: string;
+/** The Octokit instance, if defined to allow token revokation after the action executes. */
+let _github: Octokit | null = null;
+/** Get the shared instance of Octokit, first creating the instance if necessary. */
+async function github() {
```
devversion, line 91: Nit: Could just pass through Github to the extra function.

github-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,106 @@
...
+  if (_github === null) {
+    token = await getAuthTokenFor(ANGULAR_ROBOT);
+    _github = new Octokit({token});
+  }
+  return _github;
+}
+
+try {
+  await run().catch((e: Error) => {
+    core.error(e);
```
devversion, line 101: I recommend using console.error. Switched to that recently. Action error function did not print full stack traces even though it said so

github-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,106 @@
...
+  pr: `${context.issue.number}`,
+};
+type WorkflowInputs = typeof pullRequestFromContext;
+
+/** Create a workflow dispatch event to trigger the pr to be evaluated for mergeability. */
+function createWorkflowForPullRequest(prInfo?: Partial<WorkflowInputs>) {
+  const inputs = {...pullRequestFromContext, ...prInfo};
+  core.info(`Requesting workflow run for: ${JSON.stringify(inputs)}`);
+  return github().then((api) =>
+    api.actions.createWorkflowDispatch({
```
devversion, line 71: should be returned right? / or awaited

devversion (approved):
Nice, clean

github-actions/branch-manager/lib/main.ts
josephperrott: There are paths through that will not need to create a Octokit instance, so we don't want to create one until we have to, so that we don't unnecessarily create access tokens for the github app.

.github/workflows/assistant-to-the-branch-manager.yml
josephperrott: Honestly just copied it from the last time I set it up.  I will rewrite it to make a bit more sense.

github-actions/branch-manager/lib/main.ts
josephperrott: From MDN:
>returns another pending promise: the fulfillment/rejection of the promise returned by then will be subsequent to the resolution/rejection of the promise returned by the handler. Also, the resolved value of the promise returned by then will be the same as the resolved value of the promise returned by the handler

Since its not in a `{}` block, it is returned as the value for the `onFufilled` calllback.

devversion: haha, yeah, no need to quote MDN. I just didn't parse that it's part of an arrow function.

github-actions/branch-manager/lib/main.ts
```diff
@@ -0,0 +1,106 @@
...
+async function run() {
+  if (context.eventName === 'push') {
+    const {ref} = context.payload as PushEvent;
+    if (ref.startsWith('refs/tags/')) {
+      core.info('No evaluation needed as tags do not cause branches to need to be rechecked');
+    }
+
+    if (ref !== 'refs/heads/main') {
+      // TODO: support pushes to all releasable branches rather than just main.
+      core.info('Skipping evaluation as the push does not affect the main branch');
```
devversion, line 21: Do you want to return here and above?

devversion (approved):

github-actions/branch-manager/lib/main.ts
josephperrott: Yep, good call

PR #895, opened 2022-11-07
fix(github-actions): labels sync action not actually checking target repos

devversion:


devversion:
@josephperrott When we merge this, I think there may be labels that need to be renamed. The whole switch wasn't performed before. e.g. FW still has `action: caretaker note`, so might need to rename first?

josephperrott (approved):
LGTM

PR #897, opened 2022-11-07
fix(github-actions): properly pass auth token to octokit

devversion:
The octokit options literaly is weakly typed and doesn't validate that `auth` is set. The token is currently provided to an unknown field that isn't respected. This is likely the reason for the bearer token header logic, so I removed this.


josephperrott (approved):
LGTM

PR #898, opened 2022-11-14
feat(github-actions): setup action for enabling bazel remote exec

devversion:
This will be useful for actions like the dev-app previews in the components repo. We wouldn't want to use CircleCI there. The action relies on the Circle logic- but that is acceptable since CircleCI is still our primary provider here.


github-actions/setup-bazel-remote-exec/action.yml
```diff
@@ -0,0 +1,17 @@
+name: 'Setup Bazel remote execution'
+description: 'Sets up Bazel remote execution.'
+author: 'Angular'
+inputs:
+  bazelrc:
```
josephperrott, line 5: Can we have a default here since its usually the same place?

devversion: Not all repositories use the same pattern where `--config=remote` should live, so it's not a default (similar in the orb)

josephperrott: 👍

PR #901, opened 2022-11-15
fix(github-actions): set remote exec script as executable and prevent regression

devversion:
Note, I considered implementing another rule like `generated_file_test` that supports ensuring file permissions- but it's not worth the effort and a simple test can be implemented to verify permissions.


josephperrott (approved):
LGTM

PR #909, opened 2022-11-28
feat(github-actions): introduce actions for safely deploying previews

devversion:
This action will be used to supersede/simplify the existing Github action workflows in `angular/components` and the AIO docker VM setup.


github-actions/deploy-previews/pack-and-upload-artifact/action.yml
```diff
@@ -0,0 +1,49 @@
...
+    required: true
+    description: |
+      Name of the artifact that should be deployed. A workflow may contain
+      multiple artifacts but only a single one can be picked for deployment.
+
+  pull-number:
+    required: true
+    description: Pull request for which the artifact is built for.
+
+  artifact-build-revision:
```
josephperrott, line 23: Is there a compelling reason to not just have this required to be the Git SHA?

github-actions/deploy-previews/pack-and-upload-artifact/inject-artifact-metadata.js
```diff
@@ -0,0 +1,34 @@
...
+  await fs.promises.writeFile(path.join(deployDirPath, artifactMetadata["pull-number"]), prNumber);
+  await fs.promises.writeFile(path.join(deployDirPath, artifactMetadata["build-revision"]), buildRevision);
+}
+try {
+  await main();
+} catch (e) {
+  console.error(e);
+  process.exit(1);
+}
+/**
```
josephperrott, line 28: Can this be at the top of the file?

github-actions/deploy-previews/pack-and-upload-artifact/lib/inject-artifact-metadata.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+/**
+ * Injects metadata information into the given unpacked artifact. An artifact
+ * is expected to contain metadata such as the pull request it was built for.
+ *
+ * The deploy job later will extract this information when it fetched the artifact.
```
josephperrott, line 13:
```suggestion
 * The deploy job later will extract this information when it fetches the artifact.
```

github-actions/deploy-previews/upload-artifacts-to-firebase/lib/fetch-workflow-artifact.ts
```diff
@@ -0,0 +1,52 @@
...
+  const artifacts = await github.actions.listWorkflowRunArtifacts({
+    owner,
+    repo,
+    run_id: workflowId,
+  });
+
+  const matchArtifact = artifacts.data.artifacts.find((artifact) => artifact.name === artifactName);
+
+  if (matchArtifact === undefined) {
+    console.error(`Could not find artifact in workflow: ${workflowId}@$${artifactName}`);
```
josephperrott, line 33:
```suggestion
    console.error(`Could not find artifact in workflow: ${workflowId}@${artifactName}`);
```

That was an extra `$` right?

github-actions/deploy-previews/pack-and-upload-artifact/action.yml
devversion: No strong reason. In most cases it's the SHA, but it's just named build-revision, could have also been `build-sha`.

github-actions/deploy-previews/pack-and-upload-artifact/inject-artifact-metadata.js
```diff
@@ -0,0 +1,34 @@
+
+import {createRequire as __cjsCompatRequire} from 'module';
+const require = __cjsCompatRequire(import.meta.url);
+
+
+//
```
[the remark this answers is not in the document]
devversion: This is the generated file 😄 ESBuild does this but we strip bazel-platform specific paths

github-actions/deploy-previews/pack-and-upload-artifact/inject-artifact-metadata.js
devversion: Same comment as above. This is the generated file

github-actions/deploy-previews/pack-and-upload-artifact/action.yml
josephperrott: Mostly by using the sha, we can determine exactly what the state of the repo was when the build occurred, thinking of "attacks" on it, it is valuable to just automatically use the sha value.

devversion: I think there are two things:

1. I wouldn't want to automatically use the SHA here because the action should not assume/require a `pull_request` event type. There may be cases where we want to trigger this action on other triggers.
2. The SHA is not super significant because anybody could change this value to something else (e.g. an attacker). And in the case of previews for non-PRs a SHA might also not be available/easy to retrieve.

I'm happy to change this to `build-sha` but automatically using it from the event payload, and hence assuming it- I would prefer to avoid

josephperrott: ahh gotcha, that works for me

josephperrott (approved):
LGTM

PR #912, opened 2022-11-28
feat(github-actions): create a unified status action

josephperrott:
Create a github action which tracks all of the statuses on a pull request to create a unified status value.


github-actions/unified-status-check/action.yml
```diff
@@ -0,0 +1,11 @@
+name: 'Unified Status Check'
+description: 'Monitor the statuses on a pull request to ensure all expected stauses are passing'
```
devversion, line 2: nit typo

github-actions/unified-status-check/action.yml
```diff
@@ -0,0 +1,11 @@
...
+author: 'Angular'
+inputs:
+  angular-robot-key:
```
devversion, line 5: should the action not take a list of statuses that should be enforced? this would be similar to the existing robot and I think this is the key feature.

(I know at the same time GitHub branch protections could be used too- even for the current code in this PR? do you recall what the value of the status was?)

github-actions/unified-status-check/src/main.ts
```diff
@@ -0,0 +1,194 @@
...
+  ['NEUTRAL', 'EXPECTED'],
+  ['SKIPPED', 'EXPECTED'],
+  ['STALE', 'ERROR'],
+  ['STARTUP_FAILURE', 'FAILURE'],
+  ['SUCCESS', 'SUCCESS'],
+  ['TIMED_OUT', 'FAILURE'],
+]);
+
+/** GraphQL schema for requesting the status information for a given pull request. */
+const PR_SCHEMA = params(
```
devversion, line 135: isn't all of this similar in ng-dev? can we just use that from there or dedupe?

josephperrott: We could, but to do so we would have to use the `AuthenticatedGithubApi` class that is in `ng-dev`, which is more complicated because of the access to the configuration files.

github-actions/unified-status-check/action.yml
josephperrott: I think we should flip the model and have it be an ignore list rather than a include list.

devversion: I don't know I'd agree with it. I think we should clarify why we have the action vs. branch protection required status checks.

Even with a list of "required checks", all other non-required could be incorporated. I think that is how it's currently working in the gh robot.

github-actions/unified-status-check/src/main.ts
devversion: The client does not depend on configs AFAIK: https://github.com/angular/dev-infra/blob/main/ng-dev/utils/git/github.ts

josephperrott: I was thinking of the `AuthenticatedGitClient` rather than the Github one.  Will see what I can do.

josephperrott: From looking into it more, we can leverage the ng-dev stuff, but we don't really save ourselves that much code since we still have to describe what the shape of the PR response we want is.

github-actions/unified-status-check/action.yml
devversion: > My understanding/intention with this is that we use this to enforce "expected" statuses, but allowing for us to build out the contextual awareness of the check. For instance, allowing for ignoring when the pull request is in a draft state.

That's interesting. At the current time our contextual awareness would only be `if (isDraft) ? A : B`. With the current configuration model I cannot imagine more conditional enforcements (that can change) -- I see that argument 👍 

> As for the ignore rather than include list, I think its a better model to always include the checks and intentionally opt them out rather than creating a brittle set up of a name change in one of the statuses making it impossible for it to pass.

I would disagree with this. A good example is if e.g. CircleCI is not triggered at all, then the check would not see any failure/pending statuses and would show as "green".

josephperrott: I suppose thats true, I wonder if there is a way to meet in the middle of it so that we don't create a brittle system but we make sure that we don't have false successes from just nothing running.

Let me see what I can do on that.

josephperrott: Added the ability to do both ignore or match for required

github-actions/unified-status-check/action.yml
```diff
@@ -0,0 +1,17 @@
+name: 'Unified Status Check'
+description: 'Monitor the statuses on a pull request to ensure all expected statuses are passing'
+author: 'Angular'
+inputs:
+  angular-robot-key:
+    description: 'The private key for the Angular Robot Github app.'
+    required: true
+  ignored:
+    description: 'Status matchers to be ignored.'
```
devversion, line 9: Nit: Can you clarify in the description what a matcher is. e.g. regex pattern or what it takes. same below

github-actions/unified-status-check/src/main.ts
```diff
@@ -0,0 +1,202 @@
...
+      })
+    ).repository.pullRequest;
+    /** The status check rollup results. */
+    const statusChecks = pullRequest.commits.nodes[0].commit.statusCheckRollup;
+
+    const setStatus = async (
+      state: 'pending' | 'error' | 'failure' | 'success',
+      description?: string,
+    ) => {
+      await github.repos.createCommitStatus({
```
devversion, line 35: One interesting thing I just realized. IIRC the Angular GitHub stored the status in Firebase right?

Problem is that calling `createCommitStatus` too often will quickly flood a commit with too many statuses, ultimately preventing more statuses to be created. I faced this in the presubmit status service when it was not debounced, so now a new status on the commit is only applied if the state changes.

Just calling this out- we might be able to hit this, if we update whenever statuses change. That can evolve pretty quickly.

github-actions/unified-status-check/src/main.ts
```diff
@@ -0,0 +1,202 @@
...
+        context: unifiedStatusCheckName,
+        state,
+        description,
+      });
+      return;
+    };
+
+    /** If no status checks are present, or if the pull request is in a draft state the unified status is in a pending state. */
+    if (!statusChecks || pullRequest.isDraft) {
+      await setStatus('pending', 'Unified Status checks have not yet started');
```
devversion, line 47: Nit: I think we should just say `PR is still a draft` and another message for the other case.

github-actions/unified-status-check/src/main.ts
```diff
@@ -0,0 +1,202 @@
...
+            state: checkOrStatus.state,
+            name: checkOrStatus.context,
+          };
+        }
+        if (checkOrStatus.__typename === 'CheckRun') {
+          return {
+            state: checkOrStatus.conclusion
+              ? checkConclusionStateToStatusStateMap.get(checkOrStatus.conclusion)!
+              : 'PENDING',
+            name: checkOrStatus.name || 'Unknown Check Run',
```
devversion, line 64: Are the status names with spaces allowed (if real ones from GitHub). Maybe use `<unknown-check-run>` or something? (not seeing any issue with this currently- but it would be more consistent then)

github-actions/unified-status-check/src/main.ts
```diff
@@ -0,0 +1,202 @@
...
+      })
+      .filter(({name}) => !ignored.includes(name));
+
+    const missedStatuses = required.filter(
+      (matcher) => !statuses.some(({name}) => name.match(matcher)),
+    );
+    if (missedStatuses.length > 0) {
+      await setStatus(
+        'pending',
+        `Waiting for missing required status: ${missedStatuses.join(', ')}`,
```
devversion, line 77: There is a limit in the status message IIRC. Might want to just say `Pending statuses: X` and also cut if needed?

devversion: Maybe also print the count at the beginning as that is always visible then ?

github-actions/unified-status-check/src/main.ts
```diff
@@ -0,0 +1,202 @@
...
+      {passing: 0, failing: 0, pending: 0},
+    );
+
+    if (counts.failing > 0) {
+      await setStatus('failure', `${counts.failing} expected status(es) failing`);
+      return;
+    }
+
+    if (counts.pending > 0) {
+      await setStatus('pending', 'Expected statuses are still pending');
```
devversion, line 104: Expected is ambiguous IMO because they are not declared as "required". Maybe, `Other statuses are still pending`?

github-actions/unified-status-check/src/main.ts
devversion: I thought we are already using a similar schema (almost identical, maybe with a bit more stuff). If that's not the case, then yes. I'm fine keeping it here. Maybe want to extract it into a separate file for more readability? it's odd having this at the end of a file too (not relying on hoisting for better readability)

devversion (approved):
LGTM. some minor suggestions, but nice work!

github-actions/unified-status-check/src/main.ts
josephperrott: Added in a check for if we are just doing the same update on the status.

github-actions/unified-status-check/src/main.ts
josephperrott: I will do the refactor into another file in a follow up because I want to expand a bit of what its doing and I would rather land this as is for now.

devversion: no worries, sgtm

devversion (approved):

PR #915, opened 2022-11-29
fix(github-actions): do not nest deploy directory in final tmp dir

devversion:
The github action creates a temporary directory for adding the metadata files. Right now the cp logic is incorrect and nestes the base folder inside it. We can ensure the deploy folder becomes the tmp dir 1:1 by not creating it before copying.

### DIFF
--- a/github-actions/deploy-previews/pack-and-upload-artifact/action.yml
+++ b/github-actions/deploy-previews/pack-and-upload-artifact/action.yml
@@ -39,8 +39,7 @@ runs:
       id: copy
       shell: bash
       run: |
-        dir="$RUNNER_TEMP/pack-and-upload-tmp-dir"
-        mkdir -p $dir
+        dir="$RUNNER_TEMP/pack-and-upload-tmp-dir/"
         cp -R "${{inputs.deploy-directory}}" "$dir"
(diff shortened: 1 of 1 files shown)


github-actions/deploy-previews/pack-and-upload-artifact/action.yml
```diff
@@ -39,8 +39,7 @@ runs:
       id: copy
       shell: bash
       run: |
-        dir="$RUNNER_TEMP/pack-and-upload-tmp-dir"
-        mkdir -p $dir
+        dir="$RUNNER_TEMP/pack-and-upload-tmp-dir/"
```
josephperrott, line 42: I am a bit confused about how this prevents the nesting?

devversion: `cp A B` will create `B/A` if `B` already exists. If `B` would not exist, `A` will be copied directly as `B`

josephperrott: Oh okay, didn't realize that. I think when I accidentally find myself in that situation I just erase and try again.

josephperrott (approved):
LGTM

PR #921, opened 2022-12-02
Unified status check refactors

josephperrott:
This PR does some refactoring to pull some of the logic into different files.

Of note, I created a "validation" concept which allows us to easily move toward running multiple validators and return multiple results rather than exiting on first failure.

Additionally, this could set us up nicely in the future for moving this result into a Github check if we want to provide more information that whats contained in a regular status.


github-actions/unified-status-check/src/main.ts
```diff
@@ -79,7 +76,7 @@ async function main() {
     };
 
     /** If no status checks are present, or if the pull request is in a draft state the unified status is in a pending state. */
-    if (pullRequest.isDraft) {
+    if (isDraft(pullRequest).state === 'PENDING') {
```
devversion, line 79: Did you mean to use the message from `isDraft()`?

devversion (approved):

PR #922, opened 2022-12-05
refactor(github-actions): extract pull request status validations

josephperrott:


github-actions/unified-status-check/src/main.ts
```diff
@@ -2,58 +2,24 @@ import * as core from '@actions/core';
...
-          };
-        }
-        throw Error('CheckOrStatus was not found to be a check or a status');
-      });
-    const statuses = allStatuses.filter(({name}) => !ignored.includes(name));
-    const unifiedCheckStatus = allStatuses.find(({name}) => name === unifiedStatusCheckName);
-
-    const setStatus = async (state: StatusState, description?: string) => {
+    const setStatus = async (
+      state: Omit<StatusState, 'ERROR' | 'EXPECTED'>,
```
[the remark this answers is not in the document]
devversion: I think we should translate it directly. It doesn't seem type-safe the current way, looking at the cast below

github-actions/unified-status-check/src/main.ts
```diff
@@ -67,92 +29,37 @@ async function main() {
...
-          count.failing += 1;
-        }
-        return count;
-      },
-      {passing: 0, failing: 0, pending: 0},
-    );
-
-    if (counts.failing > 0) {
-      await setStatus('FAILURE', `${counts.failing} expected status(es) failing`);
+    const hasRequiredStatusesResult = checkRequiredStatuses(pullRequest);
```
devversion, line 47: Nit:

```suggestion
    const requiredStatusesResult = checkRequiredStatuses(pullRequest);
```

github-actions/unified-status-check/src/main.ts
```diff
@@ -67,92 +29,37 @@ async function main() {
...
+    if (hasRequiredStatusesResult.state === 'pending') {
+      await setStatus(hasRequiredStatusesResult.state, hasRequiredStatusesResult.description);
       return;
     }
 
-    if (counts.pending > 0) {
-      await setStatus('PENDING', 'Other tracked statuses are still pending');
+    const hasPassingStatusesResult = checkOnlyPassingStatuses(pullRequest);
```
devversion, line 53:
```suggestion
    const onlyPassingStatusesResult = checkOnlyPassingStatuses(pullRequest);
```

github-actions/unified-status-check/src/main.ts
```diff
@@ -67,92 +29,37 @@ async function main() {
...
-        }
-        return count;
-      },
-      {passing: 0, failing: 0, pending: 0},
-    );
-
-    if (counts.failing > 0) {
-      await setStatus('FAILURE', `${counts.failing} expected status(es) failing`);
+    const hasRequiredStatusesResult = checkRequiredStatuses(pullRequest);
+    if (hasRequiredStatusesResult.state === 'pending') {
```
devversion, line 48: Out of curiosity: Do you want to use `!== "success"` here? What if the required status result would technically return `failure`? It seems quite context-dependent (i.e. need to know what this function is doing)

github-actions/unified-status-check/src/pull-request.ts
```diff
@@ -1,9 +1,82 @@
...
+  failing: NormalizedStatus[] = [];
+  passing: NormalizedStatus[] = [];
+  ignored: NormalizedStatus[] = [];
+  all: NormalizedStatus[] = [];
+  unifiedCheckStatus: NormalizedStatus | undefined;
+
+  populate(statuses: NormalizedStatus[]) {
+    const ignored = [
+      unifiedStatusCheckName,
+      ...core.getMultilineInput('ignored', {trimWhitespace: true}),
```
devversion, line 25: Do you just want to do this in the class field initializers or constructor?

devversion (approved):
LGTM, but some minor comments (and one question about `failure`)

github-actions/unified-status-check/src/main.ts
josephperrott: I am going to leave it for now as I will end up redoing this logic in a follow up so that we can get the combined result of multiple validations.

devversion: Sounds good. Just be aware that `${statuses.failing} expected status(es) failing` is not working right now then.

PR #927, opened 2022-12-08
feat(github-actions): add label checks to unified status check

josephperrott:
This pull requests adds the labels checks to the already established pieces, a follow up with contain the logic for checking multiple validations at a time rather than eager exiting.

### DIFF
--- a/github-actions/unified-status-check/src/labels.ts
+++ b/github-actions/unified-status-check/src/labels.ts
@@ -1,0 +1,47 @@
+import {actionLabels, targetLabels} from '../../../ng-dev/pr/common/labels.js';
+import {PullRequest} from './pull-request.js';
+import {ValidationFunction} from './validation.js';
+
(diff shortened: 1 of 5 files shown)


github-actions/unified-status-check/src/labels.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (pullRequest.labels.includes(actionLabels.ACTION_CLEANUP.name)) {
+    return {
+      state: 'failure',
+      description: `Pull request is marked for merge but still has the "${actionLabels.ACTION_CLEANUP.name}" label on the pull request`,
+    };
+  }
+
+  return {
+    state: 'success',
+    description: `Pull request is marked for merge by the "${actionLabels.ACTION_MERGE.name}" label on the pull request`,
```
devversion, line 45: I believe these messages are way too long to display. Can you try shorten these?

devversion (approved):

PR #928, opened 2022-12-08
Misc improvements

devversion:
See individual commits


bazel/spec-bundling/spec-bundle.bzl
```diff
@@ -0,0 +1,106 @@
...
+    """
+
+    is_browser_test = platform == "browser"
+    is_legacy_cjs = platform == "cjs-legacy"
+    is_node = platform == "node"
+
+    package_name = native.package_name()
+    esbuild_attrs = dict()
+
+    spec_entrypoint(
```
devversion, line 46: Note: This code is totally untouched but just has been moved from `index.bzl` into a new file.

crisbeto (approved):

PR #934, opened 2022-12-09
fix(github-actions): only create a single preview channel per PR

devversion:
This reduces the amount of preview channels we create, avoiding
quota issues. See: https://b.corp.google.com/issues/261985459


josephperrott (approved):
LGTM

PR #935, opened 2022-12-12
Feature: Downlevel native/async await and allow for ES2020 target

devversion:


bazel/spec-bundling/esbuild.config-tmpl.mjs
```diff
@@ -26,12 +26,23 @@ async function fetchAndCreateLinkerEsbuildPlugin() {
...
 const plugins = TMPL_RUN_LINKER ? [await fetchAndCreateLinkerEsbuildPlugin()] : [];
 
+// List of supported features as per ESBuild. See:
+// https://esbuild.github.io/api/#supported.
+const supported = {};
+
+// Async/Await can be downleveled so that ZoneJS can intercept. See:
+// https://github.com/angular/angular-cli/blob/afe9feaa45913/packages/angular_devkit/build_angular/src/builders/browser-esbuild/index.ts#L313-L318.
+if (TMPL_DOWNLEVEL_ASYNC_AWAIT) {
+  supported['async-await'] = false;
```
josephperrott, line 36: Just to make sure here, if we indicate to downlevel async-await, we mark it as not supported?

bazel/spec-bundling/test/async-await.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  it('should properly detect changes', fakeAsync(() => {
+    TestBed.configureTestingModule({
+      declarations: [AppComponent],
+    });
+    const fixture = TestBed.createComponent(AppComponent);
+    const el = fixture.nativeElement as HTMLElement;
+
+    fixture.detectChanges();
```
josephperrott, line 31: Optional: Can you put in comments on why we need to detect changes and flush in this specific order?

josephperrott (approved):
LGTM

bazel/spec-bundling/esbuild.config-tmpl.mjs
devversion: Correct

bazel/spec-bundling/test/async-await.spec.ts
devversion: Done.

shared-scripts/angular-optimization/ensure-no-linker-decl.mjs
```diff
@@ -0,0 +1,36 @@
...
+    Identifier: (astPath) => {
+      if (isNgDeclareCallExpression(astPath)) {
+        throw astPath.buildCodeFrameError(
+          `Found Angular declaration that has not been linked. ${filePath}`,
+          Error,
+        );
+      }
+    },
+  });
+}
```
devversion, line 36: This is the existing linker logic, just moved into a separate file.

shared-scripts/angular-optimization/esbuild-plugin.d.ts
```diff
@@ -0,0 +1,18 @@
...
+export interface OptimizationOptions {
+  enableLinker?: {
+    ensureNoPartialDeclaration: boolean;
+    linkerOptions?: object;
+  };
+  optimize?: {
+    isSideEffectFree?: (absoluteDiskPath: string) => boolean;
+  };
+  downlevelAsyncGeneratorsIfPresent?: boolean;
+}
```
devversion, line 18: The optimization plugin is a `.mjs` file and it needs to stay like that. The `d.ts` is just for our own documentation and potential downstream tools can even leverage it.

Long-term we might be able to replace this with an actual `.mts` file.

devversion:
@josephperrott I've fixed one more issue and simplified/improved the esbuild setup. It's a little more unreadable (I acknowledge that) but it's a necessary perf optimization and isn't much different in CLI.

josephperrott (approved):
LGTM

PR #936, opened 2022-12-12
refactor(github-actions): use class for pull request instead of object

josephperrott:


github-actions/unified-status-check/src/pull-request.ts
```diff
@@ -5,59 +5,146 @@ import {CheckConclusionState, PullRequestState, StatusState} from '@octokit/grap
...
+      head_sha: this.sha,
+      status: state === 'pending' ? 'in_progress' : 'completed',
+      conclusion: state === 'pending' ? undefined : state,
+      output: {
+        summary: description,
+      },
+    };
+
+    if (this.previousRunId) {
+      await this.github.checks.update({check_run_id: this.previousRunId, ...parameters});
```
devversion, line 141: Ahh, and you changed it from a status to a check 👍 (commit didn't  say that - but nice!)

github-actions/unified-status-check/src/pull-request.ts
```diff
@@ -5,59 +5,146 @@ import {CheckConclusionState, PullRequestState, StatusState} from '@octokit/grap
...
+          state: checkOrStatus.conclusion
+            ? checkConclusionStateToStatusStateMap.get(checkOrStatus.conclusion)!
+            : 'pending',
+          description: checkOrStatus.title || undefined,
+          name: checkOrStatus.name || 'unknown-check-run',
+        };
+
+        // If the check is the previous result for the unified check, store the previous run id.
+        if (checkOrStatus.name === PullRequest.checkName) {
+          this.previousRunId = checkOrStatus.databaseId;
```
devversion, line 90: is this really called `databaseId`? just asking out of curiosity. seems odd from GitHub

devversion (approved):

github-actions/unified-status-check/src/pull-request.ts
josephperrott: Yeah, it seems to be.  I was surprised as well.

PR #937, opened 2022-12-12
fix(bazel): spec-bundling rule should keep original names

devversion:
Kudos to @josephperrott who found this back when we used ESBuild for legacy saucelabs bundle creation

https://angular-team.slack.com/archives/D0ZBU43EZ/p1636491354025800


josephperrott (approved):
LGTM

PR #938, opened 2022-12-12
feat(github-actions): merge all validation results instead of exiting on first non-passing validation

josephperrott:


devversion (approved):
Nice, that's cool!

PR #939, opened 2022-12-12
fix(bazel): app bundle rule does not declare other files as side-effect free

devversion:
Fixes a typo that prevented non-entry-point files from being marked as side-effect free.


josephperrott (approved):
LGTM

PR #940, opened 2022-12-12
fix(shared-scripts): properly use `isSideEffectFree` function

devversion:
There was another typo, preventing use of `isSideEffectFree`.


josephperrott (approved):
LGTM

PR #944, opened 2022-12-13
test: update size golden after `isSideEffectFree` fix

devversion:
The size test changed when we fixed `isSideEffectFree`. Note that I investigated why there is a minimal increase even though we should treat more files as side-effect-free. The reason is that due to the pure annotations Terser optimizes slightly different. In larger apps, this would be a size reduction but here due to slightly differently optimized code (i.e. some functions are inlined / some not)- it causes a small increase.

e.g. rxjs `createErrorClass` is only used once but now is no longer inlined, hence having a `function` modifier compared to it previously being inlined without the function declaration overhead.

### DIFF
--- a/bazel/map-size-tracking/test/size-golden.json
+++ b/bazel/map-size-tracking/test/size-golden.json
@@ -1,5 +1,5 @@
 {
-  "unmapped": 6101,
+  "unmapped": 6117,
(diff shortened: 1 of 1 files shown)


josephperrott (approved):
LGTM

PR #950, opened 2022-12-14
fix(bazel): http server should support loading jsonp files

devversion:
An actual file may be resolved, but with URL params. The params should be omitted when we resolve a file from the runfiles.


josephperrott (approved):
LGTM

PR #955, opened 2022-12-16
feat(bazel): support explicit files in spec-entrypoint

devversion:
This will be helpful for using the rule in `jasmine_node_test` where explicit specs may be passed through `srcs`.


josephperrott (approved):
LGTM

PR #956, opened 2022-12-16
feat(bazel): always consider transitive specs for spec-entrypoint

devversion:
We recently switched all from transitive to just direct. We should keep this for the bootstrap code as there is a legitimate ordering issue if a bootstrap file depends on another and they would then be brought into with an unexpected order.

For specs we should consider all transitive ones too, just to be on the safe side of never missing any specs.

### DIFF
--- a/bazel/spec-bundling/test/should_be_picked_up.spec.ts
+++ b/bazel/spec-bundling/test/should_be_picked_up.spec.ts
@@ -1,0 +1,9 @@
+// See `ensure-transitive-loaded` for explanation. This verifies this
+// file is actually loaded.
+(global as any).iWillAlwaysFail = false;
+
+describe('should be picked up', () => {
+  it('should work', () => {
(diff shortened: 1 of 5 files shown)


bazel/spec-bundling/test/should_be_picked_up.spec.ts
```diff
@@ -0,0 +1,11 @@
+// See `ensure-transitive-loaded` for explanation. This verifies this
+// file is actually loaded.
+(global as any).iWillAlwaysFail = false;
+
+console.error('LOaded');
```
josephperrott, line 5: We can drop this console error right?

josephperrott (approved):
LGTM

bazel/spec-bundling/test/should_be_picked_up.spec.ts
devversion: good catch. done

PR #958, opened 2022-12-20
Initial work for supporting exceptional minors

devversion:
* Supports detection of exceptional minors
* Reworks tests to handle much more scenarios & be easier to write/understand
* Cleans up the old exceptional minor RC/FF hack.
* Fixes the merge tool target label/branch error handling
* Switches away from duplicating labels to the dev-infra canonical labels (this was an oversight when the label definitions were moved into dev-infra I think)


ng-dev/release/versioning/version-branches.ts
```diff
@@ -60,33 +69,28 @@ export async function getVersionOfBranch(
   if (!content) {
     throw Error(`Unable to read "package.json" file from repository.`);
   }
-  const {version} = JSON.parse(Buffer.from(content, 'base64').toString()) as {
+  const {version, __ngDevExceptionalMinor__} = JSON.parse(
```
josephperrott, line 72: This deconstruction works because we define the value for `exceptionalMinorPackageIndicator` on line 43, but if we were to change it there this deconstruction fails right?

Should we just put the string directly in the key for the object?

josephperrott (approved):
LGTM

A couple comments

ng-dev/release/versioning/version-branches.ts
devversion: oops, very good catch. that's not supposed to be like this

PR #959, opened 2022-12-21
refactor(ng-dev): simplify pre-release version compute and allow for reuse of pre-release actions

devversion:
See individual commits


ng-dev/release/publish/actions/shared/branch-off-next-branch.ts
```diff
@@ -8,35 +8,46 @@
...
+    this.git,
+    this.config,
+    this.projectDir,
+  );
+  private _rcPrerelease = new CutNpmNextReleaseCandidateAction(
+    new ActiveReleaseTrains({...this.active, releaseCandidate: this.active.next}),
+    this.git,
+    this.config,
+    this.projectDir,
+  );
```
devversion, line 49: Note: I think this makes this action simpler to understand. Main reason it's done is to avoid relying on functions like  `computeNewPrelreleaseVersionForNext`. This method doesn't fit well into making the pre-release actions "re-usable". Conceptually it makes sense relying on these actions since it's basically what we perform too.

ng-dev/release/publish/actions/shared/cut-prerelease.ts
```diff
@@ -0,0 +1,95 @@
+/**
```
devversion, line 1: FYI: This file and the followed ones are 99% the same, except:

* For the abstract class members
* Removal of `isActive` — that's left to the implementing classes
* Usage of the abstract members when determining `newVersion` and `releaseNotesCompareVersion`.

ng-dev/release/versioning/npm-command.ts
```diff
@@ -17,12 +17,20 @@ export abstract class NpmCommand {
    * Runs NPM publish within a specified package directory.
    * @throws With the process log output if the publish failed.
    */
-  static async publish(packagePath: string, distTag: NpmDistTag, registryUrl: string | undefined) {
-    const args = ['publish', '--access', 'public', '--tag', distTag];
-    // If a custom registry URL has been specified, add the `--registry` flag.
+  static async publish(
+    packagePath: string,
+    distTag: NpmDistTag | null,
```
josephperrott, line 22: Does it make more sense to decouple publish and tagging? And we publish and then tag the newly published version, rather than having the login for whether or not to tag be in the publish logic?

josephperrott (approved):
LGTM

ng-dev/release/versioning/npm-command.ts
devversion: Good question. I believe the answer is that the decoupled tagging is already possible (we use it for lts tagging), but it doesn't hurt supporting here in the signature of publishing. Especially since it's the common case to set a tag when publishing (and the signature is also a great reminder to not forget about tagging)

PR #970, opened 2023-01-09
fix(ng-dev/auth): remove loopback authentication schema

josephperrott:
Remove the loopback authentication schema as a method of authentication as we can continue to use the device code method and this reduces complexity and maintainence.


devversion (approved):
Always great!

PR #979, opened 2023-01-12
feat(ng-dev): allow for additional custom stamping

devversion:
This allows downstream projects to have custom stamping properties along with the normal stamping. e.g. when a framework peer dependency variable for substitution should be computed based on the current project version.


josephperrott (approved):
LGTM

PR #983, opened 2023-01-12
ci: enable renovate cache

alan-agius4:
Renovate provides a way to cache repo data between runs using the `persistRepoData` option. This should help reduce `rate-limit-exceeded` errors in the renovate action which are causing it to fail silently.

See: https://docs.renovatebot.com/self-hosted-configuration/#persistrepodata

### DIFF
--- a/.github/workflows/ng-renovate.yml
+++ b/.github/workflows/ng-renovate.yml
@@ -16,13 +16,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # tag=v3.3.0
+      - name: Get current date
+        id: date
+        run: echo "date=$(date +'%Y-%m-%d')" >> $GITHUB_OUTPUT
+      # Note: we use `date` as part of the key hash so that we create a new cache every day.
+      # We do not use `restore-keys` so that the cache does not keep growing.
       - uses: actions/cache@v3
         with:
           path: |
             .github/ng-renovate/.yarn/cache
-          key: v4-${{hashFiles('.github/ng-renovate/yarn.lock')}}
-          restore-keys: v4-
-
+            /tmp/renovate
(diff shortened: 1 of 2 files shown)


.github/workflows/ng-renovate.yml
```diff
@@ -16,13 +16,14 @@ jobs:
...
+      # Note: we use `github.run_id` as part of the key hash so that we create a new cache after each run.
       - uses: actions/cache@v3
         with:
           path: |
             .github/ng-renovate/.yarn/cache
-          key: v4-${{hashFiles('.github/ng-renovate/yarn.lock')}}
-          restore-keys: v4-
-
+            /tmp/renovate
+          key: v2-renovate-${{hashFiles('.github/ng-renovate/yarn.lock')}}-${{ github.run_id }}
```
alan-agius4, line 25: Maybe a daily cache would suffice?

devversion: yeah, I feel like that might be simpler.

alan-agius4: Updated to use date

.github/workflows/ng-renovate.yml
```diff
@@ -16,13 +16,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # tag=v3.3.0
+      - name: Get current date
+        id: date
+        run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
```
devversion, line 21: Nit set output is deprecated: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

devversion (approved):

.github/workflows/ng-renovate.yml
alan-agius4: Thanks, updated.

alan-agius4:
@devversion, ready for merge.

devversion (approved):

PR #1006, opened 2023-02-09
ci: setup the bes for remote executions

josephperrott:
Set up BES for remote executions


devversion (approved):

devversion:
FYI: Might be good explaining in the commit message/or PR _why_ we are doing this (for later referencing)

PR #1015, opened 2023-02-14
Update post approval changes

josephperrott:
See individual commits.

### DIFF
--- a/github-actions/post-approval-changes/lib/main.ts
+++ b/github-actions/post-approval-changes/lib/main.ts
@@ -118,7 +118,7 @@ async function isGooglerOrgMember(client: Octokit, username: string): Promise<bo
     return isGooglerOrgMemberCache.get(username)!;
   }
   return await client.orgs
-    .checkMembershipForUser({org: 'googler', username})
+    .checkMembershipForUser({org: 'googlers', username})
     .then(
       ({status}) => (status as number) === 204,
       () => false,
--- a/github-actions/post-approval-changes/lib/main.ts
+++ b/github-actions/post-approval-changes/lib/main.ts
@@ -54,13 +54,17 @@ async function runPostApprovalChangesAction(client: Octokit): Promise<void> {
   }
   const {pull_request: pr} = context.payload as PullRequestEvent;
 
-  if (googlers.includes(pr.user.login)) {
-    core.info('PR author is a googler, skipping as post approval changes are allowed.');
+  const actionUser = context.actor;
+
+  if (googlers.includes(actionUser)) {
+    core.info('Action performed by a googler, skipping as post approval changes are allowed.');
     return;
   }
 
-  if (googleOwnedRobots.includes(pr.user.login)) {
-    core.info('PR author is a robot owned by Google. Post approval changes are allowed.');
+  if (googleOwnedRobots.includes(actionUser)) {
+    core.info(
+      'Action performed by a robot owned by Google, skipping as post approval changes are allowed.',
+    );
     return;
   }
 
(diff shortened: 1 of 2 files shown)


github-actions/post-approval-changes/lib/main.ts
```diff
@@ -131,6 +110,25 @@ async function runPostApprovalChangesAction(client: Octokit): Promise<void> {
...
 
+/** Set of membership lookup results, used as cache for lookups. */
+const isGooglerOrgMemberCache = new Map<string, boolean>([]);
+
+async function isGooglerOrgMember(client: Octokit, username: string): Promise<boolean> {
+  if (isGooglerOrgMemberCache.has(username)) {
+    return isGooglerOrgMemberCache.get(username)!;
+  }
+  return await client.orgs
+    .checkMembershipForUser({org: 'googler', username})
```
devversion, line 121: This means everybody needs to have the membership public. I suggested that initially, so I'm good with it.

Though, two questions:

* Do you want to make sure we use a token that has access to the organization, so that we can even see for Googlers having their visibility private? obviously some googlers might not be part of the org at all..
* Are we announcing the change if we are doing it?

github-actions/post-approval-changes/lib/main.ts
```diff
@@ -131,6 +110,25 @@ async function runPostApprovalChangesAction(client: Octokit): Promise<void> {
...
+    .then(
+      ({status}) => (status as number) === 204,
+      () => false,
+    )
+    .then((result) => {
+      isGooglerOrgMemberCache.set(username, result);
+      return result;
```
devversion, line 128: Nit: Why not use await/try/finally like we do in most places in the repo?

devversion (approved):

github-actions/post-approval-changes/lib/main.ts
devversion:
```suggestion
    .checkMembershipForUser({org: 'googlers', username})
```

github-actions/post-approval-changes/lib/main.ts
josephperrott: I think it ends up less readable in this case:

```ts
async function isGooglerOrgMember(client: Octokit, username: string): Promise<boolean> {
  if (isGooglerOrgMemberCache.has(username)) {
    return isGooglerOrgMemberCache.get(username)!;
  }
  let result: boolean;
  try {
    result = await client.orgs.checkMembershipForUser({org: 'googler', username}).then(
      ({status}) => (status as number) === 204,
      () => false,
    );
  } catch {
    result = false;
  } finally {
    isGooglerOrgMemberCache.set(username, result!);
    return result!;
  }
}
```

github-actions/post-approval-changes/lib/main.ts
josephperrott: I can work on getting the token access for checking private membership in Googler org.  But for now we can rely on public.  And all Googlers are supposed to register internally which places them in that org, so it allows for us to have a nice proxy for checking the internal database.


We definitely should announce if it changes.

devversion: Haha yes, I was suggesting that GitHub org initially: https://docs.google.com/document/d/1Ab91dcLx87fNmqeiSW-CurZVDdSHYl_P5gmgQ_jtEWE/edit?usp=drivesdk&resourcekey=0-k_m13wmNFURdP-1bbVX_pA&disco=AAAAjO1Vazg

Yes, we should try supporting a token here for private visibility & announce this because not everybody is public in that org.

github-actions/post-approval-changes/lib/main.ts
devversion: I think it could be simplified further, but that's not a big deal either.

devversion (approved):

github-actions/post-approval-changes/lib/main.ts
josephperrott: I have set up the access for our Github App on the `googlers` organization. So now we are able to check private and public membership.

PR #1024, opened 2023-02-16
refactor(ng-dev): use the graphql method from octokit instead of standalone graphql

josephperrott:
Simplify what we do by just using the graphql method rather than creating our own.

### DIFF
--- a/ng-dev/utils/git/github.ts
+++ b/ng-dev/utils/git/github.ts
@@ -8,7 +8,6 @@
 
 import type {OctokitOptions} from '@octokit/core/dist-types/types.js';
 
-import {graphql} from '@octokit/graphql';
 import {Octokit} from '@octokit/rest';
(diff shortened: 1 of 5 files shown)


ng-dev/utils/git/github.ts
```diff
@@ -8,7 +8,6 @@
 
 import type {OctokitOptions} from '@octokit/core/dist-types/types.js';
 
-import {graphql} from '@octokit/graphql';
```
devversion, line 11: Can we drop the `BUILD.bazel` dependency and extra npm install maybe?

Also: We expose some class from `@octokit/graphql` for error handling (in some other util file). Do we know this will still work, given it being potentially a different instance? (e.g. `instanceof` may no longer work)

josephperrott:
Ends up that octokit just relies on graphql elsewhere so its the same thing anyway.

PR #1038, opened 2023-02-23
Atomic push

josephperrott:
See individual commits

### DIFF
--- a/ng-dev/pr/merge/strategies/autosquash-merge.ts
+++ b/ng-dev/pr/merge/strategies/autosquash-merge.ts
@@ -77,10 +77,12 @@ export class AutosquashMergeStrategy extends MergeStrategy {
     const localBranch = this.getLocalTargetBranchName(githubTargetBranch);
     /** The SHA of the commit pushed to github which represents closing the PR. */
     const sha = this.git.run(['rev-parse', localBranch]).stdout.trim();
+    // Wait five seconds to account for rate limiting of the token usage during merge.
+    await new Promise((resolve) => setTimeout(resolve, 5000));
     // Github automatically closes PRs whose commits are merged into the main branch on Github.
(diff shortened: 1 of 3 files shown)


ng-dev/pr/merge/strategies/autosquash-merge.ts
```diff
@@ -77,10 +77,12 @@ export class AutosquashMergeStrategy extends MergeStrategy {
     const localBranch = this.getLocalTargetBranchName(githubTargetBranch);
     /** The SHA of the commit pushed to github which represents closing the PR. */
     const sha = this.git.run(['rev-parse', localBranch]).stdout.trim();
+    // Wait five seconds to account for rate limiting of the token usage during merge.
+    await new Promise((resolve) => setTimeout(resolve, 5000));
```
devversion, line 81: five seconds is pretty long :(

when I merge a couple of PRs, I would like to not wait 5seconds extra personally. Should we try guarding it by an environment variable so that affected folks can test using that for now?

josephperrott: Done. PTAL

ng-dev/pr/merge/strategies/autosquash-merge.ts
```diff
@@ -77,10 +77,14 @@ export class AutosquashMergeStrategy extends MergeStrategy {
     const localBranch = this.getLocalTargetBranchName(githubTargetBranch);
     /** The SHA of the commit pushed to github which represents closing the PR. */
     const sha = this.git.run(['rev-parse', localBranch]).stdout.trim();
+    // Wait five seconds to account for rate limiting of the token usage during merge.
```
devversion, line 80: Super nit: comment outdated, and ideally we'd have a todo. Non-blocking though

devversion (approved):
Nice! --atomic is cool

PR #1061, opened 2023-03-10
feat(ng-dev): support special action to cut lts minor release

devversion:
Needs for publishing the optimize directive backported to v14, v13 and v12.


josephperrott (approved):
LGTM

PR #1064, opened 2023-03-13
feat(ng-dev/release): properly mark releases as latest on GitHub 

devversion:
See individual commits

### DIFF
--- a/ng-dev/release/publish/actions.ts
+++ b/ng-dev/release/publish/actions.ts
@@ -605,14 +605,14 @@ export abstract class ReleaseAction {
    * @param versionBumpCommitSha Commit that bumped the version. The release tag
    *   will point to this commit.
    * @param isPrerelease Whether the new version is published as a pre-release.
-   * @param shouldShowAsLatestOnGitHub Whether the version released will represent
+   * @param showAsLatestOnGitHub Whether the version released will represent
    *   the "latest" version of the project. I.e. GitHub will show this version as "latest".
    */
   private async _createGithubReleaseForVersion(
     releaseNotes: ReleaseNotes,
     versionBumpCommitSha: string,
     isPrerelease: boolean,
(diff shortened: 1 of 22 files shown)


ng-dev/release/publish/actions.ts
```diff
@@ -635,6 +638,7 @@ export abstract class ReleaseAction {
       name: `v${releaseNotes.version}`,
       tag_name: tagName,
       prerelease: isPrerelease,
+      make_latest: shouldShowAsLatestOnGitHub ? 'true' : 'false',
```
josephperrott, line 641: Woohoo! I didn't know they finally completed my feature request!

ng-dev/release/publish/actions.ts
```diff
@@ -605,11 +605,14 @@ export abstract class ReleaseAction {
...
    *   will point to this commit.
    * @param isPrerelease Whether the new version is published as a pre-release.
+   * @param shouldShowAsLatestOnGitHub Whether the version released will represent
+   *   the "latest" version of the project. I.e. GitHub will show this version as "latest".
    */
   private async _createGithubReleaseForVersion(
     releaseNotes: ReleaseNotes,
     versionBumpCommitSha: string,
     isPrerelease: boolean,
+    shouldShowAsLatestOnGitHub: boolean,
```
josephperrott, line 615:
```suggestion
    useAsLatest: boolean,
```

josephperrott (approved):
LGTM

ng-dev/release/publish/actions.ts
devversion: I had that initially but went with this to make it very clear what it's doing. "latest" itself in this context is quite an ambiguous term.

devversion: I've renamed it to omit the `should`, but other than that- I think this term is quite good

PR #1078, opened 2023-03-20
Begin migration to GHA

josephperrott:
This GHA migration PR is for testing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,3 +29,19 @@ jobs:
         shell: bash
       - run: yarn tsc -p tsconfig.json
         shell: bash
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # tag=v3.3.0
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - uses: ./github-actions/setup-bazel-remote-exec
+      - uses: actions/cache@v3
+        with:
+          path: |
+            .yarn/cache
+          key: v1-${{hashFiles('yarn.lock')}}
+      - run: yarn install --immutable
+        shell: bash
+      - run: yarn bazel test -- //...
+        shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,12 +1,22 @@
 name: CI
 
-on: pull_request_target
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+
+concurrency:
+  group: ci-${{ github.sha }}
+  cancel-in-progress: true
 
 permissions: {}
 
 jobs:
   lint:
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,101 @@
+name: CI
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
```
devversion, line 5: Have we considered having jobs in individual workflow files? I would love that for code readability. Obviously, when actions depend on each other it will become more difficult, but here these are mostly separate/self-contained.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,101 @@
...
+
+permissions: {}
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    if: ${{ contains(fromJSON('["synchronize", "opened", "reopened"]'), github.event.action) }}
```
devversion, line 19: This is not needed at this point

.github/workflows/ci.yml
```diff
@@ -0,0 +1,101 @@
...
+    steps:
+      - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # tag=v3.3.0
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - uses: ./github-actions/setup-bazel-remote-exec
+      - uses: actions/cache@v3
+        with:
+          path: |
+            .yarn/cache
+            .yarn/install-state.gz
```
devversion, line 54: We should extract the caching into a common action, like we do in components: https://github.com/angular/components/blob/main/.github/actions/yarn-install/action.yml

This makes cache key maintenance much more central, easier, and removes bloat in individual job steps

.github/workflows/ci.yml
```diff
@@ -0,0 +1,101 @@
...
+          key: v1-${{hashFiles('yarn.lock')}}
+      - run: yarn install --immutable
+        shell: bash
```
devversion, line 57: Not needed right (and in other places) because bash is the default?

bazel/integration/test_runner/file_system_utils.ts
```diff
@@ -26,7 +26,10 @@ export async function isExecutable(filePath: string): Promise<boolean> {
  * the actual case-exact path for the current platform would be: `C:\Users\<..>`.
  */
 export async function getCaseExactRealpath(filePath: string): Promise<string> {
-  return trueCasePath(filePath);
+  return trueCasePath(filePath).then(
+    (f) => f,
+    () => filePath,
+  );
```
devversion, line 32: Can you elaborate on this?

.github/workflows/ci.yml
josephperrott: Definitely, I wanted to start with leaving things in place during this testing/migration in dev-infra and then extract before we roll wider.

.github/workflows/ci.yml
josephperrott: I think eventually we can go to that, but I want to play with it to understand how to best do this.  I have some ideas with `workflow_calls` that could be a nice option.

PR #1107, opened 2023-04-18
fix(ng-dev): prevent user git config from being used during ng-dev ac…

josephperrott:
…tions

Ignore the local git configuration for the user when running automations via ng-dev


ng-dev/utils/git/git-client.ts
```diff
@@ -107,6 +107,11 @@ export class GitClient {
       // Encoding is always `utf8` and not overridable. This ensures that this method
       // always returns `string` as output instead of buffers.
       encoding: 'utf8',
+      env: {
+        ...process.env,
+        'GIT_CONFIG': '/dev/null',
```
devversion, line 112: This seems risky to me. Can we:

1. Have a comment here, explaining why we do it. The rationale should be in code I think. Is there a link to the environment variable docs? maybe we can also use `GIT_CONFIG_NOSYSTEM`?
2. Do we know if this impacts existing tooling? I'm thinking about cases like on Windows where a user-wide config ensures we always only commit LF line endings- but now that might not be the case anymore, or maybe the commit author is missing? (for the release tool?)

PR #1111, opened 2023-04-21
fix(github-actions): gracefully attempt to delete PR branches

devversion:
Fixes errors like:
https://angular-team.slack.com/archives/C042EU9T5/p1682064243477639

### DIFF
--- a/github-actions/create-pr-for-changes/lib/main.ts
+++ b/github-actions/create-pr-for-changes/lib/main.ts
@@ -284,12 +284,32 @@ async function cleanUpObsoleteBranches(
       (obsoleteBranches.length === 0 ? '-' : obsoleteBranches.join(', ')),
   );
 
+  const failedToDelete: string[] = [];
+  let deletedCount = 0;
+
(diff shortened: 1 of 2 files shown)


alan-agius4 (approved):

github-actions/create-pr-for-changes/lib/main.ts
```diff
@@ -284,12 +284,32 @@ async function cleanUpObsoleteBranches(
       (obsoleteBranches.length === 0 ? '-' : obsoleteBranches.join(', ')),
   );
 
+  const failedToDelete: string[] = [];
+  let deletedCount = 0;
+
   // Delete the obsolete branches.
   for (const branchName of obsoleteBranches) {
```
alan-agius4, line 291: Nit you don't need the `deletedCount`, as you can use `obsoleteBranches.length - failedToDelete.length`

devversion: good point :D we can clean up in the future.

PR #1112, opened 2023-04-21
feat(apps): create the code of conduct frontend application

josephperrott:
Create the initial code of conduct frontend


apps/BUILD.bazel
```diff
@@ -19,8 +20,12 @@ nodejs_binary(
         # Firebase function files
         "//apps/functions:functions_compiled",
         "//apps/functions:functions_files",
+
+        # Firebase hosted application files
+        "//apps/code-of-conduct:application_files",
     ],
     entry_point = "@npm//:node_modules/firebase-tools/lib/bin/firebase.js",
+    tags = ["ibazel_notify_changes"],
```
devversion, line 28: does `copy_to_bin` support ibazel stdin watch mode?

apps/code-of-conduct/app/BUILD.bazel
```diff
@@ -0,0 +1,55 @@
...
+        "//apps/code-of-conduct/app/user-table",
+        "//apps/shared/account",
+        "@npm//@angular/common",
+        "@npm//@angular/core",
+        "@npm//@angular/router",
+    ],
+)
+
+sass_binary(
+    name = "app-style",
```
devversion, line 26:
```suggestion
    name = "app_style",
```

apps/code-of-conduct/app/block-user/block-user.component.html
```diff
@@ -0,0 +1,137 @@
...
+            formControlName="username"
+          />
+          <mat-label>Username</mat-label>
+        </mat-form-field>
+      </span>
+      <span class="block-until">
+        <mat-form-field>
+          <mat-label>Block Until</mat-label>
+          <input
+            [style.display]="blockUserForm.controls.blockUntil.value === false ? 'none': ''"
```
devversion, line 27: Maybe have this as a method in the component as you repeat it later as well

apps/code-of-conduct/app/block-user/block-user.component.html
```diff
@@ -0,0 +1,137 @@
...
+        <textarea
+          formControlName="comments"
+          matInput
+          placeholder="The user broke the code of conduct by..."
+        ></textarea>
+      </mat-form-field>
+    </div>
+    <div class="two-column-row form-item" *ngIf="editMode">
+      <span>
+        <mat-form-field floatLabel="always">
```
devversion, line 86: I thought you set this via a global option

apps/code-of-conduct/app/block-user/block-user.component.scss
```diff
@@ -0,0 +1,51 @@
...
+  }
+}
+
+.comments {
+  mat-form-field {
+    width: 100%;
+  }
+
+  textarea {
+    height: 3.7em; // Allows three lines of text.
```
devversion, line 35: why not `textarea[rows=3]` in the html? the `rows` attribute should do it I think

apps/code-of-conduct/app/block-user/block-user.component.ts
```diff
@@ -0,0 +1,134 @@
...
+    return date;
+  })();
+  year = (() => {
+    const date = new Date();
+    date.setFullYear(date.getFullYear() + 1);
+    return date;
+  })();
+
+  blockUserForm = new FormGroup({
+    username: new FormControl<string>('', {nonNullable: true}),
```
devversion, line 81: would it be more correct to have them nullable instead of having an empty string?

apps/code-of-conduct/app/block-user/block-user.component.ts
```diff
@@ -0,0 +1,134 @@
...
+  });
+
+  async blockUser() {
+    this.blockUserForm.disable();
+    this.dialogRef.disableClose = true;
+
+    this.blockService
+      .block(this.blockUserForm.getRawValue())
+      .then(() => {
+        this.dialogRef.close();
```
devversion, line 96: use async/await. the promise is not awaited while you have `async blockUser`.

devversion: same below for `updateUser`

apps/code-of-conduct/app/block.service.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+export interface BlockedUser extends BlockUserParams {
+  blockedBy: string;
+  blockedOn: Date;
+}
+
+export type BlockedUserFromFirestore = BlockedUser & {
+  blockedOn: number;
+  blockUntil: number | false;
```
devversion, line 35: What does number represent? please add a JSdoc. e.g. unix timestamp

apps/code-of-conduct/app/user-table/user-table.component.html
```diff
@@ -0,0 +1,43 @@
...
+    <mat-header-cell *matHeaderCellDef> Username </mat-header-cell>
+    <mat-cell class="username-cell" *matCellDef="let user">
+      <img src="https://github.com/{{user.get('username')}}.png">
+      <span>{{user.get('username')}}</span>
+    </mat-cell>
+  </ng-container>
+
+  <ng-container matColumnDef="blockUntil">
+    <mat-header-cell *matHeaderCellDef> Blocked Until </mat-header-cell>
+    <mat-cell *matCellDef="let user"> {{user.get('blockUntil') === false ? 'Blocked Indefinitely' : user.get('blockUntil').seconds * 1000 | date}} </mat-cell>
```
devversion, line 12: I see you are doing the *1000 unix conversion here again. Why can't these table rows not also use the `BlockedUser` converter that already does this?

apps/code-of-conduct/esbuild.config.mjs
```diff
@@ -0,0 +1,24 @@
...
+export default {
+  // Note: We support `.mjs` here as this is the extension used by Angular APF packages.
+  resolveExtensions: ['.mjs', '.js'],
+  format: 'esm',
+  plugins: [
+    await createEsbuildAngularOptimizePlugin({
+      enableLinker: {
+        linkerOptions: {
+          // We enable JIT mode as unit tests may need it for overriding.
+          linkerJitMode: true,
```
devversion, line 19: you don't have any tests 😛 so, either we add some for best practice, in  a follow-up? or we should remove this..

apps/code-of-conduct/styles.scss
```diff
@@ -0,0 +1,41 @@
...
+  typography: mat.define-typography-config(),
+  density: -1,
+));
+
+// Include theme styles for core and each component used in your app.
+// Alternatively, you can import and @include the theme mixins for each component
+// that you are using.
+@include mat.all-component-themes($code-of-conduct-theme);
+
+/* You can add global styles to this file, and also import other style files */
```
devversion, line 38: We can remove most of these comments that are just for welcoming new users

apps/firebase.json
```diff
@@ -3,13 +3,49 @@
...
+          "source": "*.[0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f].+(css|js)",
+          "headers": [
+            {
+              "key": "Cache-Control",
+              "value": "public,max-age=31536000,immutable"
+            }
+          ]
+        },
+        {
+          "source": "/@(ngsw-worker.js|ngsw.json)",
```
devversion, line 22: We don't have a worker right?

apps/firebase.json
```diff
@@ -3,13 +3,49 @@
...
     "indexes": "firestore.indexes.json"
   },
+  "hosting": [
+    {
+      "target": "code-of-conduct",
+      "public": "./code-of-conduct",
+      "ignore": ["**/.*"],
+      "headers": [
+        {
+          "source": "*.[0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f].+(css|js)",
```
devversion, line 13: where does this come from? I feel like we could have a simpler pattern

apps/shared/account/account.service.ts
```diff
@@ -25,15 +33,21 @@ export class AccountService {
   /** The display name for the user is available. */
   displayName: string | undefined;
   /** The current accounts github token, if available. */
```
devversion, line 35: Comment needs an update. Also shouldn't it be `null` if we are not authenticated?

apps/shared/account/account.service.ts
```diff
@@ -48,7 +62,9 @@ export class AccountService {
 
   /** Sign the user out of the application. */
   async signOut() {
-    await this.auth.signOut();
+    await this.auth.signOut().then(() => {
+      return this.router.navigate(['login']);
```
devversion, line 66: I don't fully recall the layout- but do we know the route is _always_ called `login`?

apps/firebase.json
josephperrott: This is generated by firebase for hosting an angular app.

apps/shared/account/account.service.ts
josephperrott: I am not sure what you mean here.

apps/shared/account/account.service.ts
josephperrott: We do at this point, I put a todo in to make this better.

apps/shared/account/account.service.ts
devversion: Unless I miss something, the property no longer is the token- but a GitHub octokit API instance. The comment states that the property is the "token".

Also right now you always initialize `Octokit`. Would it make sense to only do it when authenticated?

apps/code-of-conduct/app/block-user/block-user.component.html
```diff
@@ -24,19 +24,19 @@ <h1 class="mat-headline-3" mat-dialog-title>
         <mat-form-field>
           <mat-label>Block Until</mat-label>
           <input
-            [style.display]="blockUserForm.controls.blockUntil.value === false ? 'none': ''"
+            [style.display]="blockUserForm.controls.blockUntil.value !== false"
```
devversion, line 27: Do we miss a value here?

devversion (approved):
LGTM- aside from the unresolved comments

josephperrott:
I am going to merge as is as I will need to do some refactoring to make it testable do to how injection/access works for @angular/fire

PR #1120, opened 2023-04-28
build: update angular dependencies

josephperrott:


devversion (changes requested):
just the lock file?

devversion (approved):
doesn't seem Angular related anymore, but LGTM haha

bazel/map-size-tracking/test/size-golden.json
```diff
@@ -1,57 +1,68 @@
...
+          "tslib/": {
+            "modules/": {
+              "index.js": 519,
+              "size": 519
+            },
+            "size": 6128,
+            "tslib.js": 5609
+          }
+        },
+        "size": 11761,
```
devversion, line 31: this is scary but I don’t think it’s super relevant

PR #1134, opened 2023-05-05
feat(github-actions): provide preview url as an output of the firebase previews action

josephperrott:
Provide the url to the preview server as an output of the deploy-previews action. This url can then be used to extend functionality of the preview server for things such as lighthouse testing.


clydin (approved):

devversion (approved):
I don't know what the plan here is, but exposing makes sense in general first

PR #1173, opened 2023-05-23
refactor: clean-up unnecessary type safety escape

devversion:
We recently added `as any` to some Octokit API call because the types were incorrect. This commit reverts the `as any` cast because the root cause got fixed and we would like to benefit from strong typing.


josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit d7c94a7a1834b8d89cece851aa16945248be5d61.

PR #1174, opened 2023-05-23
refactor: resolve TODO to remove type non-null cast

devversion:
The octokit types for `PullRequest.labels` were incorrect for quite a while, but it looks like this is now resolved. This commit removes the cast as we can rely on actual proper typings now.


josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 673976bafe81c87615f32b82fc5e59c09ea61d40.

PR #1184, opened 2023-05-26
fix(bazel): ensure integration test runner works on windows with short paths

devversion:
Some Windows setups use short paths:
https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#short-vs-long-names

These throw off the true-case path detection, so we fix that by first resolving short paths to the long paths that can then be mapped to the actual true-case paths.


josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit c225f5ff16119fd1c9dfa1e6c9ff169f20737cbf.

PR #1196, opened 2023-06-01
ci: update concurrency group name

josephperrott:
Use the github ref to trigger on the fully formed ref name rather than the sha, which could theoretically appear in two pull requests.


devversion (approved):

josephperrott:
This PR was merged into the repository by commit 620b5241ee5f337f3827f680006b865abbcaecbf.

PR #1223, opened 2023-06-14
build: update ng-renovate lock file

alan-agius4:
This is needed as currently yarn installation fail due to `The lockfile would have been modified by this install, which is explicitly forbidden.`


josephperrott (approved):
LGTM

PR #1228, opened 2023-06-16
feat(bazel): tag benchmark tests for bazel query discoverability

devversion:
This allows us to query for benchmark test targets without having to look for them via rule kind name.


josephperrott (approved):
LGTM

PR #1234, opened 2023-06-22
ci: include bazel output/cache directories in action caching

josephperrott:


.github/workflows/ci.yml
```diff
@@ -83,11 +97,16 @@ jobs:
...
+          node-version-file: '.nvmrc'
+          cache: 'yarn'
       - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # tag=v3.3.1
         with:
           path: |
-            .yarn/cache
-            .yarn/install-state.gz
-          key: v1-${{hashFiles('yarn.lock')}}
+            ~/Library/Caches/bazelisk
+          key: ${{ github.job }}-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
```
devversion, line 108: Bazlisk cache should be guarded by the bazelversion file I think

devversion: Do we know that this is where the bazel repository cache lives? IIRC we set an explicit bazelrc option for CircleCI here

josephperrott: Whoops, yes it should be the bazelversion file.

josephperrott: Added a TODO for the `--repository_cache` flag stuff.  I think its something we will need to put into the github action we use, but want to really consider how we go about it, because this caching might also want to move into it as well.

.github/workflows/ci.yml
```diff
@@ -64,13 +75,19 @@ jobs:
...
           enableCrossOsArchive: true
+          # TODO: setup `build --repository_cache` flag and cache the result
           path: |
-            .yarn/cache
-            .yarn/install-state.gz
-          key: v1-${{hashFiles('yarn.lock')}}
+            %AppData%/bazelisk
+          key: ${{ github.job }}-${{ runner.os }}-${{ hashFiles('**/.bazelversion') }}
+          restore-keys: |
+            ${{ github.job }}-${{ runner.os }}-
```
devversion, line 90: One final nit: I think we can actually remove this as there is no incremental use if e.g. no cache hit was for Bazel 6.0.0- it won't help if we restore 5.0.0 then- it would rather just bloat the cache over time

devversion (approved):

.github/workflows/ci.yml
josephperrott: Good call

josephperrott:
This PR was merged into the repository by commit 8969d1fdbde275edcbba8b006631d422e21d3057.

PR #1238, opened 2023-06-26
ci: set bazel repository cache for bazel remote executions

josephperrott:


circleci-orb/scripts/setup-bazel-remote-exec/index.ts
```diff
@@ -14,6 +14,12 @@ import path from 'path';
 import fs from 'fs';
 import os from 'os';
 
+/** The configuration block added to the bazelrc for configuration of remote execution on CI. */
+const bazelRcAddition = `
+build --config=remote
+build --repository_cache=${process.env['HOME']}/bazel_repository_cache
```
devversion, line 20: I personally don't think this should live in the remote execution action to be honest. We could have a separate composite action that would setup RBE + other things.

devversion: The composite action could then also wire up the cache action and other things

josephperrott: PTAL, see if the refactor is what you were thinking.

github-actions/setup-bazel-remote-exec/set_bazelrc_for_remote_exec.sh
```diff
@@ -0,0 +1,4 @@
+echo """
+build --config=remote
+build --repository_cache=$HOME/bazel_repository_cache
```
devversion, line 3: I honestly was thinking we'd have a totally separate action where we do this. Like `setup_bazel` and then that composite action would call the setup RBE action, and later on also add this to the bazelrc.

This action is scoped to remote execution- and repository caching is unrelated to RBE

josephperrott: Are there cases where we don't want to set up remote execution?

devversion: Well this action sets up remote execution- but repository caching has nothing to do with it

josephperrott: True, what if this action is converted into "setting up bazel for CI" and encompasses both things?

devversion: I'd prefer to keep it separate as an individual building block- especially because it's already non-trivial (and intentionally a little obfuscated)

josephperrott:
@devversion PTAL

github-actions/bazel/setup/add-repository-cache.sh
```diff
@@ -0,0 +1,6 @@
+echo """
+build --repository_cache=~/.cache/bazel_repo_cache
```
devversion, line 2: Nit: We can share this using an environment variable to not repeat it twice.

github-actions/bazel/setup/action.yml
```diff
@@ -0,0 +1,38 @@
...
+  steps:
+    - name: Configure action caching for bazel version downloaded by bazelisk
+      uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # tag=v3.3.1
+      with:
+        path: |
+          ~/.cache/bazelisk
+          \%LocalAppData\%/bazelisk
+        key: bazel-version-${{ hashFiles('**/.bazelversion') }}
+        restore-keys: |
+          bazel-version-
```
devversion, line 23: We should not have a restore key suffix here. If the bazel version changes, it's better to not re-use any cache so that we do not grow indefinitely with more Bazel versions

devversion (approved):
Nicee! this looks great now 🎉

github-actions/bazel/setup/action.yml
josephperrott: Ahh good call, didn't think about that one.

github-actions/bazel/setup/add-repository-cache.sh
josephperrott: Unfortunately for our case here `env` values are not shared between "steps" in composite actions

PR #1239, opened 2023-06-27
Create a utility directory within github-actions and set up first couple utils

josephperrott:
See individual commits.


github-actions/utils/setup-bazel-remote-exec/action.yml
```diff
@@ -3,6 +3,7 @@ description: 'Sets up Bazel remote execution.'
 author: 'Angular'
 inputs:
   bazelrc:
+    default: '.bazelrc'
```
devversion, line 6: Is this right? Setting up the remote bazelrc was made intentionally opt-in because it might be different per setup. I think it was components repo or something.

Also IIRC we used the bazelrc user for this because we don't want to cause Git changes

github-actions/utils/checkout-and-setup-node/action.yml
```diff
@@ -0,0 +1,48 @@
...
+        node-version: ${{ inputs.node-version }}
+        cache: 'yarn'
+
+    - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # tag=v3.3.1
+      if: ${{ inputs.cache-node-modules }} == 'true'
+      with:
+        enableCrossOsArchive: true
+        path: |
+          **/node_modules
+        key: ${{ github.job }}--${{ hashFiles('**/yarn.lock') }}
```
devversion, line 48: Note for discussion: What if the repository will use pnpm when it's using `rules_js`?

github-actions/utils/checkout-and-setup-node/action.yml
```diff
@@ -0,0 +1,48 @@
...
+        node-version-file: ${{ inputs.node-version-file-path }}
+        node-version: ${{ inputs.node-version }}
+        cache: 'yarn'
+
+    - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # tag=v3.3.1
+      if: ${{ inputs.cache-node-modules }} == 'true'
+      with:
+        enableCrossOsArchive: true
+        path: |
+          **/node_modules
```
devversion, line 47: This might cache other things because cache collects all node modules as a post action. Should we limit to the top-level node module? or better IMO, have an input for the "working directory".

github-actions/utils/checkout-and-setup-node/action.yml
```diff
@@ -0,0 +1,48 @@
...
+        fetch-depth: ${{ inputs.fetch-depth }}
+        persist-credentials: false
+
+    - uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # tag=v3.6.0
+      with:
+        node-version-file: ${{ inputs.node-version-file-path }}
+        node-version: ${{ inputs.node-version }}
+        cache: 'yarn'
+
+    - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # tag=v3.3.1
```
devversion, line 42: So, this action manages the caching of node modules, but we don't install them as part of this action right? That might be a little too magical. WDYT?

github-actions/utils/checkout-and-setup-node/action.yml
```diff
@@ -0,0 +1,48 @@
...
+    default: '.nvmrc'
+    description: |
+      Relative path to the nvm version file to set node version, exclusive with
+      node-version-file-path input.  Defaults to .nvmrc
+
+  cache-node-modules:
+    type: boolean
+    description: |
+      Whether to cache the node_modules directory, this is helpful when postinstalls are performed
+      and modify the resulting node_modules directory.  Defaults to False.
```
devversion, line 26:
```suggestion
      and modify the resulting node_modules directory. Defaults to False.
```

github-actions/utils/checkout-and-setup-node/action.yml
```diff
@@ -0,0 +1,48 @@
...
+  node-version:
+    description: |
+      A specific version of node to use for the node environment, exclusive with
+      node-version-file-path input.
+
+  node-version-file-path:
+    default: '.nvmrc'
+    description: |
+      Relative path to the nvm version file to set node version, exclusive with
+      node-version-file-path input.  Defaults to .nvmrc
```
devversion, line 20:
```suggestion
      node-version-file-path input. Defaults to .nvmrc
```

github-actions/utils/checkout-and-setup-node/action.yml
```diff
@@ -0,0 +1,48 @@
...
+
+  cache-node-modules:
+    type: boolean
+    description: |
```
devversion, line 24: Can we specify an explicit default?

.github/workflows/ci.yml
```diff
@@ -52,16 +48,12 @@ jobs:
...
-      - uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # tag=v3.6.0
-        with:
-          node-version-file: '.nvmrc'
-          cache: 'yarn'
+      # Because the checkout and setup node action is contained in the dev-infra repo, we must
+      # checkout the repo to be able to run the action we have created.  Other repos will skip
+      # this step.
+      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # tag=v3.5.3
+      - uses: ./github-actions/utils/checkout-and-setup-node
+      - uses: ./github-actions/utils/setup-bazel-remote-exec
```
devversion, line 56: One question: what's the context for moving these into `utils`? aren't all actions workflow utils?

github-actions/utils/checkout-and-setup-node/action.yml
josephperrott: We can update this action to look at the hash files for `pnpm-lock.yaml` as well

github-actions/utils/checkout-and-setup-node/action.yml
josephperrott: will update to just be ./node_modules.

I would like to hold off on the input until we find we need it though.

github-actions/utils/checkout-and-setup-node/action.yml
josephperrott: Correct we don't install them.  This is just about setting up the environment, not performing the actual installs and builds and such.

github-actions/utils/checkout-and-setup-node/action.yml
devversion: What about AIO?

.github/workflows/ci.yml
josephperrott: They are different as they are the building blocks for the environment and setup for github actions.

Typically you would only use one of the other actions we define here, but these ones are used in conjunction (as preparation) for all of the others.

github-actions/utils/checkout-and-setup-node/action.yml
josephperrott: Oh yeah, that was why I ended up at `**/node_modules`

josephperrott: Added the input.

.github/workflows/ci.yml
devversion: I see what you're seeing, but I consider other actions like for preview deployments similar buildings blocks- so not sure if the separate folder is a useful distinction- but that's really just my pers feeling

josephperrott: I think I am just trying to find a categorization/structure that prevents us from having one directory with like 50 actions thats hard to search/understand.

github-actions/utils/checkout-and-setup-node/action.yml
devversion: I like the action overall, but just want to make sure we understand this. Do you think it's worth doing that in here? I'm imagining a complicated workflow where we'd want to cache other things too, or e.g. bazelisk, or baze repositories? Would we have multiple cache archives? (as we know from Circle- caching is not free).

I'm a little unhappy that we restore cache here, but the package manager needs to be invoked manually.

We can fine-tune afterwards, so LGTM 😄

.github/workflows/ci.yml
devversion: Why would the preview workflow build blocks not be part of this though? should we rather have more categorized folders? e.g.

* github-actions
    * bazel
    * npm
    * previews

devversion (approved):

.github/workflows/ci.yml
josephperrott: We could categorize like this if you would like.  I am not entirely 100% set on the best structure and happy to see where it can best end up as we go

github-actions/utils/checkout-and-setup-node/action.yml
josephperrott: Lets fine tune as we go for sure, but I think that these multiple caches are worthwhile as they are not always managed by the same key.

For instance, caching bazel (bazelisk) and node_modules, are based on different things that don't change at the same time. So it would be better to have them in separate caches. The overhead/speed on Github actions seems much better than Circle

.github/workflows/ci.yml
josephperrott: Updated to use your proposed directories.  We can follow up to categorize and structure/sort the others in follow ups.

devversion (approved):

PR #1258, opened 2023-07-10
build: update octokit dependencies

josephperrott:


ng-dev/utils/git/github.ts
```diff
@@ -31,7 +31,7 @@ export interface GithubRepo {
 /** A Github client for interacting with the Github APIs. */
 export class GithubClient {
   /** The octokit instance actually performing API requests. */
-  private _octokit = new Octokit(this._octokitOptions);
+  protected _octokit = new Octokit({...this._octokitOptions, request: {fetch}});
```
devversion, line 34: Any reason for this?

josephperrott: Starting with the new major of Octokit, you have to provide fetch, or it will attempt to fall back to the native fetch, but this is only available in node 18+.

devversion: Can you add a TODO for node 18 to remove this then?

devversion (approved):

josephperrott:
This PR was merged into the repository by commit 222de0bc054a6d2231243f5ea5ea97991a6a722a.

PR #1266, opened 2023-07-12
feat(github-actions): dynamically set the resources available on the runner during bazel setup

josephperrott:


josephperrott:
@devversion I know we discussed offline about this maybe not being necessary. I investigated and I don't believe that the underlying Bazel issue was resolved yet, and while it theoretically could have the same problem, I have seen a reduction of random windows failures (from workers ending too soon) in my testing.

devversion:
@josephperrott Should we re-open the bug on Bazel's side? also my main point is that the logic for determining the resources via NodeJS may _still not be correct_. i.e. it may surface the same problems as the Bazel CPU/RAM determination.

github-actions/bazel/setup/configure-bazelrc.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/** Path to the provided bazelrc file. */
+const bazelRcPath = process.env['BAZELRC_PATH'];
+
+if (bazelRcPath === undefined) {
+  throw Error('BAZELRC_PATH environment variable was not provided');
+}
+
+/** Total CPU cores available on the system. */
+const totalCpus = cpus().length;
```
devversion, line 12: Are we sure that this is more correct than the CPUs/RAM calculation from Bazel? My understanding is that these methods were returning wrong information because of the docker VM.

PR #1271, opened 2023-07-13
Rename RBE setup Github action and use RBE for windows CI runs

josephperrott:


github-actions/bazel/configure-remote/setup_remote_exec.sh
```diff
@@ -1,7 +1,8 @@
...
+build --config=remote-cache`:u+=`
+build --config=remote`,await r.default.promises.writeFile(e,u,"utf8")}}async function D(e){try{return await r.default.promises.readFile(e,"utf8")}catch{return""}}P(process.env.BAZELRC_PATH).catch(e=>{console.error(e),process.exitCode=1})});M();})();
```
devversion, line 5: Do you re-deploy a new orb version? we should do that:

1. Because sometimes this shell script is too long and doesn't work (especially on Windows shell limits)
2. If we don't do the update now, we end up with weird situations like in the past e.g. where we the service key had to be rotated for security reasons- and the updates ended up being extremely frustrating

devversion: In general, we should not put NodeJS scripts in Shell in github actions because the whole brittle shell mechanism only existed for CircleCI

devversion (approved):
@josephperrott LGTM. Can we please at some point look into updating to these new action names in the primary repos? I don't think we get updates after the renames

github-actions/bazel/configure-remote/setup_remote_exec.sh
josephperrott: I did not deploy a new orb version as I was waiting until it was merged.

I refactored things around to no longer use the a shell script for the github action.

josephperrott:
@devversion  A bit of overhaul, PTAL

devversion (approved):
Nice! we should look into updating the action usages/orb to not incur future complications when we e.g. need to rotate a key./

josephperrott:
This PR was merged into the repository by commit 6afa7d0c9e20968c74e1d3700c2fe76d31d8a72f.

PR #1327, opened 2023-08-17
fix(ng-dev): fix Incomplete string escaping or encoding

### LINKED ISSUE #1326: Fix code scanning alert - Incomplete string escaping or encoding
Tracking issue for:
- [ ] https://github.com/angular/dev-infra/security/code-scanning/33

josephperrott:
Addresses security alert 33

Fixes #1326

### DIFF
--- a/ng-dev/caretaker/check/github.ts
+++ b/ng-dev/caretaker/check/github.ts
@@ -93,7 +93,7 @@ export class GithubQueriesModule extends BaseModule<GithubQueryResults | void> {
         {
           type: 'ISSUE',
           first: MAX_RETURNED_ISSUES,
-          query: `"${repoFilter} ${query.replace(/"/g, '\\"')}"`,
+          query: `"${repoFilter} ${query.replace(/\"/g, '\\"')}"`,
         },
         {...GithubQueryResultFragment},
       );


ng-dev/caretaker/check/github.ts
```diff
@@ -93,7 +93,7 @@ export class GithubQueriesModule extends BaseModule<GithubQueryResults | void> {
         {
           type: 'ISSUE',
           first: MAX_RETURNED_ISSUES,
-          query: `"${repoFilter} ${query.replace(/"/g, '\\"')}"`,
+          query: `"${repoFilter} ${query.replace(/\"/g, '\\"')}"`,
```
devversion, line 96: I think this is incorrect. Maybe something like?

```suggestion
          query: `"${repoFilter} ${query.replace(/([\\"])/g, '\\$1')}"`,
```

josephperrott: I think what you have here will replace `\"` instead of replacing `"` with `\"`

Maybe I am misreading it though?

devversion: I thought the security concern was that we do not escape a `\` directly. Not `\"`

josephperrott: Since we are trying to replace all `"` with `\"` I think its the failure to escape the `"` character

devversion: The error is `This does not escape backslash characters in the input.`. We can just try what you have, but essentially `\"` is a noop and just turns into `"` I think.

devversion (approved):
LGTM if we want to try this, but I'm guessing it's not fixing the security issue

josephperrott:
This PR was merged into the repository by commit 1f04ab9faf5a6d410de843699ac657ae8b3d1540.

PR #1353, opened 2023-08-29
feat(github-actions): improve CI bazel options when setting up bazel

devversion:
* Re-enables some options that were lost from the CircleCI -> GHA migration, like color, verbose failures, and RC announcing

* Attempts to fix Bazel caching by not using tilde, but actually expanding the path into an absolute path. Similar to how it's done in CircleCI.

### DIFF
--- a/github-actions/bazel/setup/action.yml
+++ b/github-actions/bazel/setup/action.yml
@@ -29,8 +29,9 @@ runs:
     - name: Configure action caching for bazel repository cache
       uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # tag=v3.3.1
       with:
-        enableCrossOsArchive: true
+        # Note: Bazel repository cache is located in system locations and cannot use
+        # a shared cache between different runner operating systems.
         path: '~/.cache/bazel_repo_cache'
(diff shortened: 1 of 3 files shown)


github-actions/bazel/setup/setup-ci-bazelrc.sh
```diff
@@ -0,0 +1,17 @@
+echo """
+# Print all the options that apply to the build.
+# This helps us diagnose which options override others
+# (e.g. /etc/bazel.bazelrc vs. tools/bazel.rc)
+build --announce_rc
+
+# Avoids re-downloading NodeJS/browsers all the time.
+build --repository_cache=$HOME/.cache/bazel_repo_cache
```
josephperrott, line 8: This ends up finding the wrong location on windows if I remember correctly

devversion: We were using a slash here as well, so it never really worked I think. Fixed by using a realpath

josephperrott (commented):
Looks like this fails in the node script:

```
Error: ENOENT: no such file or directory, realpath '/home/runner/.cache/bazel_repo_cache'
```

PR #1381, opened 2023-09-12
feat(api-gen): create skeleton for api doc extraction

jelbourn:
This creates two packages for generating API reference documentation: extraction and rendering.

Extraction includes the rule `extract_api_to_json` that uses the Angular compiler to extract API information and write it to a json file.

Rendering consumes the extracted json data, processes it into a renderable form, and renders the final HTML documents.

Some notes: 
* I wasn't sure if the new rules should go in the `bazel/` directory or if they made more sense as a new top-level directory; happy to move them if that's desired. 
* For now the nunjucks template is part of the rule itself, but I might refactor this such that the template can live in another repo for faster iteration (and also making the program easier to run from the command line without bazel)


api-gen/extraction/index.ts
```diff
@@ -0,0 +1,11 @@
+import {writeFileSync} from 'fs';
+
+function main() {
+  const [srcs, outputFileName] = process.argv.slice(2);
```
devversion, line 4: Optional Nit/For a follow-up, or here: I generally like to use the term `outputExecrootRelativePath` or something as this just makes it clear that this a execroot-relative path that is expected to be passed to the action (I know it's long-but just explicit)

api-gen/rendering/BUILD.bazel
```diff
@@ -0,0 +1,38 @@
...
+        "@npm//@types/nunjucks",
+        "@npm//marked",
+        "@npm//nunjucks",
+    ],
+)
+
+# nunjucks templates for generating API reference docs.
+filegroup(
+    name = "templates",
+    srcs = glob(["templates/*.njk"]),
```
devversion, line 23: Just to capture the thoughts here. I know we talked about potentially using something else for templates. Any strong reason on why we use nunjucks? is it for re-using of dgeni templates?

package.bzl
```diff
@@ -32,6 +32,7 @@ NPM_PACKAGE_SUBSTITUTIONS = select({
 BZL_DEFAULTS_ALLOW_PACKAGES = [
     ".github/local-actions",
     "",
+    "api-gen",
```
devversion, line 35: FYI: If we end up shipping the `.bzl` files and `BUILD` files of `/api-gen` (which is not done here anyway), then we actually _cannot_ use the defaults and this line needs to be reverted/removed.

devversion (approved):

josephperrott (approved):
LGTM

package.bzl
jelbourn: Ah, good catch. Done.

jelbourn:
Comments addressed ✅

PR #1387, opened 2023-09-13
feat(api-gen): add more structure to the transform pipeline

jelbourn:
Based on top of #1381.

This commit:
* Expands the processing of raw docs entries to "renderable" entries.
* Adds a nunjucks file loader that lets templates use macros
* Adds some actual templating for classes

### DIFF
--- a/api-gen/rendering/templates/class.njk
+++ b/api-gen/rendering/templates/class.njk
@@ -1,0 +1,6 @@
+{% from "class-member-list.njk" import classMemberList %}
+
+<h1>{{name}}</h1>
+<p>{{htmlDescription | safe}}</p>
+
+{% call classMemberList(members) -%}{%- endcall %}
--- a/api-gen/rendering/entities.ts
+++ b/api-gen/rendering/entities.ts
@@ -21,6 +21,7 @@ export enum EntryType {
   Enum = 'enum',
   Function = 'function',
   Interface = 'interface',
+  NgModule = 'ng_module',
   Pipe = 'pipe',
   TypeAlias = 'type_alias',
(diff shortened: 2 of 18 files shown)


api-gen/rendering/renderables.ts
```diff
@@ -0,0 +1,106 @@
...
+    params: ParameterEntryRenderable[];
+  };
+
+/** Sub-entry for a single function parameter augmented transformed content for rendering. */
+export interface ParameterEntryRenderable extends ParameterEntry {
+  htmlDescription: string;
+}
+
+/** A doc entry that has jsdoc tags. */
+export interface HasJsDocTags {
```
devversion, line 79: Wondering if we should move all these `HasX` interfaces to a separate file and only have the actual renderable entities here?

api-gen/rendering/templates/class.njk
```diff
@@ -0,0 +1,6 @@
+{% from "class-member-list.njk" import classMemberList %}
+
+<h1>{{name}}</h1>
+<p>{{htmlDescription | safe}}</p>
```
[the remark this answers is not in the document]
jelbourn: Changed them all to have a space ✅

api-gen/rendering/renderables.ts
jelbourn: I was on the fence on that, too. I've now split them out to a separate file (`traits.ts`) and then actually moved that plus a few other related files to a new `entities` directory under `rendering`.

jelbourn:
Comments addressed ✅

devversion (approved):

PR #1392, opened 2023-09-15
feat(api-gen): switch from nunjucks to tsx

jelbourn:
Based on top of #1387

This changes rendering from nunjucks to TSX with preact. Benefits:
* Full type safety in templates
* preact + tsx is well-supported and widely used
* TSX is much more familiar for a much wider set of developers

Note that the tsconfig change here is not the most up-to-date recommendation for TSX with preact. For some reason, trying to specify `tsx: 'react-jsx'` doesn't seem to carry through the way it's supposed to and `tsc` always ends up receiving `tsx: 'react'`. This older configuration still works fine, though.

During a chat with @dgp1130, he gave me the idea to use preact as a means to render TSX to static HTML. Another hour or so of tinkering got this working.


api-gen/rendering/BUILD.bazel
```diff
@@ -5,30 +5,27 @@ package(default_visibility = ["//api-gen:__subpackages__"])
 
 ts_library(
     name = "render_api_to_html_lib",
-    srcs = glob(["**/*.ts"]),
+    srcs = glob([
+        "**/*.ts",
+        "**/*.tsx",
```
devversion, line 10: I'm glad `ts_library` supports `.tsx`. That's a lucky coincidence

api-gen/rendering/templates/class-member.tsx
```diff
@@ -0,0 +1,24 @@
+/** @jsx h */
```
devversion, line 1: do we need these with the tsconfig options set? (`jsxFactory`)

api-gen/rendering/templates/parameter.tsx
```diff
@@ -0,0 +1,20 @@
+/** @jsx h */
+import {h} from 'preact';
+import {ParameterEntryRenderable} from '../entities/renderables';
+import {RawHtml} from './raw-html';
+
+
+/** Component to render a function or method parameter reference doc fragment. */
+export function Parameter(props: {param: ParameterEntryRenderable}) {
```
devversion, line 8: I know this will end up being a Preact construct, but do you think an explicit type can be added? like `h.JSX.Element`?

devversion (approved):
This is so much better! 🎉 I was quite unhappy about the loss of type safety and also brought up the comment last time + looked for type-safe templating. Great that you took another stab!

api-gen/rendering/templates/class-member-list.tsx
```diff
@@ -0,0 +1,12 @@
+/** @jsx h */
+import {h} from 'preact';
```
dgp1130, line 2: Just a drive by comment, these imports weren't necessary when I was playing around with it, though I did need an otherwise unused dependency on `//:node_modules/preact`. I suspect these two lines can be dropped with the right configuration?

Example: https://github.com/dgp1130/rules_prerender/blob/85bfe6cf5a2cf42fbed1c842dba5c0cccabbe306/docs/components/header/header.tsx
`tsconfig.json`: https://github.com/dgp1130/rules_prerender/blob/85bfe6cf5a2cf42fbed1c842dba5c0cccabbe306/tsconfig.json#L13-L14

Of course I'm not that familiar with Preact/JSX either, so it's possible I'm doing something wrong.

api-gen/rendering/templates/class-member.tsx
jelbourn: Apparently not, removed.

api-gen/rendering/templates/parameter.tsx
jelbourn: I'd prefer to rely on the inferred type for now since it's not a "public" API; any type mismatches should be caught in the normal composition of components.

api-gen/rendering/templates/class-member-list.tsx
jelbourn: The `/** @jsx h */` commented ended up not being necessary, but the `import` is definitely load-bearing. I think it _wouldn't_ be necessary if the `jsxImportSource` setting worked, but there's some incompatibility there with this older `ts_library` implementation.

jelbourn:
Comments addressed and rebased ✅

api-gen/rendering/templates/parameter.tsx
devversion: Fair enough. Explicit types just speed up type-checking and IDE's so I'm generally a fan of those. But this is a nit

PR #1393, opened 2023-09-15
feat(api-gen): set up Ngtsc program for raw api extraction

jelbourn:
Based on top of #1392

This adds the setup for creating an Ngtsc program to pull API data from files.

Immense thanks to @devversion for helping me with the ESM compatibility workaround 🥇

### DIFF
--- a/api-gen/extraction/BUILD.bazel
+++ b/api-gen/extraction/BUILD.bazel
@@ -44,5 +44,8 @@ nodejs_binary(
     # Do not use the NodeJS linker because:
     #  - it's brittle and causes race conditions on Windows.
     #  - it requires additional work to setup the runtime linker.
-    templated_args = ["--bazel_patch_module_resolver"],
+    templated_args = [
+        "--bazel_patch_module_resolver",
+        "--node_options=--preserve-symlinks-main",
+    ],
 )
--- a/api-gen/extraction/BUILD.bazel
+++ b/api-gen/extraction/BUILD.bazel
@@ -41,11 +41,9 @@ nodejs_binary(
         "@npm//@angular/compiler-cli",
     ],
     entry_point = "bin.mjs",
-    # Do not use the NodeJS linker because:
-    #  - it's brittle and causes race conditions on Windows.
-    #  - it requires additional work to setup the runtime linker.
-    templated_args = [
-        "--bazel_patch_module_resolver",
-        "--node_options=--preserve-symlinks-main",
-    ],
+    # Note: Using the linker here as we need it for ESM. The linker is not
+    # super reliably when running concurrently on Windows- but we have existing
+    # actions using the linker. An alternative would be to:
+    #   - bundle the Angular compiler into a CommonJS bundle
+    #   - use the patched resolution- but also patch the ESM imports (similar to how FW does it).
 )
--- a/api-gen/extraction/BUILD.bazel
+++ b/api-gen/extraction/BUILD.bazel
@@ -1,7 +1,24 @@
 load("@npm//@bazel/concatjs:index.bzl", "ts_library")
 load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_binary")
+load("//bazel/esbuild:index.bzl", "esbuild")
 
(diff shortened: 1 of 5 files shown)


api-gen/extraction/index.ts
```diff
@@ -1,11 +1,22 @@
 import {writeFileSync} from 'fs';
+import ts from 'typescript';
+// @ts-ignore This compiles fine, but Webstorm doesn't like the ESM import in a CJS context.
```
devversion, line 3: FYI: Might be able to get this working by adding a `package.json` with `type: module`. See `ng-dev/package.json`

devversion (approved):

api-gen/extraction/index.ts
jelbourn: Ahh, I had tried adding a `tsconfig.ide.json` here and it worked for compiler-cli, but not for `typescript`. I'll give it a shot.

jelbourn: I tried adding a `package.json` and a `tsconfig.ide.json`, and that did fix the problems I had where it saw `@angular/compiler-cli` and `typescript` as incompatible, but there was a _different_ problem of it not seeing `CompilerOptions` or `CompilerHost` from the compiler-cli entry point.

jelbourn:
@devversion @josephperrott any tips on what might be causing the CI failure here? Everything under `//api-gen` builds successfully for me locally, but `yarn bazel build -- //...` fails locally for me even on `main` unrelated to anything in `api-gen/`.

api-gen/extraction/BUILD.bazel
```diff
@@ -12,16 +27,19 @@ ts_library(
...
 )
 
 # Action binary for the api_gen bazel rule.
 nodejs_binary(
     name = "extract_api_to_json",
     data = [
-        ":extract_api_to_json_lib",
+        ":bin",
+        "@npm//@angular/compiler",
+        "@npm//@angular/compiler-cli",
```
devversion, line 40: typescript needs to be added there

devversion:
I've replied on a line. That should fix it. On RBE bazel is even more strict. The linker from `rules_nodejs` just makes it work locally because it uses non-idiomatic symlinks. The linker still runs here.

jelbourn:
@devversion that didn't seem to have helped, unfortunately. I figured out that I could eliminate the direct import to typescript, though, since compiler-cli also has an API for creating a compiler host. After making that change, though, it's just the same error with `@angular/compiler-cli`

```
ERROR: /home/runner/work/dev-infra/dev-infra/api-gen/extraction/test/BUILD.bazel:3:20: Action api-gen/extraction/test/api.json failed: (Exit 1): extract_api_to_json.sh failed: error executing command 
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@angular/compiler-cli' imported from /b/f/w/bazel-out/k8-opt-exec-58157A25/bin/api-gen/extraction/extract_api_to_json.sh.runfiles/dev-infra/api-gen/extraction/bin.mjs
  (cd /home/runner/.cache/bazel/_bazel_runner/[48](https://github.com/angular/dev-infra/actions/runs/6203361135/job/16843837958?pr=1393#step:7:49)e0b44e0189c0be191a57f5927353cf/execroot/dev-infra && \
    at new NodeError (node:internal/errors:393:5)
  exec env - \
  bazel-out/k8-opt-exec-58157A25/bin/api-gen/extraction/extract_api_to_json.sh api-gen/extraction/test/fake-source.ts bazel-out/k8-fastbuild/bin/api-gen/extraction/test/api.json)
```

`"@npm//@angular/compiler-cli"` is in the deps of every rule involved here, so my thinking is that it's in the bazel sandbox, but maybe something going wrong with module resolution? In either case I'm out of steam for this week

devversion:
@jelbourn Pushed two commits to fix this for you. The latest commit includes a small description of what happened and why it works locally (but not with RBE- where only "specified" files are delivered to the VM)

josephperrott:
This PR was merged into the repository by commit 264811d5dad9272ab8a4ec189f1509151e9a0752.

PR #1394, opened 2023-09-15
fix(github-actions): use the workflow-artifact-name to disambiguate multiple previews across channels

josephperrott:
Allow the workflow artifact name to be used to prevent multiple previews from using the same channel.

### DIFF
--- a/github-actions/previews/upload-artifacts-to-firebase/action.yml
+++ b/github-actions/previews/upload-artifacts-to-firebase/action.yml
@@ -89,7 +89,7 @@ runs:
         expires: 10d
         projectId: '${{inputs.firebase-project-id}}'
         entryPoint: '${{inputs.firebase-config-dir}}'
-        channelId: pr-${{github.repository}}-${{steps.artifact-info.outputs.unsafe-pull-number}}
+        channelId: pr-${{github.repository}}-${{steps.artifact-info.outputs.unsafe-pull-number}}-${{inputs.workflow-artifact-name}}
 
     - uses: marocchino/sticky-pull-request-comment@efaaab3fd41a9c3de579aba759d2552635e590fd # v2
       with:


github-actions/previews/upload-artifacts-to-firebase/action.yml
```diff
@@ -89,7 +89,7 @@ runs:
         expires: 10d
         projectId: '${{inputs.firebase-project-id}}'
         entryPoint: '${{inputs.firebase-config-dir}}'
-        channelId: pr-${{github.repository}}-${{steps.artifact-info.outputs.unsafe-pull-number}}
+        channelId: pr-${{inputs.workflow-artifact-name}}-${{github.repository}}-${{steps.artifact-info.outputs.unsafe-pull-number}}
```
devversion, line 92: iirc there is a length limit and it will truncate the PR from the visible url. so might be worth swapping order

josephperrott: I am not sure I understand what swap you are meaning.  Happy to reorder the elements here, just not sure

devversion (approved):

github-actions/previews/upload-artifacts-to-firebase/action.yml
devversion: Just to capture the thougt:

Firebase truncates the URL that will be given to the user. This may cut off the PR number and make the preview URLs rather meaningless. Hence I propose swapping the PR number with the artifact name so that the PR number is more likely to be included. The full URL is still used internally in Firebase to identify the preview channel (different to the truncated URL derived from the id)

josephperrott:
This PR was merged into the repository by commit 73777547f819464aa5f48071b966eed34fd2c402.

PR #1402, opened 2023-09-20
feat(api-gen): call getApiDocumentation to produce real output

jelbourn:
The latest `next` version of `@angular/compiler-cli` includes the new API, so we can now run it for real. This also adds an `entry_point` to the extraction rule, which is expected by the compiler-cli API.

### DIFF
--- a/api-gen/extraction/index.ts
+++ b/api-gen/extraction/index.ts
@@ -3,17 +3,16 @@ import {writeFileSync} from 'fs';
 import {NgtscProgram, CompilerOptions, createCompilerHost} from '@angular/compiler-cli';
 
 function main() {
-  const [moduleName, srcs, outputFilenameExecRootRelativePath] = process.argv.slice(2);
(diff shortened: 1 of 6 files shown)


api-gen/extraction/index.ts
```diff
@@ -3,17 +3,16 @@ import {writeFileSync} from 'fs';
 import {NgtscProgram, CompilerOptions, createCompilerHost} from '@angular/compiler-cli';
 
 function main() {
-  const [moduleName, srcs, outputFilenameExecRootRelativePath] = process.argv.slice(2);
+  const [moduleName, entryPoint, srcs, outputFilenameExecRootRelativePath] = process.argv.slice(2);
```
devversion, line 6: Maybe in a follow-up?

```suggestion
  const [moduleName, entryPointExecRootRelativePath, srcs, outputFilenameExecRootRelativePath] = process.argv.slice(2);
```

PR #1405, opened 2023-09-21
refactor(api-gen): move api-gen into bazel/ directory

jelbourn:
I discovered the package substitutions that run when building the npm_package. It seemed more appropriate/consistent to add api-gen to the bazel package rather than adding another parallel substitution.

Based on top of #1402


devversion:
LGTM but looks like this PR just copies code, but doesn't move it over?

jelbourn:
Oh whoops, forgot to actually delete the original. Fixed!

josephperrott (approved):
LGTM

PR #1413, opened 2023-09-25
feat(github-actions): use checkout actions filter argument

josephperrott:
Use the filter argument in the checkout actions to allow for all commit messages to be checked out without requiring full depth checkouts of the commits.

### DIFF
--- a/github-actions/npm/checkout-and-setup-node/action.yml
+++ b/github-actions/npm/checkout-and-setup-node/action.yml
@@ -2,12 +2,6 @@ name: 'Checkout and Setup Node'
 description: 'Checks out the repository and sets up node environment.'
 author: 'Angular'
 inputs:
-  fetch-depth:
-    default: 0
-    description: |
-      The number of commits to include in the fetch, defaulting to all commits in the history
-      of the repository
-
   ref:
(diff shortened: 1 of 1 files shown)


github-actions/npm/checkout-and-setup-node/action.yml
```diff
@@ -38,11 +32,11 @@ inputs:
 runs:
   using: composite
   steps:
-    - uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+    - uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
       with:
-        ref: ${{ inputs.ref }}
-        fetch-depth: ${{ inputs.fetch-depth }}
+        filter: tree:0
```
devversion, line 37: do we also want the blobless filter? My understanding is that right now things like `git log` would result in the tree's being fetched- but also the blobs unnecessarily then. Filters can be combined as per git man

josephperrott: On rereading it I think your right, but I am a bit fuzzy still.

PR #1415, opened 2023-09-25
build: update octokit dependencies

josephperrott:


yarn.lock
```diff
@@ -5105,6 +5105,21 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@octokit/core@npm:^5.0.0":
```
devversion, line 5108: Looks like we are ending up with duplicate versions. Do we need to combine these? or do we have some mismatch in versions somewhere?

josephperrott: I am not sure why yarn didn't merge those,  fixed.

josephperrott:
This PR was merged into the repository by commit 0464d2fd8f7d8f677b5fbf048213cdd56e9de414.

PR #1424, opened 2023-09-30
refactor(api-gen): adjust visibility and tsconfig for outside consump…

jelbourn:
…tion

In working on getting these rules working in the framework repo, I found that the rules need:
* To specify `:tsconfig.json` instead of just `:tsconfig` for the packaging substitution to apply
* To make the nodejs_binary targets public.

Changing the `_extraction` target of the macro from private just makes debugging easier.


josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 658f04e73ec1c7f7379672ab3f4f8b916f26d539.

PR #1436, opened 2023-10-05
feat(api-gen): add `import_map` attr

jelbourn:
Adds an `import_map` attribute to extract_api_to_json in order to enable resolving Angular packages, e.g. `@angular/core`. This is necessary for api extraction in order to fully resolve types when they're based on types in other packages.

### DIFF
--- a/bazel/api-gen/extraction/extract_api_to_json.bzl
+++ b/bazel/api-gen/extraction/extract_api_to_json.bzl
@@ -19,6 +19,18 @@ def _extract_api_to_json(ctx):
     json_output = ctx.outputs.output_name
     args.add(json_output.path)
 
+    # Pass the import path map
+    # TODO: consider module_mappings_aspect to deal with path mappings instead of manually
+    # specifying them
+    # https://github.com/bazelbuild/rules_nodejs/blob/5.x/internal/linker/link_node_modules.bzl#L236
+    path_map = {}
+    for target, path in ctx.attr.import_map.items():
+        files = target.files.to_list()
+        if len(files) != 1:
+            fail("Expected a single file in import_map target %s" % target.label)
+        path_map[path] = files[0].path
+    args.add(json.encode(path_map))
+
     # Define an action that runs the nodejs_binary executable. This is
     # the main thing that this rule does.
     run_node(
@@ -54,6 +66,10 @@ extract_api_to_json = rule(
             doc = """Source file entry-point from which to extract public symbols""",
             mandatory = True,
             allow_single_file = True,
+        ),
+        "import_map": attr.label_keyed_string_dict(
+            doc = """Map of import path to the index.ts file for that import""",
+            allow_files = True,
         ),
         "module_name": attr.string(
             doc = """JS Module name to be used for the extracted symbols""",
--- a/bazel/api-gen/extraction/index.ts
+++ b/bazel/api-gen/extraction/index.ts
@@ -1,12 +1,30 @@
 import {writeFileSync} from 'fs';
+import path from 'path';
 // @ts-ignore This compiles fine, but Webstorm doesn't like the ESM import in a CJS context.
 import {NgtscProgram, CompilerOptions, createCompilerHost} from '@angular/compiler-cli';
 
 function main() {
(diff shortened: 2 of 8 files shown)


bazel/api-gen/extraction/index.ts
```diff
@@ -1,12 +1,29 @@
...
+    entryPointExecRootRelativePath,
+    srcs,
+    outputFilenameExecRootRelativePath,
+    serializedPathMapWithExecRootRelativePaths,
+  ] = process.argv.slice(2);
 
-  const compilerOptions: CompilerOptions = {};
+  // The path map is a serialized JSON map of import path to index.ts file.
+  // For example, {'@angular/core': 'path/to/some/index.ts'}
+  const pathMap = JSON.parse(serializedPathMapWithExecRootRelativePaths) as Record<string, string>;
```
devversion, line 17: Note: This is probably fine, but there are limits in how long a shell command can be. That's why we use a Bazel params file most of the time. we can switch to that when we need though.

bazel/api-gen/test/BUILD.bazel
```diff
@@ -4,5 +4,8 @@ generate_api_docs(
     name = "test",
```
devversion, line 4: At some point, we should likely have some actual `test` targets that will ensure this works (also catching issues like the RBE one)

bazel/api-gen/extraction/extract_api_to_json.bzl
```diff
@@ -19,6 +19,12 @@ def _extract_api_to_json(ctx):
     json_output = ctx.outputs.output_name
     args.add(json_output.path)
 
+    # Pass the import path map
+    path_map = {}
+    for target, path in ctx.attr.import_map.items():
+        path_map[path] = target.files.to_list()[0].path
```
devversion, line 25: I recommend checking via `len` if there are more files and throwing then (using `fail`)

bazel/api-gen/extraction/test/BUILD.bazel
```diff
@@ -4,8 +4,14 @@ package(default_visibility = ["//bazel/api-gen:__subpackages__"])
...
 extract_api_to_json(
     name = "test",
-    srcs = ["fake-source.ts"],
+    srcs = [
+        "fake-source.ts",
+        "//bazel/api-gen/extraction/test/dummy-entry-point:dummy_package",
+    ],
     entry_point = "fake-source.ts",
+    import_map = {
+        "//bazel/api-gen/extraction/test/dummy-entry-point:index.ts": "@angular/dummy-package",
```
devversion, line 13: How do we deal with secondary entry-points? e.g. if something imports `@angular/dummy-package/testing`? sounds like we'd need to wire up all individually. Usually we'd have a path mapping like `@angular/bla/*` for such then? do we want to automatically add that? (unless there is an explicit mapping obviously)

devversion (approved):

devversion:
I'm good with this approach but it's a little unfortunate that we would need to maintain multiple places path mappings. Currently the mappings are computed from Bazel graph itself via `ts_library#module_name`.

There is an aspect from `rules_nodejs` that could give us that information nicely from a TS target. Haven't used it in a while, but here it is: https://github.com/bazelbuild/rules_nodejs/blob/5.x/internal/linker/link_node_modules.bzl#L236, and also google/tsec does a similar thing:

I think:

- We should use that in the future. Maybe a TODO or you look into it. I don't suspect it being difficult.
- We should error if we come across module imports that are unresolved. Right now it's too easy to end up with `any`/`unknown` without realizing I think

jelbourn:
WRT the aspect- added a TODO, I'll look into it once the pipeline is more up-and-running 

+1 to erroring if a type is unknown / unresolvable; presumably this would mean actually performing type-checking when we do extraction in compiler-cli, so I'll add that to my list

bazel/api-gen/extraction/test/BUILD.bazel
devversion: The long-term solution is to just derive these based on the aspect. Then we'd have explicit sub-entries for all secondary/tertiary entry-points as well

devversion (approved):

PR #1451, opened 2023-10-18
feat(api-gen): add rule to generate an api manifest

jelbourn:
This commit adds a new rule that ingests extracted API data and produces a manifest of all APIs contained therein.


bazel/api-gen/manifest/index.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/** Manifest that maps each module name to a list of API symbols. */
+type Manifest = Record<string, ManifestEntry[]>;
+
+function main() {
+  const [srcs, outputFilenameExecRootRelativePath] = process.argv.slice(2);
+
+  const sourceContents = srcs
+    .split(',')
+    .map((srcPath) => readFileSync(srcPath, {encoding: 'utf8'}));
```
josephperrott, line 22: You could do this "in parallel" with promises.


```suggestion
  const sourceContents = await Promise.all(srcs
    .split(',')
    .map((srcPath) => readFile(srcPath, {encoding: 'utf8'}));
```

josephperrott (approved):
LGTM

bazel/api-gen/manifest/index.ts
jelbourn: My understanding is that the sync APIs are actually faster because node is single threaded, so the overhead of managing the async operations just adds overhead (though it's been years since I've internalized, not sure if there's new info I don't have)

josephperrott: Okay, works for me

PR #1467, opened 2023-10-24
fix(api-gen): decorators aren't classes

jelbourn:
Removes decorator doc entries from being classified as classes since they don't have members.


josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 4ccf7eed68bba0d25f7edd47997852774f61755d.

PR #1487, opened 2023-10-31
fix(api-gen): support multiple collections for one module in manifest

jelbourn:
Currently the manifest generation assumes each module has exactly one API collection. This change makes it so that a module may be split up into multiple collections, all of which will be unioned together into the manifest. The impetus for this change is adding hand-written collections for Angular element and block APIs.


josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 2af5f1084ebc91654d57eaddecf9fc0136712386.

PR #1497, opened 2023-11-06
refactor(bazel): remove ESM bundling for markdown to HTML

devversion:
follow up fixes to the markdown to html PR.


josephperrott (approved):
LGTM

jelbourn (approved):
LGTM if the relative imports still work

devversion:
They should as this is pure CommonJS and no ESM is involved.

PR #1501, opened 2023-11-06
refactor(api-gen): use real entities from compiler-cli

jelbourn:
Earlier work used a copied version of entities; this updtaes to using the real types from the compiler-cli, as well as removing the `@tsignore` comments for imports to compiler-cli in IDEs with an additional tsconfig and package.json


josephperrott (approved):
LGTM

jelbourn:
I'm confused by these CI failures, trying to figure out what's up

jelbourn:
I can't figure out why module resolution for `@angular/compiler-cli` fails on CI (but works locally). The build rule for manifest generation is effectively identical and it works fine there. I'll close this for now and come back to it as a cleanup for later.

PR #1511, opened 2023-11-09
Markdown to html

josephperrott:


bazel/markdown_to_html/src/index.ts
```diff
@@ -0,0 +1,24 @@
...
+import {marked} from 'marked';
+import {hooks} from './hooks';
+import {readFile} from 'fs/promises';
+
+export async function parseMarkdown(filepath: string): Promise<string> {
+  marked.use({
+    hooks,
+    // Marked will return a promise if the async option is true.
+    // The async option will tell marked to await any walkTokens functions before parsing the tokens and returning an HTML string.
+    // More details: https://marked.js.org/using_pro#async
```
jelbourn, line 18: Expand this comment to call out explicitly why this setting matters for our use of marked?

bazel/markdown_to_html/test/markdown_to_html.spec.ts
```diff
@@ -1,8 +0,0 @@
-import {markdownToHtml} from '../markdown_to_html';
-
-describe('markdown to html', () => {
-  it('should transform standard markdown to html', () => {
-    const markdown = `**important**`;
-    expect(markdownToHtml(markdown)).toContain('<strong>important</strong>');
```
jelbourn, line 6: Could we preserve the separation between index.ts as the bazel program that reads files from disk and the library function that acts on a string? I think this way is easier to test and will make it easier to use the marked transforms from other rules/tools.

jelbourn (approved):

bazel/markdown_to_html/test/markdown_to_html.spec.ts
josephperrott: Yes, I can ensure that is done.

josephperrott:
This PR was merged into the repository by commit 54cfd9eddef68f1d37ce77940e92f9beb0493032.

PR #1513, opened 2023-11-09
feat(bazel/markdown-to-html): support headings in markdown to html

josephperrott:
Properly handle headings for custom styling


bazel/markdown_to_html/src/tranformations/heading.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import {RendererApi} from 'marked';
+
+import {getPageTitle} from '../utils';
+import {getHeaderId} from '../state';
+
+export const headingRender: RendererApi['heading'] = (text, level, raw) => {
+  if (level === 1) {
+    return `
+    <!-- Document Header -->
```
jelbourn, line 17: Is the comment necessary? (admittedly small, but still extra bytes over the wire) 

If it is load bearing in some way, we should have a comment that calls that out 

(similar for other html comments)

bazel/markdown_to_html/src/tranformations/heading.ts
```diff
@@ -0,0 +1,41 @@
...
+  // regex aren't perfect for that but this one should be "good enough"
+  const regex = /<a\s+(?:[^>]*?\s+)?href.*?>(.*?)<\/a>/gi;
+  const anchorLessText = text.replace(regex, '$1');
+
+  const link = getHeaderId(anchorLessText);
+  const label = anchorLessText.replaceAll(/`(.*?)`/g, '<code>$1</code>');
+
+  return `
+  <h${level} id="${link}">
+    <a href="#${link}" class="docs-anchor" tabindex="-1">${label}</a>
```
jelbourn, line 39: I don't think the anchor should be a child of the header. From an a11y standpoint, you would want the anchor to immediately follow the header in navigation order.

jelbourn (approved):
LGTM aside from the anchor nested inside headers

bazel/markdown_to_html/src/tranformations/heading.ts
```diff
@@ -0,0 +1,40 @@
...
+  // regex aren't perfect for that but this one should be "good enough"
+  const regex = /<a\s+(?:[^>]*?\s+)?href.*?>(.*?)<\/a>/gi;
+  const anchorLessText = text.replace(regex, '$1');
+
+  const link = getHeaderId(anchorLessText);
+  const label = anchorLessText.replaceAll(/`(.*?)`/g, '<code>$1</code>');
+
+  return `
+  <h${level} id="${link}">${label}</h${level}>
+  <a href="#${link}" class="docs-anchor" tabindex="-1"></a>
```
jelbourn, line 38: Maybe a TODO, but this anchor would also need either an `aria-label` or visually hidden content like `"Link to ${label}"`

josephperrott:
This PR was merged into the repository by commit 7264838eafd6e989a1410c273e20bd0d90af2e39.

PR #1515, opened 2023-11-10
feat(ng-dev): support enforcing certain checks to be running

devversion:
This can be used in the future to enforce the unified status check, or to enforce additional thing. Right now it can be used to enforce some checks that aren't guaranteed to run otherwise.

In the future we may remove the configuration option and always default to the unified status check. This is the goundwork for future changes.

### DIFF
--- a/.ng-dev/pull-request.mts
+++ b/.ng-dev/pull-request.mts
@@ -3,9 +3,7 @@ import {PullRequestConfig} from '../ng-dev/pr/config/index.js';
 /** Configuration for interacting with pull requests in the repo. */
 export const pullRequest: PullRequestConfig = {
   githubApiMerge: false,
-  mergeReadyLabel: 'action: merge',
-  caretakerNoteLabel: 'merge note',
-  commitMessageFixupLabel: 'needs commit fixup',
+  requiredStatuses: [{name: 'test', type: 'check'}],
 
   // Disable target labeling in the dev-infra repo as we don't have
   // any release trains and version branches.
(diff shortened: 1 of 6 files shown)


.ng-dev/pull-request.mts
```diff
@@ -3,9 +3,7 @@ import {PullRequestConfig} from '../ng-dev/pr/config/index.js';
 /** Configuration for interacting with pull requests in the repo. */
 export const pullRequest: PullRequestConfig = {
   githubApiMerge: false,
-  mergeReadyLabel: 'action: merge',
-  caretakerNoteLabel: 'merge note',
-  commitMessageFixupLabel: 'needs commit fixup',
+  requiredStatuses: [{name: 'test', type: 'check'}],
```
josephperrott, line 6: Should this be `CI / test (pull_request)`?

josephperrott (approved):
LGTM

.ng-dev/pull-request.mts
devversion: Verified with GitHub's API that it would just be `test`.

```graphql
query {
  repository(owner:"angular", name:"dev-infra"){
    pullRequest(number:1515){
      commits(last:1)	{
        nodes{
          commit {
            abbreviatedOid,
            statusCheckRollup{
              contexts(first: 100) {
                nodes {
                  ... on CheckRun {
                    name
                    
                  }
                  ... on StatusContext {
                    context
                  }
           
                }
              }
            }
          }
        }
      }
    }
  }
}
```

josephperrott: weird, thats kind of annoying that its different than the rest of things.

devversion: Haha yeah, there seems to be some special logic for GHA I think.. Wasn't able to find anything in the responses of `CheckRun`

PR #1531, opened 2023-11-15
Fixes

josephperrott:
I know that JSDOM is not the fastest, but I think its a much better/readable experience and from intial testing the cost isn't that bad.  This only occurs within our pipeline so its not a cost users will pay.

### DIFF
--- a/bazel/markdown_to_html/src/extensions/docs-code/docs-code-multifile.ts
+++ b/bazel/markdown_to_html/src/extensions/docs-code/docs-code-multifile.ts
@@ -7,7 +7,7 @@
  */
 
 import {Tokens, Token, TokenizerThis, RendererThis} from 'marked';
-import {buildAttr} from './format/styling';
+import {JSDOM} from 'jsdom';
 
 /** Marked token for a multifile custom docs element. */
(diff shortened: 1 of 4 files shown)


bazel/markdown_to_html/src/extensions/docs-code/docs-code-multifile.ts
```diff
@@ -57,12 +57,15 @@ export const docsCodeMultifileExtension = {
...
-    <div class="docs-code-multifile"${attributes}>
+    const el = JSDOM.fragment(`
+    <div class="docs-code-multifile">
     ${this.parser.parse(token.paneTokens)}
     </div>
-    `;
+    `).firstElementChild!;
+
+    token.path && el.setAttribute('path', token.path);
+    token.preview && el.setAttribute('preview', `${token.preview}`);
```
jelbourn, line 67: I'm not really a fan of using `&&` as a shorthand `if` statement- can we make these actual conditionals? 

(here and elsewhere)

jelbourn (approved):

josephperrott:
This PR was merged into the repository by commit d90597faf10c369773f3fa3e0b8d323e6e9bba68.

PR #1535, opened 2023-11-16
fix(bazel/markdown-to-html): set jsdom as external

josephperrott:
Set jsdom as external since it uses workers and dynamic imports


clydin (approved):

josephperrott:
This PR was merged into the repository by commit ad4e5a46dfd676b6f8225df37c1558e22ea51b4e.

PR #1548, opened 2023-11-27
feat(shared-docs): Create @angular/docs package

josephperrott:
Create initial shared docs package


jelbourn (approved):
LGTM

docs/services/table-of-contents-scroll-spy.service.spec.ts
```diff
@@ -0,0 +1,78 @@
+/*!
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import {TestBed, discardPeriodicTasks, fakeAsync, tick} from '@angular/core/testing';
```
devversion, line 9: are we not running these specs? I worry we'll forget doing this in the future, if we don't do it now?

josephperrott: I knew there was something on my todo list for this branch...

devversion: Thanks! can you also provide additional context on the reason for this in the commit message? would be good to capture this just for the future (and maybe for me :D )

josephperrott: Done.

docs/services/BUILD.bazel
```diff
@@ -0,0 +1,36 @@
...
+    srcs = [
+        "index.ts",
+    ],
+    visibility = ["//docs:__subpackages__"],
+    deps = [
+        ":lib",
+    ],
+)
+
+ts_library(
```
devversion, line 14: The specs in here are not configured.

docs/utils/BUILD.bazel
```diff
@@ -0,0 +1,43 @@
...
+
+ts_library(
+    name = "testing-helper",
+    srcs = [
+        "testing-helper.ts",
+    ],
+    visibility = ["//docs:__subpackages__"],
+)
+
+ts_library(
```
devversion, line 22: The specs here are not configured.

devversion (approved):
LGTM overall, but need to be sure we are not having unused spec files

docs/utils/BUILD.bazel
josephperrott: turns out the spec file that was here wasn't used or a spec file.

josephperrott:
This PR was merged into the repository by commit b09daf3df3011515795db63d7c8206b7ce9ee3b6.

PR #1679, opened 2023-11-29
feat(shared-docs): migrate inject-async service

josephperrott:
Move the inject async service to @angular/shared from angular/angular

### DIFF
--- a/docs/services/inject-async.service.ts
+++ b/docs/services/inject-async.service.ts
@@ -1,0 +1,92 @@
+/*!
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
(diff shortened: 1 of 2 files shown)


docs/services/inject-async.service.ts
```diff
@@ -0,0 +1,94 @@
...
+  Injector,
+  Provider,
+  ProviderToken,
+  Type,
+  createEnvironmentInjector,
+  inject,
+} from '@angular/core';
+
+/**
+ * inject a service asynchronously
```
jelbourn, line 23: optional: it would be nice to have a better comment on what this service actually does

(it seems like it is mostly a convenience for lazy-loading the injection token)

jelbourn (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 1d3d98fabb0f3bb3a18bea96534f2a390a2b09b4.

PR #1680, opened 2023-11-29
fix(ng-dev/release): santize commit messages before generating change logs

josephperrott:
We are seeing instances of including `@for` in a commit message title and Github tries to link to an account.  This change wraps these types of occurrences in a codespan so that it does not occur.


jessicajaniuk (approved):
LGTM!

josephperrott:
This PR was merged into the repository by commit 3b93d927624e63c84b79f412fd708e9bcd2092e0.

PR #1684, opened 2023-12-01
refactor(shared-docs): move markdown guides pipeline to @angular/docs

josephperrott:
Move the markdown guides pipeline to @angular/docs as it is directly docs related.

Additionally, this set up the pipeline to use a compiled script rather than building locally in the depdendant repository.

@devversion Moves away from the markdown_to_html name


package.bzl
```diff
@@ -19,6 +20,7 @@ basePackageSubstitutions = {
     "//shared-scripts/": "@npm//@angular/build-tooling/shared-scripts/",
     "//shared-scripts:": "@npm//@angular/build-tooling/shared-scripts:",
     "//:tsconfig.json": "@npm//@angular/build-tooling:tsconfig.json",
+    "//docs": "@npm//@angular/docs",
```
devversion, line 23: Nit: would probably have two entries (like with `/ bazel`) for correctness

devversion (approved):
LGTM, thank you! Note that I think there might be follow-up PRs because of how differently things run in FW then.

josephperrott:
This PR was merged into the repository by commit 2113cd7f66a089ac0208ea84eee672b2529f4f6c.

PR #1686, opened 2023-12-04
feat(shared-docs): create stackblitz post API pages generator

josephperrott:
A script and associated bazel rule for creating a stackblitz post API html page for examples.

### DIFF
--- a/docs/markdown/_stackblitz.bzl
+++ b/docs/markdown/_stackblitz.bzl
@@ -1,14 +1,16 @@
 load("@build_bazel_rules_nodejs//:providers.bzl", "run_node")
 
 def _generate_stackblitz(ctx):
-    """Implementation of the markdown rule"""
+    """Implementation of the stackblitz generator rule"""
 
     # Determine the stackblitz template base directory
-    stackblitz_template = " " * 1000
-    for file in ctx.files.stackblitz_template:
+    # Create an initial string of length greater than the length the stackblitz template path will
+    # be. This is used so that the first file path we encounter is used as the initial value.
+    template_srcs = " " * 10000
+    for file in ctx.files.template_srcs:
         file_path = file.dirname
-        if (len(file_path) < len(stackblitz_template)):
-            stackblitz_template = file_path
+        if (len(file_path) < len(template_srcs)):
+            template_srcs = file_path
 
     # File declaration of the generated html file
     html_output = ctx.actions.declare_file("%s.html" % ctx.attr.name)
@@ -20,22 +22,22 @@ def _generate_stackblitz(ctx):
     args = ctx.actions.args()
 
     # Path to the example being generated.
-    args.add(ctx.attr.example.label.package)
+    args.add(ctx.attr.example_srcs.label.package)
 
     # Path to the actions temporary directory.
     args.add(tmp_directory.short_path)
 
     # Path to the stackblitz template
-    args.add(stackblitz_template)
+    args.add(template_srcs)
 
     # Path to the html output file to write to.
     args.add(html_output.path)
 
-    ctx.runfiles(files = ctx.files.stackblitz_template)
+    ctx.runfiles(files = ctx.files.template_srcs)
 
     run_node(
         ctx = ctx,
-        inputs = depset(ctx.files.example + ctx.files.stackblitz_template),
+        inputs = depset(ctx.files.example_srcs + ctx.files.template_srcs),
         executable = "_generate_stackblitz",
         outputs = [html_output, tmp_directory],
         arguments = [args],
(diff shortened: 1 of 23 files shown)


docs/markdown/_stackblitz.bzl
```diff
@@ -0,0 +1,68 @@
...
+    return [DefaultInfo(files = depset([html_output]))]
+
+generate_stackblitz = rule(
+    # Point to the starlark function that will execute for this rule.
+    implementation = _generate_stackblitz,
+    doc = """Rule that renders markdown sources to html""",
+
+    # The attributes that can be set to this rule.
+    attrs = {
+        "example": attr.label(
```
jelbourn, line 55: Maybe `example_srcs` ?

jelbourn (approved):
LGTM

docs/markdown/examples/stackblitz/file-system.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import {existsSync} from 'fs';
+import {copyFile, mkdir, readdir, rm, stat} from 'fs/promises';
+import {join} from 'path';
+
+/**
+ * Recurusivelt copy folder and contents to the destigation, creating the destination folder
```
devversion, line 14: Nit: typo

docs/markdown/examples/stackblitz/file-system.ts
```diff
@@ -0,0 +1,57 @@
...
+ * if necessary.
+ **/
```
devversion, line 16: Why can't we use FS copy methods? they seem to support recursive copies?

docs/markdown/examples/stackblitz/builder.ts
```diff
@@ -0,0 +1,177 @@
...
+      form.appendChild(element);
+    }
+  }
+
+  return doc.documentElement.outerHTML;
+}
+
+function createBaseStackblitzHtml(primaryFile: string) {
+  const file = `?file=${primaryFile}`;
+  const action = `https://stackblitz.com/run${file}`;
```
devversion, line 145: could that be just inlined? (a separate variable seems unnecessary)

docs/markdown/_stackblitz.bzl
```diff
@@ -0,0 +1,68 @@
...
+            doc = """Files used for the stackblitz generation.""",
+        ),
+        "stackblitz_template": attr.label(
```
devversion, line 58:
```suggestion
        "template_srcs": attr.label(
```

devversion (approved):

docs/markdown/examples/stackblitz/file-system.ts
devversion: sounds good. created an issue so that we don't forget about this

josephperrott:
This PR was merged into the repository by commit 1337ad7c046566310c1748a4bdf3a12d89207a0d.

docs/markdown/_stackblitz.bzl
```diff
@@ -1,14 +1,16 @@
...
-    """Implementation of the markdown rule"""
+    """Implementation of the stackblitz generator rule"""
 
     # Determine the stackblitz template base directory
-    stackblitz_template = " " * 1000
-    for file in ctx.files.stackblitz_template:
+    # Create an initial string of length greater than the length the stackblitz template path will
+    # be. This is used so that the first file path we encounter is used as the initial value.
+    template_srcs = " " * 10000
+    for file in ctx.files.template_srcs:
```
devversion, line 10: Have we considered specifying a base path, and simply having a default here? It should be "short paths" anyway. I'm thinking in some cases the template _could_ intentionally be nested deeper in the "final result".

docs/markdown/_stackblitz.bzl
```diff
@@ -1,14 +1,16 @@
...
 def _generate_stackblitz(ctx):
-    """Implementation of the markdown rule"""
+    """Implementation of the stackblitz generator rule"""
 
     # Determine the stackblitz template base directory
-    stackblitz_template = " " * 1000
-    for file in ctx.files.stackblitz_template:
+    # Create an initial string of length greater than the length the stackblitz template path will
+    # be. This is used so that the first file path we encounter is used as the initial value.
+    template_srcs = " " * 10000
```
devversion, line 9: Nit: A more robust approach would be to use `None` and check it below as an `or` case. And then have a safety check at the end to make sure we actually saw some sources

devversion (approved):
LGTM. Two minor nits

PR #1690, opened 2023-12-06
feat(shared-docs): move embedded editor to @angular/docs

josephperrott:
Move embedded editor code to shared code location.


docs/editor/code-editor/code-editor.component.scss
```diff
@@ -0,0 +1,185 @@
...
+
+  --code-tags: var(--bright-blue);
+  --code-component: var(--primary-contrast);
+  --code-type-name: var(--vivid-pink);
+  --code-self: var(--orange-red);
+
+  position: relative;
+}
+
+.adev-code-editor-tabs {
```
jelbourn, line 68: I don't think we should do this now, but at some point we're going to want to replace these "adev" prefixes with "docs"

jelbourn (approved):
LGTM, assuming all (most?) of the TS code is copied from blossom

docs/editor/code-editor/code-editor.component.scss
josephperrott: Yeah, agreed, there is a ton of that type of cleanup that needs to happen.

josephperrott:
This PR was merged into the repository by commit 3bdb358b99b17910167b2d663088b4f1d5d0ca09.

PR #1703, opened 2023-12-12
test(shared-docs): `EditorUiState` service tests

leoortizz:
add some tests for the `EditorUiState` service


josephperrott (approved):
LGTM

PR #1713, opened 2023-12-13
refactor: fix config discovery to work with latest prettier

alan-agius4:
Fix config discovery to work with latest prettier. See: https://github.com/prettier/prettier/pull/15363/files

Fix for https://github.com/angular/angular-cli/pull/26571

### DIFF
--- a/ng-dev/format/formatters/prettier.ts
+++ b/ng-dev/format/formatters/prettier.ts
@@ -33,7 +33,10 @@ export class Prettier extends Formatter {
    * to discover it repeatedly for each execution.
    */
   private configPath = this.config['prettier']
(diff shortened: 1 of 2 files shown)


ng-dev/format/formatters/prettier.ts
```diff
@@ -33,7 +33,10 @@ export class Prettier extends Formatter {
    * to discover it repeatedly for each execution.
    */
   private configPath = this.config['prettier']
```
alan-agius4, line 35: I wonder if we should prettier just find the right config on it's own.

josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 5613c1d64fcb093c6a07cf920c0cde88fcb46cd0.

PR #1725, opened 2023-12-20
fix(shared-docs): use correct types

josephperrott:
Use the types from interfaces instead of any.


josephperrott:
This PR was merged into the repository by commit f2dbdeb62601fa259e10be581758fc147eb75094.

PR #1726, opened 2023-12-20
fix(ng-dev): don't require a branch confirmation if no targets are used

josephperrott:
If there is no target labelling then no branch prompt is necessary


clydin (approved):

josephperrott:
This PR was merged into the repository by commit d01e0c2916c4ba253852fc12819c413f382cde73.

PR #1743, opened 2024-01-10
fix(bazel): api goldens resolving node modules and resulting in differences

devversion:
Currently, the API golden generation may accidentally pick up user node modules with `bazel run` and end up causing different output than what's expected with `bazel test.`


josephperrott (approved):
LGTM

PR #1771, opened 2024-01-31
refactor(shared-docs): use `docs-` prefix instead of `adev-` for class names

josephperrott:
Use the generic `docs-` rather than `adev-` as we aren't always using this in adev.


jelbourn (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 5254150825dd7a7bf8a86cf5aeccd31a8076dee5.

PR #1781, opened 2024-02-09
feat(shared-docs): process mermaid code blocks into styled SVG images

josephperrott:
Include processing of mermaid code blocks in markdown generation.


docs/markdown/guides/mermaid/index.ts
```diff
@@ -0,0 +1,62 @@
...
+  const require = createRequire(import.meta.url);
+  return (mermaidScriptTagData = {path: require.resolve('mermaid/dist/mermaid.js')});
+}
+
+/** Replace the code block content with the mermaid generated SVG element string in place. */
+export async function processMermaidCodeBlock(token: DocsCodeToken) {
+  /** Browser instance to run mermaid within. */
+  const browser = await chromium.launch({
+    headless: true,
+    executablePath: join(runfiles.resolveWorkspaceRelative('.'), process.env['CHROME_BIN']!),
```
devversion, line 36: We should resolve the `CHROME_BIN` directly and not rely on the workspace execroot assumption

docs/markdown/guides/mermaid/index.ts
```diff
@@ -0,0 +1,62 @@
...
+/** The full path to the mermaid script. */
+let mermaidScriptTagData: {path: string} | undefined;
+
+/** Get the mermaid script file path, resolving it if necessary first. */
+function getMermaidScriptTagData() {
+  if (mermaidScriptTagData) {
+    return mermaidScriptTagData;
+  }
+  const require = createRequire(import.meta.url);
+  return (mermaidScriptTagData = {path: require.resolve('mermaid/dist/mermaid.js')});
```
devversion, line 28: Ideally resolve via `runfiles` as well as this is a CJS/ESM interop we shouldn't be introducing I think

docs/markdown/guides/mermaid/index.ts
```diff
@@ -0,0 +1,62 @@
...
+/** Replace the code block content with the mermaid generated SVG element string in place. */
+export async function processMermaidCodeBlock(token: DocsCodeToken) {
+  /** Browser instance to run mermaid within. */
+  const browser = await chromium.launch({
+    headless: true,
+    executablePath: join(runfiles.resolveWorkspaceRelative('.'), process.env['CHROME_BIN']!),
+    args: ['--no-sandbox'],
+  });
+  /** Page to run mermaid in. */
+  const page = await browser.newPage({bypassCSP: true});
```
devversion, line 40: Nit: Add a comment on why `bypassCSP`

docs/markdown/guides/mermaid/index.ts
```diff
@@ -0,0 +1,62 @@
...
+import {createRequire} from 'node:module';
+import {Mermaid, MermaidConfig} from 'mermaid';
+import {runfiles} from '@bazel/runfiles';
+import {join} from 'node:path';
+import {mermaidStyling} from './styling';
+
+// Declare mermarid in the context of this file so that typescript doesn't get upset when we
+// access it within the `page.evaluate` function. At runtime the context in with the method
+// is run difference than this file, but this makes typescript happy.
+declare const mermaid: Mermaid;
```
devversion, line 12: 👍

docs/markdown/BUILD.bazel
```diff
@@ -99,9 +103,18 @@ nodejs_binary(
...
     name = "markdown",
     data = [
+        "//bazel/browsers/chromium:chromium-for-generation",
         "@npm//jsdom",
+        "@npm//mermaid",
+        "@npm//playwright-core",
     ],
     entry_point = "//docs/markdown:guides.mjs",
+    env = {
+        "CHROME_BIN": "$(CHROMIUM)",
```
devversion, line 113: This has implications I think because: Chromium on Darwin has spaces in its files. Remember all the reports in Slack where macOS team memebers get errors because runfiles in bazel don't support spaces.

docs/markdown/BUILD.bazel
```diff
@@ -99,9 +103,18 @@ nodejs_binary(
...
+    },
+    toolchains = [
+        "//bazel/browsers/chromium:toolchain_alias",
```
devversion, line 116: Are there ways to have sub-targets for guides with mermaid to only fetch the browsers when needed?

docs/markdown/guides/testing/BUILD.bazel
```diff
@@ -20,12 +20,16 @@ jasmine_node_test(
...
     data = [
         "@npm//jsdom",
+        "@npm//mermaid",
+        "@npm//playwright-core",
     ] + glob([
         "**/*.md",
         "**/*.svg",
         "**/*.ts",
     ]),
-    # JSDOM should not be bundled because it has workers and dynamic imports.
```
devversion, line 28: Nit: Would be good to keep that comment and add one for playwright.

docs/markdown/guides/mermaid/index.ts
```diff
@@ -0,0 +1,62 @@
+import {DocsCodeToken} from '../extensions/docs-code/docs-code';
+import {chromium} from 'playwright-core';
```
devversion, line 2: Why aren't we using Pupeteer?

docs/markdown/guides/mermaid/index.ts
```diff
@@ -0,0 +1,62 @@
...
+    return mermaidScriptTagData;
+  }
+  const require = createRequire(import.meta.url);
+  return (mermaidScriptTagData = {path: require.resolve('mermaid/dist/mermaid.js')});
+}
+
+/** Replace the code block content with the mermaid generated SVG element string in place. */
+export async function processMermaidCodeBlock(token: DocsCodeToken) {
+  /** Browser instance to run mermaid within. */
+  const browser = await chromium.launch({
```
jelbourn, line 34: How much overhead is there in launching the browser per diagram? Would it be valuable to keep the browser running until the end of this process in case the current run has multiple diagrams?

docs/markdown/guides/mermaid/styling.ts
```diff
@@ -0,0 +1,61 @@
+import {type MermaidConfig} from 'mermaid';
+
+/** Styling related mermaid configuration to use when generating mermaid charts. */
+export const mermaidStyling: MermaidConfig = {
+  themeCSS: `
+  background-color: var(--page-background) !important; // svg background color
```
jelbourn, line 6: Does this CSS strictly need to be included as part of the pipeline, or could it just be included directly (perhaps as a Sass mixin) into the docs site? 

If not, could this live in a `.css` file that we import into this file?

docs/markdown/guides/mermaid/index.ts
josephperrott: I removed it, I was running into some issues at one point I think because of injecting the script tag after initial load, but it seems to work now.

docs/markdown/guides/mermaid/index.ts
josephperrott: I couldn't get puppeteer to work, I could not figure out why

docs/markdown/guides/mermaid/index.ts
josephperrott: I don't believe its that much overhead, and marked doesn't have really comprehensive hooks to use to determine when we can close the browser as far as I know.  I can put a TODO to investigate and revisit it later.

docs/markdown/guides/mermaid/styling.ts
josephperrott: It likely could, I just ported this from the previous version.

josephperrott: Done

docs/markdown/guides/mermaid/index.ts
josephperrott: `CHROME_BIN` comes in as `../org_chromium_chromium_linux_x64/chrome-linux/chrome` and doesn't resolve correctly in both test and build environments.  I think we end up having to do this because we are not in the run context, but the build context and bazel lays files out differently.

docs/markdown/BUILD.bazel
josephperrott: Should I switch to using firefox for it you think?

docs/markdown/BUILD.bazel
josephperrott: I suppose we could have a flag on a macro so that we note if it uses mermaid, but I am not sure if thats worth doing, thoughts?

docs/markdown/guides/mermaid/index.ts
devversion: We should try to get it work as we are introducing another dependency for this and we should consistent automation, if we already have webdriver + pupeteer in all repos

devversion: I'm happy to help with this

docs/markdown/BUILD.bazel
devversion: If possible, would be nice. Browsers have issues the windows spaces, so that would be good, and it does require fetching the downloads

docs/markdown/BUILD.bazel
devversion: Mhmm. Interesting idea. Might be worth doing yeah 😄

docs/markdown/guides/mermaid/index.ts
devversion: Interesting.  That is a Bazel short path and _should_ work in practice. Using it directly as a execroot path is reasonable though. I think this is what you do? please add a comment 😄

docs/markdown/BUILD.bazel
josephperrott: Done.

docs/markdown/BUILD.bazel
josephperrott: So firefox we don't have windows support.  So I am not sure what our options really are :D

docs/markdown/guides/mermaid/index.ts
josephperrott: I think that for now I would like to unblock the work within adev on angular/angular that is blocked on getting this support in place.

We can replace this at any point as its happening under the hood anyway. As I tried and investigated it seemed mostly to be issues around the fact that its trying to create a bunch of tmp files/directories in read only workspaces.

devversion: I did run it with Bazel in the past to perform screenshot diffing.

Can you create an issue for tracking? We should definitely revisit what we use for automation, so that we use the same thing consistently

josephperrott: Done https://github.com/angular/dev-infra/issues/1784

docs/markdown/BUILD.bazel
josephperrott: In my testing I did not have any issues on os x

devversion: Okay, we might need to revisit this though if comes up. See all the reports for integration tests in #fw-caretaker

docs/markdown/guides/parse.ts
```diff
@@ -48,6 +55,7 @@ export async function parseMarkdown(
       docsCodeExtension,
       docsCodeMultifileExtension,
     ],
+    walkTokens: (HANDLE_MERMAID && (await import('./handle-tokens')).handleToken) || undefined,
```
devversion, line 58: This should be in the `./handle-tokens` `mermaid` `if` right?

docs/markdown/guides/testing/bootstrap.init.ts
```diff
@@ -0,0 +1,2 @@
+// Set HANDLE_MERMAID to false to make it already available at test time.
```
devversion, line 1: Is this what you are saying?

```suggestion
// Set HANDLE_MERMAID to false as it's not needed for unit testing
```

devversion (approved):

josephperrott:
This PR was merged into the repository by commit 24bfb96956c61fe6e58c18001214acfe612704cd.

PR #1787, opened 2024-02-15
fix(shared-docs): prevent issues caused by reuse of the token in marked

josephperrott:
Ensure the proper diagram source is used instead of just assuming token.code is still the expected value


clydin (approved):

josephperrott:
This PR was merged into the repository by commit 77b43652a547e737b401c132dc76086dae9c2855.

PR #1788, opened 2024-02-15
feat(api-gen): process CLI entries as well

josephperrott:
Process the CLI entries when provided to the api-gen tooling.

### DIFF
--- a/bazel/api-gen/rendering/entities/categorization.ts
+++ b/bazel/api-gen/rendering/entities/categorization.ts
@@ -1,3 +1,4 @@
+import {CliCommand} from '../cli-entities';
 import {
   ClassEntry,
   ConstantEntry,
@@ -22,6 +23,7 @@ import {
   MemberEntryRenderable,
   MethodEntryRenderable,
   TypeAliasEntryRenderable,
+  CliCommandRenderable,
 } from './renderables';
 import {HasJsDocTags} from './traits';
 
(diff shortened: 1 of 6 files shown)


bazel/api-gen/rendering/entities/categorization.ts
```diff
@@ -102,3 +104,9 @@ export function isSetterEntry(entry: MemberEntry): entry is PropertyEntry {
 export function isDeprecatedEntry<T extends HasJsDocTags>(entry: T) {
   return entry.jsdocTags.some((tag) => tag.name === 'deprecated');
 }
+
+/** Gets whether the given entry is a cli entry. */
+export function isCliEntry(entry: unknown): entry is CliCommandRenderable;
```
jelbourn, line 109: Is there a way we could make the extracted type for a CLI command a `DocEntry` and the base class for the renderable fit into `DocsEntryRenderable`, or are they incompatibly different?

bazel/api-gen/rendering/index.ts
```diff
@@ -15,12 +17,24 @@ interface EntryCollection {
 function parseEntryData(srcs: string[]): EntryCollection[] {
   return srcs.map((jsonDataFilePath) => {
     const fileContent = readFileSync(jsonDataFilePath, {encoding: 'utf8'});
-    return JSON.parse(fileContent) as EntryCollection;
+    const fileContentJson = JSON.parse(fileContent) as unknown;
```
jelbourn, line 20: IS the unknown cast here necessary vs casting straight to `EntryCollection`?

josephperrott: `unknown` is technically correct since we don't know what it is.

I could have it be `EntryCollection | DocEntry` if you would prefer

bazel/api-gen/rendering/entities/categorization.ts
josephperrott: Yeah, the only overlap is `name` and we don't directly explicitly control the CLI json structure since its generated by tooling.

jelbourn (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 74bcca51a7577882740c024e87391165864e01fb.

PR #1791, opened 2024-02-20
fix(shared-docs): properly stringify the visible files from regions for parsing by the element

josephperrott:
Properly stringify with a wrapping `[]`, as expected by the docs-code component element.


JeanMeche (approved):
LGTM 
Btw we'll also need a fix for `visibleRegion`

PR #1796, opened 2024-02-21
docs: remove margin-top negative style

ferrorenan:
**What is the current behavior?**
Currently the **docs-steps** component has the margin-top property with a negative value, which causes some layout breaks, as you can see in the report in this issue: https://github.com/angular/angular/issues/54476

**What is the new behavior?**
Remove the negative margin-top property to avoid breaks and leave the layout with its normal behavior.

### DIFF
--- a/docs/styles/docs/_steps.scss
+++ b/docs/styles/docs/_steps.scss
@@ -5,7 +5,6 @@
 @mixin docs-steps() {
   .docs-steps {
     --gutter: 4rem;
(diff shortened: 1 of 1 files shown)


josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 9a83b3589ffc26a9618a6ed0352221b1d3858719.

PR #1805, opened 2024-02-22
feat: add cookie consent gtag event

bencodezen:
This PR adds the code to grant permission for cookies.

The other half of this PR is https://github.com/angular/angular/pull/54574

### DIFF
--- a/docs/components/cookie-popup/cookie-popup.component.ts
+++ b/docs/components/cookie-popup/cookie-popup.component.ts
@@ -11,6 +11,35 @@ import {NgIf} from '@angular/common';
 import {LOCAL_STORAGE} from '../../providers/index';
 
 export const STORAGE_KEY = 'docs-accepts-cookies';
+export function setCookieConsent(state: 'denied' | 'granted'): void {
+  try {
+    if (window.gtag) {
+      const consentOptions = {
+        ad_user_data: state,
+        ad_personalization: state,
+        ad_storage: state,
+        analytics_storage: state,
+      };
+
+      if (state === 'denied') {
+        window.gtag('consent', 'default', {
+          ...consentOptions,
+          wait_for_update: 500,
+        });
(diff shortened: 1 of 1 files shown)


josephperrott (approved):
LGTM

AndrewKushnir:
Please do not merge this PR yet, there might be a bug here, see https://github.com/angular/angular/pull/54574#pullrequestreview-1899129041.

bencodezen:
Review has finished on https://github.com/angular/angular/pull/54574#pullrequestreview-1899129041, we should be good to go now 👍

AndrewKushnir:
@bencodezen quick question (from https://github.com/angular/angular/pull/54574): it looks like an update happens in the accept action, which would work for the first time a user clicks the button, but after a page refresh, it looks like it would not be called again, so we end up in the "denied" state. Is that correct?

bencodezen:
> @bencodezen quick question (from [angular/angular#54574](https://github.com/angular/angular/pull/54574)): it looks like an update happens in the accept action, which would work for the first time a user clicks the button, but after a page refresh, it looks like it would not be called again, so we end up in the "denied" state. Is that correct?
> 
> (removed the "merge" label for now, please re-add once this question is resolved)

@AndrewKushnir I can't believe I missed that. 🤦‍♂️ Yes you are absolutely correct. I've updated the code so that it checks for the localStorage state before setting the denied state. 

In addition, I've gone ahead and refactored the code to centralize the functionality which should make maintenance easier in the future. Appreciate your review on this! 🙇‍♂️

bencodezen:
@josephperrott It looks like we're good on this! Just need your review and approval

josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 74707b165e4b6eca0b033c37191280b12539a26f.

PR #1816, opened 2024-02-27
fix(shared-docs): wrap heading around anchor element to enable clicka…

JeanMeche:
…ble headings


josephperrott (changes requested):
Looks like some tests are affected as well.

josephperrott (approved):
LGTM

PR #1817, opened 2024-02-27
fix(github-actions): update the mergeability state on the pull reques…

josephperrott:
…t earlier

Update the mergeability state on the pull request prior to sending the workflow init request

### DIFF
--- a/github-actions/branch-manager/lib/main.ts
+++ b/github-actions/branch-manager/lib/main.ts
@@ -49,7 +49,11 @@ async function run() {
     // Invoke sequentially to avoid github secondary rate limits
     // https://docs.github.com/en/rest/guides/best-practices-for-integrators?apiVersion=2022-11-28#dealing-with-secondary-rate-limits
     for (const pr of prs) {
-      await createWorkflowForPullRequest({pr});
+      await createWorkflowForPullRequest({
+        repo: context.issue.repo,
+        owner: context.issue.owner,
+        pr,
+      });
     }
   }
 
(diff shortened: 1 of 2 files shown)


github-actions/branch-manager/lib/main.ts
```diff
@@ -61,36 +65,58 @@ async function run() {
       const hasMergeLabel = payload.pull_request.labels.some(
         ({name}) => name === actionLabels.ACTION_MERGE.name,
       );
-      if (hasMergeLabel) {
-        await createWorkflowForPullRequest();
+      if (!hasMergeLabel) {
```
devversion, line 68: Should it update in both states then? 

It's also worth mentioning that the state is easily getting out of sync. E.g. consider:

1. PR opened -> I get an update on mergability. Great!
2. Pushes happening
3. My PR is not updated with mergabilty because it has no merge ready label!!

Instead, an option is to consistently check all PRs, even for push, that are more recent than e.g. last 3 months

josephperrott: Switched it back to how it was, it was an errant change.

github-actions/branch-manager/lib/main.ts
```diff
@@ -62,35 +66,57 @@ async function run() {
...
+        });
       }
     }
 
     if (context.payload.action === 'labeled') {
-      const event = context.payload as PullRequestLabeledEvent;
-      // If the merge label has been added, or if target labels have changed,
-      // another update of the merge status is needed.
+      const payload = context.payload as PullRequestLabeledEvent;
+      // If the merge label has been added, or if target labels have changed // another update of
```
devversion, line 80: Nit: formatting seems broken here

github-actions/branch-manager/lib/main.ts
```diff
@@ -62,35 +66,57 @@ async function run() {
...
+async function createWorkflowForPullRequest(inputs: WorkflowInputs) {
+  if (inputs.sha !== undefined) {
+    console.info(`Setting pending status for pr #${inputs.pr}`);
+    await github().then((api) =>
+      api.repos.createCommitStatus({
+        repo: inputs.repo,
+        owner: inputs.owner,
+        state: 'pending',
+        description: 'Running mergibility check',
+        sha: inputs.sha!,
```
devversion, line 114: Is this needed?

devversion (approved):

github-actions/branch-manager/lib/main.ts
josephperrott: Unfortunately yes, I think since the `inputs` is used in the closure, the `!== undefined` check that happens outside of the closure does not come into the typings in the closure.

I refactored to avoid the closure, but its definitely an odd case I hadn't run into before.

josephperrott:
This PR was merged into the repository by commit 662ce0169556b6e6cd69bbe6e35f981476d624cc.

PR #1828, opened 2024-03-05
fix(ng-dev): Fixes graphql query for pr files

jessicajaniuk:
This fixes the improperly formed graphql query to retrieve PR files and paginate through them

### DIFF
--- a/ng-dev/pr/common/fetch-pull-request.ts
+++ b/ng-dev/pr/common/fetch-pull-request.ts
@@ -134,13 +134,12 @@ export const PR_SCHEMA = {
 
 export type PullRequestFromGithub = typeof PR_SCHEMA;
 
-export const PR_FILES_SCHEMA = {
-  nodes: [
-    {
-      path: graphqlTypes.string,
-    },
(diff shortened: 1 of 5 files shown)


ng-dev/pr/common/fetch-pull-request.ts
```diff
@@ -134,13 +134,16 @@ export const PR_SCHEMA = {
...
 export type PullRequestFromGithub = typeof PR_SCHEMA;
 
-export const PR_FILES_SCHEMA = {
-  nodes: [
-    {
-      path: graphqlTypes.string,
-    },
-  ],
-};
+export const PR_FILES_SCHEMA = params(
```
devversion, line 137: is this needed with params already defined below?

devversion (approved):

ng-dev/pr/common/fetch-pull-request.ts
jessicajaniuk: Yes, this is passing in the initial params for the query used for the pagination call.

PR #1850, opened 2024-03-14
refactor(docs): use utils pattern for cookie consent function

bencodezen:
It looks like the export function pattern I was using is not being picked up by BAZEL and thus not available in the build when being referenced in angular/angular.

So I used the utils pattern currently in use for other functions and believe this should expose the function correctly to the package.

Let me know if I'm mistaken though!

### DIFF
--- a/docs/components/cookie-popup/BUILD.bazel
+++ b/docs/components/cookie-popup/BUILD.bazel
@@ -15,6 +15,7 @@ ng_module(
     ],
     deps = [
         "//docs/providers",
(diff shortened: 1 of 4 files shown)


josephperrott (approved):
LGTM

josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit ab4d7a403afabde6787588e0645c7c142a0d812f.

PR #1851, opened 2024-03-14
refactor(ng-dev): use the combined checks and statuses function from the github macros for publishing

josephperrott:
Use the combined check and status function from github macros for publishing checks instead of just checking statuses.


devversion (approved):

josephperrott:
This PR was merged into the repository by commit 602dd0a29dc13b234ebe0c599d2d407678688a5c.

PR #1854, opened 2024-03-15
feat(bazel): support api generation for initializer APIs

devversion:
Adds support for API generation of initializer APIs like the new `input`, `model`, `output` etc.

This commit:

* Updates `entities.ts` to match upcoming changes in the compiler to identify initializer APIs. Unfortunately cannot use the direct import yet- due to a circular dep.
* Adds a new renderable entry for initializer API
* Implements the rendering of the initializer API
* Fixes an issue where the `isDeveloperPreview` detection accidentally worked— but in practice is a field only for renderable entries.
* Syncs the `entities.ts` for other fields like `isNewType`

image

### DIFF
--- a/bazel/api-gen/rendering/entities.ts
+++ b/bazel/api-gen/rendering/entities.ts
@@ -20,11 +20,13 @@ export enum EntryType {
   Element = 'element',
   Enum = 'enum',
   Function = 'function',
+  FunctionWithOverloads = 'function_with_overloads',
   Interface = 'interface',
   NgModule = 'ng_module',
   Pipe = 'pipe',
   TypeAlias = 'type_alias',
   UndecoratedClass = 'undecorated_class',
+  InitializerApiFunction = 'initializer_api_function',
 }
 
 /** Types of class members */
@@ -74,8 +76,6 @@ export interface DocEntry {
   description: string;
   rawComment: string;
   jsdocTags: JsDocTagEntry[];
-  isDeprecated: boolean;
-  isDeveloperPreview: boolean;
 }
 
 /** Documentation entity for a constant. */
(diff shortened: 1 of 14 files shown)


bazel/api-gen/rendering/entities.ts
```diff
@@ -162,3 +163,26 @@ export interface ParameterEntry {
...
+  name: string;
+  signatures: FunctionEntry[];
+  implementation: FunctionEntry | null;
+}
+
+export interface InitializerApiFunctionEntry extends DocEntry {
+  callFunction: FunctionWithOverloads;
+  subFunctions: FunctionWithOverloads[];
+
+  __adevMetadata__?: {
```
josephperrott, line 177: can this be `__docsMetadata__` so that its not tied directly to adev?

bazel/api-gen/rendering/transforms/jsdoc-transforms.ts
```diff
@@ -143,9 +148,11 @@ function getHtmlAdditionalLinks<T extends HasJsDocTags & HasModuleName>(
   return seeAlsoLinks;
 }
 
-/** Some descriptions in the text contain HTML elements like `input` or `img`,
+/**
+ * Some descriptions in the text contain HTML elements like `input` or `img`,
```
josephperrott, line 152: Nit: indentation

```suggestion
 *  Some descriptions in the text contain HTML elements like `input` or `img`,
```

josephperrott (approved):
LGTM

bazel/api-gen/rendering/entities.ts
devversion: good idea. I'm good with this

PR #1944, opened 2024-03-27
refactor(api-gen): remove link mapper

JeanMeche:
Now that AIO doesn't have to be supported anymore, we can remove the link mapper.

Pending: angular/angular#55043


twerske (approved):
LGTM - One note, we might need the link mapper's mapping to use for redirects that @josephperrott is working on, but I'm ok removing it here and readding it in the implementation of redirects from AIO to ADEV for v18?

josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 65588bb9f321edbf3575114c0a455fa1d6b9b60b.

PR #1957, opened 2024-04-04
refactor(api-gen): show a Github link to the source for API pages

JeanMeche:
Related to angular/angular#55201 & angular/angular#52641

### DIFF
--- a/bazel/api-gen/rendering/templates/header-api.tsx
+++ b/bazel/api-gen/rendering/templates/header-api.tsx
@@ -7,7 +7,7 @@
  */
 
 import {h} from 'preact';
-import {EntryType} from '../entities';
+import {EntryType, isDocEntryWithSourceInfo} from '../entities';
 import {DocEntryRenderable} from '../entities/renderables';
 import {
   HEADER_CLASS_NAME,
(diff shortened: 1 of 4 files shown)


twerske (approved):

JeanMeche:
Note: Tests are failing due to the dependency on the extractor provided by the FW repo and updated here: #55201

bazel/api-gen/rendering/templates/header-api.tsx
```diff
@@ -21,6 +21,14 @@ import {DocsPillRow} from './docs-pill-row';
 export function HeaderApi(props: {entry: DocEntryRenderable; showFullDescription?: boolean}) {
   const entry = props.entry;
 
+  // TODO: This link point to the main branch.
+  // When ADEV is not deployed on the main branch branch anymore,
+  // We should update it to point to the tag of the released version which ADEV runs on.
+
+  const sourceUrl = isDocEntryWithSourceInfo(entry)
+    ? `https://github.com/angular/angular/blob/main${entry.source.filePath}#L${entry.source.startLine}-${entry.source.endLine}`
```
josephperrott, line 29: is `entry.source.filePath` guaranteed to be `/` prefixed?

josephperrott (approved):
LGTM

Just one small question

bazel/api-gen/rendering/templates/header-api.tsx
josephperrott: Looking again closer, it looks like the extraction test json file doesn't have it as prefixed, so it wouldn't even point to the correct place there.

JeanMeche: I just checked against the FW directly, the change is correct, the path has a leading `/`. I changed the test data to reflect that.

josephperrott:
This PR was merged into the repository by commit db01d538ad135735bc3e0692a8019921b91dc89b.

PR #1986, opened 2024-04-18
ci: use keepUpdatedLabel to control autorebasing of renovate PRs

josephperrott:
Using keepUpdatedLabel we can turn on and off the autorebasing functionality of renovate.  Rather than having renovate always autorebase, we will enable this using a label which renovate will automatically create PRs with. This will cause the current behavior to continue by default, but allow us to remove the label for PRs which are blocked, but renovate would automatically regenerate against our wishes.

The label we will use in `renovate managed`.


renovate.json
```diff
@@ -7,7 +7,9 @@
   "automerge": false,
   "separateMajorMinor": false,
   "timezone": "America/Tijuana",
-  "labels": ["comp: build", "dependencies", "action: merge"],
+  "labels": ["comp: build", "renovate managed", "action: merge"],
+  "keepUpdatedLabel": "renovate managed",
+  "rebaseWhen": "conflicted",
```
devversion, line 12: Note that sometimes multiple PRs land and those may interplay with renovate update PRs. Is there a way to rebase like every 2nd day?

devversion (approved):

clydin (approved):

renovate.json
josephperrott: Unfortunately, as I read it at least, there is not any sort of time based triggering.  The options are `never`, `conflicted` or `behind-base-branch`.

josephperrott:
This PR was merged into the repository by commit 72c79abed60d3b196ccdf9bad5a55b1d0d6c5f82.

PR #1989, opened 2024-04-18
feat(ng-dev/format): allow expansion of file paths using /...

josephperrott:
Allow the expansion to subdirectories using /... when providing files to ng-dev format

### DIFF
--- a/ng-dev/format/BUILD.bazel
+++ b/ng-dev/format/BUILD.bazel
@@ -12,6 +12,7 @@ ts_library(
         "@npm//@types/node",
         "@npm//@types/yargs",
         "@npm//cli-progress",
+        "@npm//fast-glob",
         "@npm//multimatch",
         "@npm//yargs",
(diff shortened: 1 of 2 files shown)


devversion:
Sorry for being a little conservative. Is there a strong reason for this? might be better just saying the patterns in there are glob patterns via fast-glob, compared to allowing some mixes IMO. Feel free to merge though

devversion (approved):
I guess I could see it on the command line, and with bash expansion being annoying, so LGTM

josephperrott:
Yeah, the problem is that bash will automatically try to expand it if you put `**/*` on the command line

josephperrott:
This PR was merged into the repository by commit 702d5f5c2a78541d6cbb35ba758a9cf9181733bb.

devversion:
Probably better would be to just automatically glob all files in a directory 😄

PR #1993, opened 2024-04-19
refactor(shared-docs): use puppeteer instead of playwright

josephperrott:
Additionally, remove TODO about reusing browser instance as it does not seem to be worth the effort to do in the future.


clydin (approved):

josephperrott:
This PR was merged into the repository by commit 862c199cdfd652fb116e5821223b8189edb0f30c.

PR #2004, opened 2024-04-23
fix(ng-dev): remove constraint that TESTED comment is a review comment

thePunderWoman:
The requirement of the TESTED= comment being a review comment was confusing to users and adding a barrier to landing code. This change updates the requirements to have the comment just be present from a googler.

### DIFF
--- a/ng-dev/pr/common/validation/assert-enforce-tested.ts
+++ b/ng-dev/pr/common/validation/assert-enforce-tested.ts
@@ -6,11 +6,15 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {AuthenticatedGithubClient} from '../../../utils/git/github.js';
 import githubMacros from '../../../utils/git/github-macros.js';
-import {PullRequestFromGithub} from '../fetch-pull-request.js';
+import {
+  fetchPullRequestCommentsFromGithub,
+  PullRequestFromGithub,
+  PullRequestCommentsFromGithub,
+} from '../fetch-pull-request.js';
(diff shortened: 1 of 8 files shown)


ng-dev/pr/common/validation/assert-enforce-tested.ts
```diff
@@ -44,18 +53,34 @@ function pullRequestRequiresTGP(pullRequest: PullRequestFromGithub): boolean {
   return pullRequest.labels.nodes.some(({name}) => name === requiresLabels.REQUIRES_TGP.name);
 }
 
+export class PullRequestComments {
```
josephperrott, line 56: Can this just be a function instead of class that gets instantiated and then executed?

ng-dev/pr/common/validation/assert-enforce-tested.ts
```diff
@@ -20,12 +24,17 @@ export const enforceTestedValidation = createPullRequestValidation(
 );
 
 class Validation extends PullRequestValidation {
-  async assert(pullRequest: PullRequestFromGithub, gitClient: AuthenticatedGithubClient) {
+  async assert(pullRequest: PullRequestFromGithub, gitClient: AuthenticatedGitClient) {
```
josephperrott, line 27: What are we changing the client that is used?

ng-dev/pr/common/validation/assert-enforce-tested.ts
thePunderWoman: The reason for the class is so testing has a static create function to mock out the call to the API.

josephperrott (approved):
LGTM

PR #2010, opened 2024-04-25
fix(ng-dev): show circular dependencies approve command when golden does match

alan-agius4:
Currently, the approve command is not being displayed when the golden file does not match.

### DIFF
--- a/ng-dev/ts-circular-dependencies/index.ts
+++ b/ng-dev/ts-circular-dependencies/index.ts
@@ -139,17 +139,19 @@ export function main(
         `\n   Total: ${newCircularDeps.length} new cycle(s), ${fixedCircularDeps.length} fixed cycle(s). \n`,
       ),
     );
-    if (approveCommand) {
-      Log.info(yellow(`   Please approve the new golden with: ${approveCommand}`));
-    } else {
-      Log.info(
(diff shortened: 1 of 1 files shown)


ng-dev/ts-circular-dependencies/index.ts
```diff
@@ -139,17 +139,19 @@ export function main(
...
-    }
   }
+
+  if (approveCommand) {
+    Log.info(yellow(`   Please approve the new golden with: ${approveCommand}`));
+  } else {
+    Log.info(
+      yellow(
+        `   Please update the golden. The following command can be ` +
+          `run: yarn ts-circular-deps approve ${getRelativePath(process.cwd(), goldenFile)}.`,
```
josephperrott, line 150: I know this is just copied from before but we should correct it
```suggestion
          `run: yarn ng-dev ts-circular-deps approve ${getRelativePath(process.cwd(), goldenFile)}.`,
```

josephperrott (dismissed):

josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit eb7a061177826db42777592223ee2e83f49f650a.

PR #2020, opened 2024-04-27
refactor(api-gen): add support for function overloads

JeanMeche:
To fix angular/angular#55556

### DIFF
--- a/bazel/api-gen/rendering/templates/deprecated-label.tsx
+++ b/bazel/api-gen/rendering/templates/deprecated-label.tsx
@@ -21,7 +21,7 @@ export function DeprecatedLabel(props: {
         <span dangerouslySetInnerHTML={{__html: entry.deprecationMessage}}></span>
       </div>
     );
-  } else if ('isDeprecated' in entry) {
+  } else if ('isDeprecated' in entry && entry.isDeprecated) {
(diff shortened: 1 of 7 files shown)


bazel/api-gen/rendering/templates/deprecated-label.tsx
```diff
@@ -21,7 +21,7 @@ export function DeprecatedLabel(props: {
         <span dangerouslySetInnerHTML={{__html: entry.deprecationMessage}}></span>
       </div>
     );
-  } else if ('isDeprecated' in entry) {
+  } else if ('isDeprecated' in entry && entry.isDeprecated) {
```
josephperrott, line 24: Shouldn't we just do `entry.isDeprecated`?

If it doesn't exist then it will be falsy anyway

josephperrott (approved):
LGTM

bazel/api-gen/rendering/templates/deprecated-label.tsx
JeanMeche: This is due to `entry` being a union that we need to discriminate.

josephperrott: 👍

josephperrott:
This PR was merged into the repository by commit a9d6b07a910be9a39d36b4f35756d22c6c083d29.

PR #2033, opened 2024-04-30
refactor(shared-docs): Zoneless updates 

atscott:
* Use ChangeDetectionStrategy.OnPush
* Use signals where components were relying on Default change detection
* Remove zone.run when setting signals because it's not needed in v18
* provide zoneless for component tests to ensure they are compatible


JeanMeche (approved):

josephperrott:
This PR was merged into the repository by commit f5a3e595d9e5feafed09ebe7079973703c7d0372.

PR #2034, opened 2024-05-01
fix(ng-dev): prevent failure on googler check

thePunderWoman:
This adds a try catch around the googler org check to make sure the script does not error out when merging

### DIFF
--- a/.github/local-actions/branch-manager/main.js
+++ b/.github/local-actions/branch-manager/main.js
@@ -68178,9 +68178,12 @@ var Validation5 = class extends PullRequestValidation {
 
 // 
 async function isGooglerOrgMember(client, username) {
-  const response = await client.orgs.checkMembershipForUser({ org: "googlers", username });
-  if (response.status === 204) {
-    return true;
+  try {
+    const response = await client.orgs.checkMembershipForUser({ org: "googlers", username });
+    if (response.status === 204) {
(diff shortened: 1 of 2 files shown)


ng-dev/pr/common/validation/assert-enforce-tested.ts
```diff
@@ -78,11 +78,11 @@ export async function pullRequestHasValidTestedComment(
   gitClient: AuthenticatedGitClient,
 ): Promise<boolean> {
   for (const {bodyText, author} of comments) {
-    if (
-      bodyText.startsWith(`TESTED=`) &&
-      (await githubMacros.isGooglerOrgMember(gitClient.github, author.login))
-    ) {
-      return true;
+    if (bodyText.startsWith(`TESTED=`)) {
+      try {
```
josephperrott, line 82: Can we do this try catch work in the macro itself? Because this issue would exist for anything that relies on the macro

josephperrott (changes requested):

ng-dev/pr/common/validation/assert-enforce-tested.ts
thePunderWoman: Sure

josephperrott (approved):
LGTM

PR #2045, opened 2024-05-03
fix(shared-docs): restructure dependencies

alan-agius4:
Move dependencies unnecessary for the components to a separate section labeled `dependencies`  Since these are not used the by application directly there is no need to add them as `peerDependencies`. As there's no risk of conflicting dependencies in the client bundle, thereby preventing any potential increase in bundle size.


devversion (approved):
Great. Thx!

PR #2053, opened 2024-05-07
fix(ng-dev): clarify isolate primitives validation rules

thePunderWoman:
This updates the isolate primitives validation to ensure that it is clear that once primitives code has been merged, PRs can include FW code as long as primitives code is present in the PR. In that case, those changes are related, which is fine. Pure FW PRs should not be merged in this case.


josephperrott:
This PR was merged into the repository by commit 9ae826d6c385ab347fb26ba5e267c1151120c798.

PR #2056, opened 2024-05-08
fix(docs-infra): prevent doc-icons from being translated

JeanMeche:
fixes angular/angular#55740


josephperrott (approved):
LGTM

JeanMeche:
Yes, this will fix the issue on every page once angular/angular#55781 is merged.

PR #2060, opened 2024-05-13
refactor(ng-dev): export non esm bundled code in ng-dev

josephperrott:
In ng-dev instead of exporting the esm bundles when importing from the @angular/ng-dev entrypoint, we instead should export the non bundled code.

When we don't do this, if we try to bundle the imported code again, we get issues with multiple `createRequire` imports.


clydin (approved):

josephperrott:
This PR was merged into the repository by commit 8b117748ae8243f98d261a9731908867b8e3d876.

PR #2072, opened 2024-05-17
refactor(shared-docs): make the SearchItem directive OnPush compatible

pkozlowski-opensource:
The SearchItem directive has a local state (isActive) used in a host binding. Prior to this change the state was updated from a method call without marking the corresponding view for check. This meant that the directive in question could not be used in an OnPush component as the directive's host bindings would not be re-evaluated.

### DIFF
--- a/docs/directives/search-item/search-item.directive.ts
+++ b/docs/directives/search-item/search-item.directive.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
-import {Directive, ElementRef, Input, inject} from '@angular/core';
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
Unrelated note: it seems to me that the `@Input() disabled` is unused - at least I couldn't find references to it. Happy to send another PR that removes this input.

JeanMeche (approved):

josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit fca0d3ffd591712f78e84c3105ede099b07d4d66.

PR #2084, opened 2024-05-27
build: set es2020 in bazel ts_library usages

crisbeto:
Currently the default `target` for the code compiled with `ts_library` is set through the `ts_library` exposed in `defaults.bzl`, however the code under the `bazel` directory isn't allowed to use `defaults.bzl` which meant that it was defaulting to es2015.

These changes add a `defaults.bzl` specific to the `bazel` directory which allows us to maintain the defaults in a single place.

This came up due to a breakage in https://github.com/angular/angular/pull/56096.

### DIFF
--- a/bazel/defaults.bzl
+++ b/bazel/defaults.bzl
@@ -1,0 +1,21 @@
+load("@npm//@bazel/concatjs:index.bzl", _ts_library = "ts_library")
+
+def ts_library(name, testonly = False, deps = [], srcs = [], devmode_module = None, **kwargs):
+    current_pkg = native.package_name()
+
+    if current_pkg != "bazel" and not current_pkg.startswith("bazel/"):
+        fail("Cannot use `bazel/defaults.bzl` outside of the `bazel` directory. " +
+             "Use //tools:defaults.bzl instead")
+
+    _ts_library(
+        name = name,
+        devmode_module = devmode_module if devmode_module != None else "umd",
+        devmode_target = "es2020",
+        prodmode_module = "esnext",
+        prodmode_target = "es2020",
+        tsconfig = kwargs.pop("tsconfig", "//:tsconfig"),
+        testonly = testonly,
+        deps = deps,
+        srcs = srcs,
+        **kwargs
+    )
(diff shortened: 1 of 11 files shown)


bazel/defaults.bzl
```diff
@@ -0,0 +1,21 @@
+load("@npm//@bazel/concatjs:index.bzl", _ts_library = "ts_library")
```
crisbeto, line 1: Technically we could've fixed the issue by adding `prodmode_target` and `devmode_target` to all the places using `ts_library`, but I figured that this is a bit more maintainable.

bazel/defaults.bzl
```diff
@@ -0,0 +1,21 @@
...
+    if current_pkg != "bazel" and not current_pkg.startswith("bazel/"):
+        fail("Cannot use `bazel/defaults.bzl` outside of the `bazel` directory. " +
+             "Use //tools:defaults.bzl instead")
+
+    _ts_library(
+        name = name,
+        devmode_module = devmode_module if devmode_module != None else "umd",
+        devmode_target = "es2020",
+        prodmode_module = "esnext",
+        prodmode_target = "es2020",
```
devversion, line 15: This is a good idea. Thx.

Do you think might make sense that `//tools:defaults.bzl` uses this? so that we have a single place to maintain for this repo

devversion (approved):

bazel/defaults.bzl
crisbeto: I think it would make sense, but I didn't do it because:
1. I didn't know the context of why `bazel` isn't allowed to import from `tools`.
2. The options are slightly different. E.g. `bazel` uses `devmode_module = "umd"` (same as `ts_library` default) whereas the one from `tools` uses `esnext`. Switching `bazel` to `esnext` ended up breaking some tests. The other difference is that `bazel` doesn't include `tslib` which there was also a test for.

devversion: bazel folder is shipped as its own package so it cannot access anything outside.

I see on the second point. Feel free to merge this.

crisbeto: It doesn't look like I have permission to merge it.

PR #2086, opened 2024-05-28
fix(bazel/api-golden): resolve downstream errors and golden name

crisbeto:
Fixes the following issues after the recent changes:
1. Removes the package check from the Bazel-specific `ts_library`, because it ended up breaking packages that consume the Bazel code.
2. Adds `.api.md` to the generated golden names. This appears to be a recent change that is now required [based on the docs](https://api-extractor.com/pages/configs/api-extractor_json/#apireportreportfilename) that say "The file extension should be .api.md, and the string should not contain a path separator such as \ or /.".


alan-agius4 (approved):

clydin (approved):

bazel/api-golden/index_npm_packages.ts
```diff
@@ -49,7 +49,7 @@ async function main(
     // entry-point we maintain a separate golden file. These golden files are
     // based on the name of the defining NodeJS exports subpath in the NPM package,
     // See: https://api-extractor.com/pages/overview/demo_api_report/.
-    const goldenName = join(subpath, 'index.md');
+    const goldenName = join(subpath, 'index.api.md');
```
crisbeto, line 52: For reference, this came up because all the golden files in the FW repo got renamed to `.api.md` after the latest update. It seems like we were supposed to do this all along, based on https://api-extractor.com/pages/configs/api-extractor_json/#apireportreportfilename

PR #2102, opened 2024-06-10
refactor(ng-dev): Add pr url support to merge tool params

thePunderWoman:
This adds support and parsing for pasting a full github url into the pr merge tool.

### DIFF
--- a/ng-dev/pr/merge/cli.ts
+++ b/ng-dev/pr/merge/cli.ts
@@ -9,9 +9,8 @@
 import {Argv, Arguments, CommandModule} from 'yargs';
 import {addDryRunFlag} from '../../utils/dry-run.js';
 
-import {mergePullRequest} from './merge-pull-request.js';
+import {mergePullRequest, parsePrNumber} from './merge-pull-request.js';
 import {addGithubTokenOption} from '../../utils/git/github-yargs.js';
-
 /** The options available to the merge command via CLI. */
 export interface MergeCommandOptions {
   pr: number;
@@ -28,7 +27,8 @@ async function builder(argv: Argv) {
     .strict()
     .positional('pr', {
       demandOption: true,
-      type: 'number',
+      coerce: (prUrlOrNumber: string) => parsePrNumber(prUrlOrNumber),
+      type: 'string',
       description: 'The PR to be merged.',
     })
     .option('branch-prompt' as 'branchPrompt', {
--- a/ng-dev/pr/merge/merge-pull-request.ts
+++ b/ng-dev/pr/merge/merge-pull-request.ts
@@ -19,7 +19,6 @@ import {
   PullRequestValidationError,
   UserAbortedMergeToolError,
 } from './failures.js';
-import {createPullRequestValidationConfig} from '../common/validation/validation-config.js';
 import {
   InvalidTargetBranchError,
   InvalidTargetLabelError,
@@ -133,4 +132,18 @@ async function createPullRequestMergeTool(flags: PullRequestMergeFlags) {
     }
     throw e;
   }
+}
+
+/**
+ * Parses the pull request number from either the number or url string
+ */
+export function parsePrNumber(prUrlOrNumber: string): number {
+  // There is no url validation here other than presence of `/`.
+  // So whatever is the last segment of that string, url or not, will be
+  // parsed as a PR number.
+  const prNumber = parseInt(prUrlOrNumber.split('/').pop()!);
+  if (isNaN(prNumber)) {
+    throw new Error('Pull Request was unable to be parsed from the parameters');
+  }
+  return prNumber;
 }
--- a/ng-dev/pr/merge/integration.spec.ts
+++ b/ng-dev/pr/merge/integration.spec.ts
@@ -25,7 +25,7 @@ import {
   PackageJson,
 } from '../../release/versioning/index.js';
 import {Prompt} from '../../utils/prompt.js';
-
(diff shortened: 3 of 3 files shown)


ng-dev/pr/merge/merge-tool.ts
```diff
@@ -316,4 +318,15 @@ export class MergeTool {
       Log.debug('Assuming correct access because this a bot account.');
     }
   }
+
+  parsePrNumber(prUrlOrNumber: number | string): number {
```
josephperrott, line 322: Rather than parsing this information in the tool, we should do the extra parsing during the parsing phase of `yargs` tooling?  https://github.com/angular/dev-infra/blob/03b8a7dffd1205e061f0bee949024ebefc2a6592/ng-dev/pr/merge/cli.ts#L25

Because I believe that `yargs` will try to coerce whatever is provided here into a number so the `typeof` check would fail anyway

You will get this failure `Variable $number of type Int! was provided invalid value`

josephperrott (changes requested):

ng-dev/pr/merge/merge-tool.ts
thePunderWoman: Ah! I wondered if I was doing this in the wrong spot. I'll take a look.

thePunderWoman: I've updated. Looks like you can't do multiple types on yargs. So I had to go with `string` and then parse from there.

ng-dev/pr/merge/cli.ts
```diff
@@ -56,7 +55,7 @@ async function handler({
   dryRun,
   ignorePendingReviews,
 }: Arguments<MergeCommandOptions>) {
-  await mergePullRequest(pr, {branchPrompt, forceManualBranches, dryRun, ignorePendingReviews});
+  await mergePullRequest(parsePrNumber(pr), {branchPrompt, forceManualBranches, dryRun, ignorePendingReviews});
```
josephperrott, line 58: Rather than parsing here, you should be able to leverage the `coerce` value in the `positional` definition: https://yargs.js.org/docs/#api-reference-coercekey-fn

That way you by the time its done parsing, it is already coerced into a number.

ng-dev/pr/merge/merge-pull-request.ts
```diff
@@ -134,3 +133,14 @@ async function createPullRequestMergeTool(flags: PullRequestMergeFlags) {
...
   }
 }
+
+/**
+ * Parses the pull request number from either the number or url string
+ */
+export function parsePrNumber(prUrlOrNumber: string): number {
+  const prNumber = parseInt(prUrlOrNumber.split('/').pop()!);
+  if (isNaN(prNumber)) {
+    throw new FatalMergeToolError('Pull Request was unable to be parsed from the parameters');
```
josephperrott, line 143: Since this failure is now moved into the yargs parsing, we aren't running within the context of the merge tool, so `FatalMergeToolError` is not necessary, we should just through a generic error.

ng-dev/pr/merge/cli.ts
thePunderWoman: Ahhh, this is what you meant. OK. Updated.

ng-dev/pr/merge/merge-pull-request.ts
```diff
@@ -134,3 +133,14 @@ async function createPullRequestMergeTool(flags: PullRequestMergeFlags) {
     throw e;
   }
 }
+
+/**
+ * Parses the pull request number from either the number or url string
+ */
+export function parsePrNumber(prUrlOrNumber: string): number {
+  const prNumber = parseInt(prUrlOrNumber.split('/').pop()!);
```
josephperrott, line 141: We should at least have a comment noting that this just takes whatever the result is from `parseInt` on the last segment.

Technically, could do `thisis123` and it would result in `123`, this isnt validating that its a github link for the repo or anything.  Not a big deal, I just want to make sure that its noted what its doing.

josephperrott (approved):

ng-dev/pr/merge/merge-pull-request.ts
thePunderWoman: Well, it's still splitting on `/`. So it wouldn't result in 123 in that case. It would still be not a number and fail. But I can add a comment explaining this.

thePunderWoman: Comment added. We could also move this function to a utils file and use it any time PR numbers are provided as a param, but that'll be a future refactor.

ng-dev/pr/merge/cli.ts
```diff
@@ -28,6 +27,7 @@ async function builder(argv: Argv) {
     .strict()
     .positional('pr', {
       demandOption: true,
+      coerce: (prUrlOrNumber) => parsePrNumber(prUrlOrNumber),
       type: 'number',
```
josephperrott, line 31: Evidently the type determines the parsing that is done by yargs automatically, then it sees the type as the result of the `coerce` function, so we need to declare the type as `string` here, and then it will get coerced into number.

```suggestion
      coerce: (prUrlOrNumber: string) => parsePrNumber(prUrlOrNumber),
      type: 'string',
```

josephperrott (changes requested):

josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 4dea6062cae911ba735a4854333ad19f6317bb88.

PR #2111, opened 2024-06-12
refactor(ng-dev): Add area labels to commit message github action

thePunderWoman:
This queries the repository for the list of area labels and determines whether they should apply by commit scope.

### DIFF
--- a/github-actions/commit-message-based-labels/lib/main.ts
+++ b/github-actions/commit-message-based-labels/lib/main.ts
@@ -19,6 +19,8 @@ class CommitMessageBasedLabelManager {
     }
   };
 
+  /** Area labels in the current repository */
+  repoAreaLabels = new Set<string>();
   /** Labels currently applied to the PR. */
   labels = new Set<string>();
   /** All commits in the PR */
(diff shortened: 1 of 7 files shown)


ng-dev/pr/common/labels/managed.ts
```diff
@@ -32,4 +32,14 @@ export const managedLabels = createTypedObject<ManagedLabel>()({
...
   },
+  DETECTED_PERF_CHANGE: {
+    description: 'Issues related to performance',
+    name: 'area: performance',
+    commitCheck: (c: Commit) => c.type === 'perf',
+  },
+  DETECTED_ADEV_CHANGE: {
+    description: 'Angular.dev documentation',
+    name: 'area: adev',
+    commitCheck: (c: Commit) => c.type === 'adev',
```
josephperrott, line 43: `adev` is not a commit type right?

thePunderWoman: Ah, yeah. Copy paste since I added all of the area labels initially. Removed.

github-actions/commit-message-based-labels/lib/main.ts
```diff
@@ -46,6 +48,13 @@ class CommitMessageBasedLabelManager {
         await this.removeLabel(name);
       }
     }
+
+    for (const commit of this.commits) {
+      const label = 'area: ' + commit.scope;
+      if (this.repoAreaLabels.has(label) && !this.labels.has(label)) {
```
josephperrott, line 54: We don't need to check if the repo as the area labels as we already know it will since its a managed label.

thePunderWoman: What about the case with an invalid commit header? The lint check would fail, but then this action would run and add an invalid area label if we don't check. Someone could be like `fix(blah): stuff` and then this would just shove `area: blah` even though that's not what we want.

josephperrott: AHh that makes sense, can you add commenting to express that for the future as well?

thePunderWoman: Comment added!

josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit fd294b581a08f22a81077eaf7e8f308e975fc618.

PR #2114, opened 2024-06-13
refactor(shared-docs): Use the router on absolute url that are on the…

JeanMeche:
… same domain

The CLI docs produce absolute URLs that may target the same domain (exception is next.angular.dev).

When we ecounter those URLs, we remove the domain and feed them to `navigateByUrl` so they are handled as a local navigation

fixes angular/angular#56437

### DIFF
--- a/docs/utils/navigation.utils.ts
+++ b/docs/utils/navigation.utils.ts
@@ -131,7 +131,7 @@ export const getBaseUrlAfterRedirects = (url: string, router: Router): string =>
   return normalizePath(route.toString());
 };
 
-export function handleHrefClickEventWithRouter(e: Event, router: Router) {
+export function handleHrefClickEventWithRouter(e: Event, router: Router, relativeUrl: string) {
   const pointerEvent = e as PointerEvent;
   if (
     pointerEvent.ctrlKey ||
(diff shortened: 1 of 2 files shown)


docs/utils/navigation.utils.ts
```diff
@@ -147,11 +147,20 @@ export function handleHrefClickEventWithRouter(e: Event, router: Router) {
...
   }
 
-  const relativeUrl = closestAnchor?.getAttribute?.('href');
-  if (relativeUrl) {
-    e.preventDefault();
-    router.navigateByUrl(relativeUrl);
+  const hrefAttr = closestAnchor?.getAttribute?.('href');
+  let relativeUrl: string;
+  if (hrefAttr?.startsWith('http')) {
+    // Url isn't relative but we're targeting the same domain
```
alan-agius4, line 153: How are you determining that it’s the same domain?

alan-agius4: Found it, it’s called responsibly to filter that. 

NIT: but would be clearer if the function is self contained and not rely on the caller to perform this filtering. As from re-usability perspective, it makes this function very error prone.

JeanMeche: agreed.

josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit da5c892b503ff0ca34e295d2ccbce907d1e401a4.

PR #2122, opened 2024-06-17
refactor(shared-docs): Update return type of content loader to includ…

atscott:
…e RedirectCommand

The content loader is only used as a resolver and can return a `RedirectCommand` to imperatively redirect in the router rather than causing a new navigation in the error case for fetching the data.


josephperrott (approved):
LGTM

PR #2127, opened 2024-06-20
fix(ng-dev): attempt to parse yarn 2+ lock files via yaml

clydin:
Yarn lock files for version 2 and higher now use a Yaml compatible format and do not require a special parser. To support both legacy 1.x and newer lock files, the existing special parser is first tried and if that fails a Yaml parser is then used.


josephperrott (approved):
LGTM

clydin:
This PR was merged into the repository by commit 24d66aa1e5c592f4784c2c878b2c2f92c374c6c4.

PR #2144, opened 2024-06-27
fix(shared-docs): include hiddenFiles in generated metadata

crisbeto:
Updates the metadata script to generate the `hiddenFiles` field.

### DIFF
--- a/docs/pipeline/tutorials/metadata.ts
+++ b/docs/pipeline/tutorials/metadata.ts
@@ -25,7 +25,9 @@ export async function generateMetadata(
     allFiles: Object.keys(files),
     tutorialFiles,
     answerFiles: await getAnswerFiles(path, config, files),
-    hiddenFiles: [],
+    hiddenFiles: config.openFiles
+      ? Object.keys(tutorialFiles).filter((filename) => !config.openFiles!.includes(filename))
+      : [],
     dependencies: {
       ...dependencies,
       ...devDependencies,


docs/pipeline/tutorials/metadata.ts
```diff
@@ -25,7 +25,9 @@ export async function generateMetadata(
     allFiles: Object.keys(files),
     tutorialFiles,
     answerFiles: await getAnswerFiles(path, config, files),
-    hiddenFiles: [],
+    hiddenFiles: config.openFiles
+      ? Object.keys(tutorialFiles).filter((filename) => !config.openFiles!.includes(filename))
```
josephperrott, line 29: Since `openFiles` could be undefined, we should just use optional chaining here

```suggestion
      ? Object.keys(tutorialFiles).filter((filename) => !config.openFiles?.includes(filename))
```

josephperrott (approved):
LGTM

docs/pipeline/tutorials/metadata.ts
crisbeto: It can't be undefined since we null check it in the ternary one level up.

josephperrott: Does typescript not reconcile that for us then?  I can merge as is then.

josephperrott:
This PR was merged into the repository by commit 506a93d7f2dba452c78348eea9c983bbe0d9d1ca.

PR #2145, opened 2024-06-28
fix(shared-docs): properly generate hiddenFiles

crisbeto:
Fixes that the `hiddenFiles` array was always coming up empty because we were filtering on the files that were filtered already.


josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 720be096e12c01c8346d7787e0931b184525b925.

PR #2154, opened 2024-07-05
refactor(api-gen): parse links with descriptions

JeanMeche:
We already parsed the links correctly for the pills but didn't for function descriptions

2nd commits also support markdown links with titles. 

fixes angular/angular#56856
fixes angular/angular#56857

### DIFF
--- a/bazel/api-gen/rendering/templates/docs-pill-row.tsx
+++ b/bazel/api-gen/rendering/templates/docs-pill-row.tsx
@@ -15,11 +15,11 @@ export function DocsPillRow(props: {links: LinkEntryRenderable[]}) {
 
   return (
     <nav class="docs-pill-row">
-      {
-        props.links.map((link) => (<a class="docs-pill" href={link.url}>
+      {props.links.map((link) => (
+        <a class="docs-pill" href={link.url} title={link.title}>
           {link.label}
-        </a>))
-      }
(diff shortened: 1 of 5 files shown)


bazel/api-gen/rendering/templates/docs-pill-row.tsx
```diff
@@ -15,11 +15,11 @@ export function DocsPillRow(props: {links: LinkEntryRenderable[]}) {
 
   return (
     <nav class="docs-pill-row">
-      {
-        props.links.map((link) => (<a class="docs-pill" href={link.url}>
+      {props.links.map((link) => (
+        <a class="docs-pill" href={link.url} title={link.title}>
```
josephperrott, line 19: Are we going to run into issues if we have `title=` because `link.title` is undefined?

josephperrott (approved):
LGTM

bazel/api-gen/rendering/templates/docs-pill-row.tsx
JeanMeche: afaik it won't be rendered if its `undefined`.

josephperrott: perfect, just wanted to sanity check.

PR #2177, opened 2024-07-17
refactor(shared-docs): improve heading id extraction 

### LINKED ISSUE #2172: docs: `getHeaderId()` cannot be localized
In the current markdown processing, each heading in the text gets its own `id` from the text content, but it supports only English. 

https://github.com/angular/dev-infra/blob/main/docs/pipeline/guides/state.ts#L11-L30

```ts
export const getHeaderId = (id: string): string => {
  const numberOfHeaderOccurrencesInTheDocument = headerIds.get(id) ?? 0;
  headerIds.set(id, numberOfHeaderOccurrencesInTheDocument + 1);

  const cleanedUpId = id
    .toLowerCase()
    .replaceAll(/<code>(.*?)<\/code>/g, '$1') // remove <code>
    .replaceAll(/<strong>(.*?)<\/strong>/g, '$1') // remove <strong>
    .replaceAll(/<em>(.*?)<\/em>/g, '$1') // remove <em>
    .replace(/\s|\//g, '-') // remove spaces and slashes
    .replace(/gt;|lt;/g, '') // remove escaped < and >
    .replace(/&#\d+;/g, '') // remove HTML entities
    .replace(/[^0-9a-zA-Z\-]/g, ''); // only keep letters, digits & dashes

  const headerId = numberOfHeaderOccurrencesInTheDocument
    ? `${cleanedUpId}-${numberOfHeaderOccurrencesInTheDocument}`
    : cleanedUpId;

  return headerId;
};
```

Previously, in "aio", we can use `{@a header-id}` to tell the transformer the correct heading id.

```md
## foobar {@a foo-bar}
```

This functionality is necessary for non-English localization, because in a vast document, guides refer to each other with links, and these links must be maintained.

## Request

- Any method to allow us to localize headings while keeping ID references. 

## Idea

It's enough that supports `{#heading-id}` syntax implemented by other tools.

https://www.markdownguide.org/extended-syntax/#heading-ids
https://github.com/markedjs/marked-custom-heading-id

```md
## foobar {#foo-bar}
```

JeanMeche:
* Add support for non ascii languages
* add suport for custom id

fixes #2172

### DIFF
--- a/docs/pipeline/guides/state.ts
+++ b/docs/pipeline/guides/state.ts
@@ -8,23 +8,32 @@
 
 const headerIds = new Map<string, number>();
 
-export const getHeaderId = (id: string): string => {
-  const numberOfHeaderOccurrencesInTheDocument = headerIds.get(id) ?? 0;
-  headerIds.set(id, numberOfHeaderOccurrencesInTheDocument + 1);
+export const getHeaderId = (heading: string): string => {
+  const numberOfHeaderOccurrencesInTheDocument = headerIds.get(heading) ?? 0;
+  headerIds.set(heading, numberOfHeaderOccurrencesInTheDocument + 1);
 
-  const cleanedUpId = id
-    .toLowerCase()
-    .replaceAll(/<code>(.*?)<\/code>/g, '$1') // remove <code>
-    .replaceAll(/<strong>(.*?)<\/strong>/g, '$1') // remove <strong>
-    .replaceAll(/<em>(.*?)<\/em>/g, '$1') // remove <em>
-    .replace(/\s|\//g, '-') // remove spaces and slashes
-    .replace(/gt;|lt;/g, '') // remove escaped < and >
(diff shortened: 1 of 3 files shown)


lacolaco:
@JeanMeche Thanks for the quick fix. However, the current approach will not work for most cases.

It is more important to be able to translate a text while maintaining the ID of the original text than to ID it in the translated language. This is because we do not translate all documents at the same time, but in stages, so we have to keep the links from the untranslated documents working.

JeanMeche:
The issue I have with the solution used on AIO is that it wasn't part of neither the standard nor the extended markdown syntax. 

But you're right we could go with the `{# myId}` syntax which is part of the extended spec.

JeanMeche:
@lacolaco Both will be supported now 👌

josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 2c0c71a7bd4c285ede149858650fba9dd2a1544e.

PR #2178, opened 2024-07-18
fix(github-actions): allow `Apache-2.0 AND BSD-3-Clause` license

alan-agius4:
google-protobuf is licensed under a combination of two licenses: Apache-2.0 and BSD-3-Clause.

### DIFF
--- a/github-actions/linting/licenses/dependency-review-config.yml
+++ b/github-actions/linting/licenses/dependency-review-config.yml
@@ -1,6 +1,7 @@
 vulnerability_check: false
 allow_licenses:
   - '0BSD'
+  - 'Apache-2.0 AND BSD-3-Clause' # Needed for google-protobuf
   - 'Apache-2.0'
   - 'Artistic-2.0'
   - 'BlueOak-1.0.0'


devversion (approved):

devversion:
Is there a way to only allow the given package?

alan-agius4:
Yes, via `allow-dependencies-licenses` not sure how to merge it with the input though.
https://github.com/angular/dev-infra/blob/53c706164e66b6b574149064e392ba7553fb055e/github-actions/linting/licenses/action.yml#L21

devversion:
Where is the action used? is this in angular-cli?

alan-agius4:
> Where is the action used? is this in angular-cli?

For the time being, but recently it was moved here so it will be be rolled out on all the repos

devversion:
Oh yeah, definitely makes sense to add, but I'm wondering if we can just allow it for e.g. `@bazel/worker`. Otherwise other dependencies with this license slip through, or is this a valid license in general?

alan-agius4:
It is a combination of 2 valid license. Both Apache-2.0 and BSD-3-Clause are allowed on their own.

devversion:
LGTM then!

PR #2180, opened 2024-07-18
ci: no longer use RBE executors for macos test, use macos-13 runner

josephperrott:
Use the macos-13 runner as it uses an intel chip. Additionally, we realized we should not not be doing these executions on RBE as point is to test in a fully osx environment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,8 @@ jobs:
 
   test-macos:
     timeout-minutes: 30
-    runs-on: macos-latest
+    # Use macos-13 runner to use intel chip.
+    runs-on: macos-13
     steps:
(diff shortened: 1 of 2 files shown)


devversion:
Hmm. should we really disable this? I feel like we often disable, but never re-enable fix. E.g. see worker in angular/components

josephperrott:
@devversion PTAL as I found that it was based on the chip that was being used in the test, it seems that its misdetecting the correct type of chip somewhere in the process. Moving to use a runner with an x86 chip worked as expected.

devversion (approved):
LGTM. At some point we should likely figure out why M1 runners would fail. Would be good coverage

josephperrott:
This PR was merged into the repository by commit de8a879b1168657494ef108db3c24bbda806b39b.

PR #2183, opened 2024-07-19
feat(apps): allow for query param usage in testing redirects

josephperrott:
By using query parameters, we are able to test if a redirect works as expected and goes to the correct location before the DNS record is set up and active.


JeanMeche (approved):

josephperrott:
This PR was merged into the repository by commit a630e3e3793237d7c68e224f2eafd53bd73fa384.

PR #2193, opened 2024-07-22
fix(apps): update query based redirecting in dns-redirecting app

josephperrott:
Require that query based redirecting only be valid when the request is made at dns.angular.dev. Additionally, we now include the full URL output when a 404 is returned for not matching any subdomain.


devversion (approved):

josephperrott:
This PR was merged into the repository by commit 2aa7705387b4e8f428c1d92584e55945fd2c00e2.

PR #2200, opened 2024-07-24
build: perform build of genfiles in parallel before updates

josephperrott:
Build all of the generated file targets before running the update command to allow the targets to be build in parallel.

### DIFF
--- a/tools/update-generated-files.ts
+++ b/tools/update-generated-files.ts
@@ -31,6 +31,9 @@ if (queryProcess.status !== 0) {
 
 const updateTargets = queryProcess.stdout.trim().split(/\r?\n/);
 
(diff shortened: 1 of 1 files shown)


josephperrott:
@devversion `bazel run` does not allow concurrency, we can't call the binary outside of bazel because the binary is a bazel action.

It works using `spawnSync` because while appears synchronous, bazel actually is just starting a server, and running actions on that server asynchronously and then returning the results of all of the items in sync.

devversion (approved):
Ohhh. I misunderstood. You are just building in parallel. Nice improvement!

josephperrott:
This PR was merged into the repository by commit d4e37342479b99925fefee010b223ea0b402e77e.

Issue #2195, opened 2024-07-23
docs: explicit header ID should be removed from heading label


lacolaco:
https://github.com/angular/dev-infra/blob/41a9f08706cb37aa49d325729adfbb4e95070aba/docs/pipeline/guides/state.ts#L16-L17

This transformation doesn't remove the `{#explicitId}` from the source heading label.  

![CleanShot 2024-07-23 at 10 11 00@2x](https://github.com/user-attachments/assets/533ed6f0-4a86-4b10-88e6-2d195698105d)

lacolaco:
@JeanMeche Could you address the issue?

lacolaco:
The label is created at here:

https://github.com/angular/dev-infra/blob/41a9f08706cb37aa49d325729adfbb4e95070aba/docs/pipeline/guides/tranformations/heading.ts#L31-L32

I think the logic should follow the steps below;

- Try to extract an explicit ID in the label
- If ID is found, use it as ID and the rest as a label.
- If not, transform the label to ID

PR #2202, opened 2024-07-25
fix(shared-docs): extract captured group from heading custom id regexp

lacolaco:
`String#match` doesn't return captured group for a "global" pattern. To get it, `RegExp#exec()` is appropriate. The previous test case was passed with a string that converted the entire header to an ID even though the capture failed, so it was fixed to match the custom id exactly.

This PR fixes my fault in #2195


lacolaco:
@JeanMeche @josephperrott Sorry for the repeated bother. Could you please take a look at this?

JeanMeche (approved):

josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 722ce5b647ec89797633452dd293e1dbac776ff1.

PR #2243, opened 2024-08-09
fix(bazel): conditionally add esbuild optimization plugins

alan-agius4:
This update prevents the inclusion of esbuild optimization plugins when both linking and downleveling are disabled. It also removes //shared-scripts/angular-optimization:js_lib from the spec-bundle unless explicit linking is required. This change addresses a circular dependency issue previously observed with this rule in the Angular CLI, leading to the removal of a related patch. For more details, see: https://github.com/angular/angular-cli/blob/3c9697a8c34a5e0f3470bde73f11f9f32107f42e/.yarn/patches/%40angular-build-tooling-https-06f4984cdf.patch

### DIFF
--- a/bazel/spec-bundling/esbuild.config-tmpl.mjs
+++ b/bazel/spec-bundling/esbuild.config-tmpl.mjs
@@ -6,13 +6,37 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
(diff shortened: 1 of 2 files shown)


devversion (approved):
love it

PR #2255, opened 2024-08-16
feat(bazel): check all known index html file names for the index file

josephperrott:
Check for `index.csr.html`, `index.html` or `index.server.html` to serve the default index html file.

### DIFF
--- a/bazel/http-server/server.ts
+++ b/bazel/http-server/server.ts
@@ -136,7 +136,11 @@ export class HttpServer {
   /** Gets the raw content of the index.html. */
   private _getIndexHtmlContent(): string {
     if (!this._index) {
-      const indexPath = this._resolveUrlFromRunfiles('/index.html');
+      // Check all of the known index file names.
+      const indexPath =
+        this._resolveUrlFromRunfiles('/index.html') ||
+        this._resolveUrlFromRunfiles('/index.csr.html') ||
+        this._resolveUrlFromRunfiles('/index.server.html');
 
       if (!indexPath) {
         throw Error('Could not resolve http server index.html');


bazel/http-server/index.bzl
```diff
@@ -80,6 +83,12 @@ http_server_rule = rule(
         "launcher": "%{name}.sh",
     },
     attrs = {
+        "serving_root": attr.string(
+            doc = """
+            The root path to serve files for the application from, defaults to the current
+            directory of the package.
```
devversion, line 89: OOC why is this needed? Isn't the serving root simply the directory where the rule is defined?

I would strongly prefer to preserve that "mental model" because all other access to "deps", or "assets" depends on that being the case as the "root dir".

devversion: This should also be already possible via `rootPaths` ?

josephperrott: In the case of an application created using architect, a `browser` and `server` directory are each created, and we further in framework put this into a `build` directory.

So we need to perform the serving for that project from `build/browser`

devversion: I think this can be achieved via `rootDirs` already?

josephperrott: Ahh I see it now, I was misunderstanding what it was looking for.

josephperrott:
@devversion PTAL, I still would like to check for multiple possible index.html file names.

devversion (approved):

josephperrott:
This PR was merged into the repository by commit 67489d145ae6769e0ca5408ecc91b6bbf8e56733.

PR #2274, opened 2024-08-27
feat(bazel): set COOP and COEP policy for http-server during local de…

josephperrott:
…velopment

Set the COOP and COEP policy in http-server to allow for service workers to work as expected.


josephperrott:
This PR was merged into the repository by commit 5c4a70584b90d5c6eb33e4612a7bb4a0d2c59d59.

PR #2348, opened 2024-09-27
fix(ng-dev): remove formatting expectations around constant only globs

josephperrott:
Remove the formatting requirement for not including constant only globs. While they are technically less efficient, they are not worth flagging for us compared to the cost of the inefficiency.


clydin (approved):

josephperrott:
This PR was merged into the repository by commit 322163a896c47a584de7e721cc7427b017f51ccb.

PR #2359, opened 2024-10-02
feat(ng-dev): create tool for automatically discovering and updating generated files

josephperrott:
Automatically query bazel and find all generated files and run the updates for them.


ng-dev/misc/generated-files/update-generated-files.ts
```diff
@@ -0,0 +1,48 @@
...
+import {BAZEL_BIN} from '../../utils/bazel-bin.js';
+import {green, Log} from '../../utils/logging.js';
+
+export async function updateGeneratedFileTargets(): Promise<void> {
+  const spinner = new Spinner('Querying for all generated file targets');
+
+  // Query for all of the generated file targets
+  const result = await ChildProcess.spawn(
+    BAZEL_BIN,
+    ['query', `"kind(nodejs_binary, //...) intersect attr(name, '.update$', //...)"`],
```
devversion, line 20: Sometimes these things are called `.accept`. Maybe we should make this tag-based?

ng-dev/utils/bazel-bin.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {join} from 'path';
+import {determineRepoBaseDirFromCwd} from './repo-directory.js';
+
+let bazelPath = process.env.BAZEL;
+const bazel = {
+  get BAZEL_BIN() {
```
devversion, line 14: Isn't this getter invoked directly below? 😄 

Love that we respect `BAZEL` env though. I use that to avoid `yarn bazel` in my personal environment

ng-dev/utils/bazel-bin.ts
```diff
@@ -0,0 +1,22 @@
...
+    if (!bazelPath) {
+      bazelPath = join(determineRepoBaseDirFromCwd(), 'node_modules/.bin/bazel');
```
devversion, line 16: IIRC `@bazel/bazelisk` does expose an export to retrieve the binary? On windows `.bin/bazel` may not exist

devversion (approved):
Nice!

ng-dev/utils/bazel-bin.ts
josephperrott: Unfortunately it doesn't support esmodules currently.

ng-dev/utils/bazel-bin.ts
josephperrott: I suppose it is, I was hoping to make it lazy, but your write this is just immediate with more steps.

ng-dev/utils/bazel-bin.ts
devversion: Couldn't you import it via interop?

ng-dev/misc/generated-files/update-generated-files.ts
josephperrott: I think I would prefer to expand the tooling to handle it. Because part of the goal here is that we capture anything that would fail in the tests.  Thoughts on that route?

devversion: I'm just guessing that there could be all types of targets for updating checked-in files. E.g. it may even be called `//:update-some-file`. Those should be marked with `tags = [generated-file]` in their macros IMO

We can always expand this though; so not feeling strongly for now. I think it's cleaner though; as long as we have control over targets

josephperrott: Lets keep it this way for now as we will have to deal with the fact that we don't have controls of the base macros in some cases for applying the tags, for example on `generated_file_test`, it comes from rules_nodejs but is actually itself a macro for both the update and the test action.

I will work on a follow up that expands what it does.  This will allow us to at least update and address updating the generated files on renovate PRs in framework.

devversion (approved):

ng-dev/utils/bazel-bin.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright Google LLC
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {join} from 'path';
+import {determineRepoBaseDirFromCwd} from './repo-directory.js';
+
+export const BAZEL_BIN =
```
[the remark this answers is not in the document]
josephperrott: Switched over to a function

PR #2374, opened 2024-10-07
fix(ng-dev): always output the list of discovered targets as a label

josephperrott:
Always output the list of discovered targets for updating generated files as a list of labels instead of including the default types and such


devversion (approved):

josephperrott:
This PR was merged into the repository by commit 358b27d9f887a9a6834023666647ea19b5b8d323.

PR #2397, opened 2024-10-25
build: update yarn to v4.5.1 

josephperrott:


.github/ng-renovate/.yarnrc.yml
```diff
@@ -0,0 +1 @@
+yarnPath: .yarn/releases/yarn-4.5.1.cjs
```
devversion, line 1: is this necessary? I don't know we vendor yarn inside `ng-renovate/`

devversion (approved):

.github/ng-renovate/.yarnrc.yml
josephperrott: I don't know that it is, but it seems that renovate adds it in during the update, and it should be fine in terms of working, so rather than fight it each time, I think its okay for it to exist here.

PR #2418, opened 2024-11-05
feat(ng-dev): create workflow performance testing tooling

josephperrott:
Creates a piece of tooling within ng-dev that allows for a set of commands to be run that emulate an expected workflow within a repository. This is then measured for the time it takes to run these commands so they can be checked or tracked over time.


.ng-dev/perf-tests/test-rerun.diff
```diff
@@ -0,0 +1,17 @@
...
+@@ -29,9 +29,9 @@ describe('G3Stats', () => {
+   });
+ 
+   function setupFakeSyncConfig(config: GoogleSyncConfig): string {
+-    const configFileName = 'sync-test-conf.json';
+-    fs.writeFileSync(path.join(git.baseDir, configFileName), JSON.stringify(config));
+-    return configFileName;
++    const somethingelse = 'sync-test-conf.json';
++    fs.writeFileSync(path.join(git.baseDir, somethingelse), JSON.stringify(config));
++    return somethingelse;
```
devversion, line 14: I know we've chatted a bit about the patching; and being able to "reverse" is great; but it's worth discussing further IMO. Maintaining these patches can probably quickly break?! so we'd need to pick files that rarely change.

The reversing of patches is a good point; but I believe `git reset` would also work right?

ng-dev/perf/workflow/cli.ts
```diff
@@ -0,0 +1,55 @@
...
+interface WorkflowsParams {
+  configFile: string;
+  json: boolean;
+}
+
+/** Builds the checkout pull request command. */
+function builder(yargs: Argv) {
+  return yargs
+    .option('config-file' as 'configFile', {
+      default: '.ng-dev/workflows.yml',
```
devversion, line 24: Maybe something like?

```suggestion
      default: '.ng-dev/dx-perf-workflows.yml',
```

ng-dev/perf/workflow/cli.ts
```diff
@@ -0,0 +1,55 @@
...
+      type: 'string',
+      description: 'The path to the workflow definitions in a yml file',
+    })
+    .option('json', {
+      default: false,
+      type: 'boolean',
+      description: 'Whether to ouput the results as a json object',
```
devversion, line 31:
```suggestion
      description: 'Whether to output the results as a json object',
```

ng-dev/perf/workflow/workflow.ts
```diff
@@ -0,0 +1,60 @@
...
+ * provided on a single line.
+ */
+async function runCommands(cmds?: string) {
+  cmds = cmds?.trim();
+  if (!cmds) {
+    return;
+  }
+  let commands = cmds
+    .split('\n')
+    .filter((_) => !!_)
```
devversion, line 54: Could we use YAML "lists" for individual commands, in case `\n` is used for some multi-line string of a command (as an example)?

ng-dev/perf/workflow/workflow.ts
```diff
@@ -0,0 +1,60 @@
...
+    .map((cmdStr: string) => cmdStr.trim().split(' '));
```
devversion, line 55: This is pretty scary in general because it would also split args with spaces in between. How about using `ChildProcess.exec` which executes "as a shell"?

devversion (approved):

.ng-dev/perf-tests/test-rerun.diff
devversion: But when we wire up more workflows in FW, or other repos, we should chat more I believe?

josephperrott: Agreed. I will leave this as is for now within this repo, and we can see how we want to address the situation when we encounter it in the other repos.

josephperrott:
This PR was merged into the repository by commit db24afe6d0137a064bf881f75895c9e70f6a5e4a.

PR #2423, opened 2024-11-06
refactor(ng-dev): extract common pattern for processing async commands in child-process

josephperrott:
Extract the common pattern for processing the ChildProcess object to be shared instead of requiring management of the same action in two places.


devversion (approved):
Awesome, thank you! so clean 🎉

josephperrott:
This PR was merged into the repository by commit 933dcd42f8837d6d6ce3a28419a760edb94f10e6.

PR #2426, opened 2024-11-07
fix(ng-dev): improve spinner experience when executed in a CI environment

josephperrott:
Improve the spinner experience in a CI environment so that it does not show the same message repeatedly, but also continues to show output to prevent CI environments from treating the execution as inactive.

### DIFF
--- a/ng-dev/utils/spinner.ts
+++ b/ng-dev/utils/spinner.ts
@@ -7,27 +7,74 @@
  */
 
 import {cursorTo, clearLine} from 'readline';
+import {green, red} from './logging.js';
 
+/** Whether execution is in a CI environment. */
+const IS_CI = process.env['CI'];
 /** ANSI escape code to hide cursor in terminal. */
 const hideCursor = '\x1b[?25l';
 /** ANSI escape code to show cursor in terminal. */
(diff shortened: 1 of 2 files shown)


ng-dev/utils/spinner.ts
```diff
@@ -7,27 +7,74 @@
  */
 
 import {cursorTo, clearLine} from 'readline';
+import {green, red} from './logging.js';
 
+/** Whether execution is in a CI environment. */
+const IS_CI = process.env['CI'];
```
devversion, line 13: At this point in time we keep making the spinner more complicated. Should we consider just using a library here?

josephperrott: This is still markedly less complicated than a library has, we can if you would like, but I do personally like the total control we have in this case.  But if you feel strongly I am happy to look into options for us.

devversion (approved):
pretty cool. LGTM.

I don't feel strongly about the npm libary, and I think the code is clean, so LGTM.

josephperrott:
This PR was merged into the repository by commit 11967c8271b12267398936a06f4dcb1379db553c.

PR #2459, opened 2024-11-18
ci: move performance tracking job into its own workflow

josephperrott:
Move the performance tracking job into its own workflow instead of the primary CI job


devversion (approved):

josephperrott:
This PR was merged into the repository by commit baf103e2b40e49cd78d987d1e9cde39b11569c7e.

PR #2460, opened 2024-11-18
refactor(ng-dev): allow for perf workflows to be run for individual workflows

josephperrott:


devversion:
I guess this is mostly for speed and scalability? is there a risk of too much resources used?

devversion (approved):

josephperrott:
Yeah, its mostly for speed and scale. Since we are just using the default runners we will be fine in terms of resources unless we fan out to 20+ jobs in which case they will queue a bit, but still will run.  The hard limit of the matrix is 256 items.

PR #2468, opened 2024-11-20
fix(github-actions): ensure that the artifacts provided for preview service are not symlinks

josephperrott:
Detect situations in which provided metadata is a symlink instead of a direct file which could be attempting to exfiltrate secrets.


devversion (approved):

josephperrott:
This PR was merged into the repository by commit 06f921f45143638238432a624ff7bc14497be1ec.

PR #2470, opened 2024-11-21
fix(github-actions): properly assess the full path for artifact metadata

josephperrott:
Properly find the full path for artifact metadata, previously we were comparing a full path against a relative path which would always fail

### DIFF
--- a/github-actions/previews/upload-artifacts-to-firebase/lib/extract-artifact-metadata.ts
+++ b/github-actions/previews/upload-artifacts-to-firebase/lib/extract-artifact-metadata.ts
@@ -21,17 +21,19 @@ import {artifactMetadata} from '../../constants.js';
 
 async function main() {
   const [artifactDirPath] = process.argv.slice(2);
+  /** The full path to the artifact directory. */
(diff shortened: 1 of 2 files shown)


github-actions/previews/upload-artifacts-to-firebase/lib/extract-artifact-metadata.ts
```diff
@@ -21,17 +21,19 @@ import {artifactMetadata} from '../../constants.js';
 
 async function main() {
   const [artifactDirPath] = process.argv.slice(2);
+  /** The full path to the artifact directory. */
+  const fullArtifactDirPath = await fs.promises.realpath(artifactDirPath);
 
   for (const [key, name] of Object.entries(artifactMetadata)) {
     /** The expected path of the artifact */
-    const expectedPath = path.join(artifactDirPath, name);
+    const expectedPath = path.join(fullArtifactDirPath, name);
```
devversion, line 29: do you want to normalize here as well?

devversion (approved):

PR #2472, opened 2024-11-22
feat(ng-dev): add support for uploading results of workflow testing to database

josephperrott:
Uploads the workflow test results our database when a specific github sha is provided as a reference key for the data.


.github/workflows/perf.yml
```diff
@@ -43,4 +45,9 @@ jobs:
       - uses: ./github-actions/npm/checkout-and-setup-node
       - uses: ./github-actions/bazel/setup
       - run: yarn install --immutable
-      - run: yarn ng-dev perf workflows --name ${{ matrix.workflow }}
+      - uses: 'google-github-actions/auth@v2'
```
devversion, line 48: Do you mind adding some information on the workload identity provider setup for future help?

ng-dev/perf/workflow/database.ts
```diff
@@ -0,0 +1,54 @@
...
+  commit_sha: string;
+  value: number;
+  name: string;
+}
+
+/** Add a single workflow performance result to the spanner database. */
+export async function addWorkflowPerformanceResult(result: WorkflowPerformanceRowResult) {
+  /** The spanner client instance. */
+  const spanner = new Spanner({
+    projectId: 'internal-200822',
```
devversion, line 40: Nit: maybe pass that from the workflow config to avoid repetition?

ng-dev/perf/workflow/cli.ts
```diff
@@ -34,30 +37,48 @@ function builder(yargs: Argv) {
...
 
-  const results: {[key: string]: number} = {};
-  for (const workflow of Object.values(workflows)) {
-    const {name, duration} = await measureWorkflow(workflow);
-    results[name] = duration;
+  if (commitSha) {
+    const spinner = new Spinner('Uploading performance results to database');
+    try {
+      for (let {value, name} of results) {
+        await addWorkflowPerformanceResult({
```
devversion, line 72: Can we still print the results?

devversion (approved):
Nice!

ng-dev/perf/workflow/cli.ts
josephperrott: The result is printed on each result inline.  It just is not longer written as a JSON output.  The result of running the command looks like this:

```

 ✓ Rerun a test: 9639.00ms
 ✓ Upload complete

ng-dev/perf/workflow/database.ts
josephperrott: I am not sure what you are wanting to be passed from the workflow config

devversion: Mostly the `projectId` since I can see the same in the workflow identity configuration section?

ng-dev/perf/workflow/cli.ts
devversion: Oh nice, didn't see that!

ng-dev/perf/workflow/database.ts
devversion: I feel differently given the workflow identity scope, but I also see what you are saying and I think it doesn't matter a lot 👍

josephperrott:
This PR was merged into the repository by commit 2ca2e982c83cbf1da2213c109f6c3654c409b02f.

PR #2474, opened 2024-11-25
feat(bazel): make RBE platform compatible with Bazel 6

devversion:
This makes the RBE platform compatible with Bazel 6


josephperrott (approved):
LGTM

PR #2503, opened 2024-12-16
ci: install `pnpm` in ng-renovate

alan-agius4:
The Angular CLI uses `pnpm` which is not pre-installed.


josephperrott:
Closing as we determined we do not want to do this at this pont.

PR #2514, opened 2024-12-19
feat(ng-dev): support releasing when operating in a hybrid pnpm rules_js repo

devversion:
Updating the `package.json` for a release also requires updating some Aspect lock files that
capture the hash of the `package.json` file.

The release tool currently doesn't do this, and some tricky manual workaround steps are
needed. This commit automatically runs the sync command.

### DIFF
--- a/ng-dev/release/publish/actions.ts
+++ b/ng-dev/release/publish/actions.ts
@@ -222,13 +222,16 @@ export abstract class ReleaseAction {
     // Commit message for the release point.
     const commitMessage = getCommitMessageForRelease(newVersion);
 
+    const filesToCommit = [workspaceRelativePackageJsonPath, workspaceRelativeChangelogPath];
+
+    // Ensure modified Aspect lock files are included in the release commit.
+    // TODO: Remove after `rules_js` migration is complete.
+    if (this.config.rulesJsInteropMode) {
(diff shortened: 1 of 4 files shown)


ng-dev/release/publish/actions.ts
```diff
@@ -221,6 +226,8 @@ export abstract class ReleaseAction {
     await this.createCommit(commitMessage, [
       workspaceRelativePackageJsonPath,
       workspaceRelativeChangelogPath,
+      // Ensure modified Aspect lock files are included in the release commit.
```
alan-agius4, line 229: In some cases, the pnpm lock file gets also updated, example when a package is deprecated. Do we want to handle this case too?

See: https://github.com/angular/angular-cli/pull/29175

alan-agius4 (approved):
LGTM, one question.

ng-dev/release/publish/actions.ts
devversion: Do you think that was caused by e.g. doug running his own version of pnpm?

alan-agius4: No, I encountered I encountered this also on my machine, it think the reasoning is that the sync task cause the lock file not to be immutable.

alan-agius4: Likely the deprecated message was added recently?

devversion: Ahh good point. I will add the lock file

PR #2539, opened 2025-01-09
fix(ng-dev): ensure release actions commit aspect lock files

alan-agius4:
Several actions within the release process were not committing the aspect lock files as expected. This fix ensures that the aspect lock files are properly committed during all relevant stages of the release workflow.


devversion (approved):
LGTM. Good catch

PR #2544, opened 2025-01-10
fix(ng-dev): fix commitCheck functions for managed labels

thePunderWoman:
Some of the managed labels were incorrectly looking at type rather than scope for their commitCheck function, which resulted in always failing to match.


josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 8076a69c44ddeb30662e1621860a2d04d57534a2.

PR #2545, opened 2025-01-10
fix(ng-dev): Add mapping between localize scope and i18n label

thePunderWoman:
This mapping was preferred by the fw-team.


josephperrott (approved):
LGTM

PR #2560, opened 2025-01-29
refactor(ng-dev): separate out logic for takeover, target and regular checkouts

josephperrott:
Separate the logic apart to be easier to parse as well as allow for intentional regular checkout even in a takeover scenario


josephperrott:
This PR was merged into the repository by commit 271a6d3a1950c30380d70725ae762339c34178ca.

PR #2576, opened 2025-02-06
Initial steps for `rules_js` support

devversion:
See individual commits


bazel/api-golden/test_api_report.ts
```diff
@@ -51,17 +50,14 @@ const _origFetchAstModuleExportInfo = ExportAnalyzer.prototype.fetchAstModuleExp
...
-  approveGolden: boolean,
   stripExportPattern: RegExp,
   typePackageNames: string[] = [],
   packageJsonPath = resolveWorkspacePackageJsonPath(),
-  customPackageName?: string,
-): Promise<ExtractorResult> {
-  // If no `TEST_TMPDIR` is defined, then this script runs using `bazel run`. We use
-  // the runfile directory as temporary directory for API extractor.
-  const tempDir = process.env.TEST_TMPDIR ?? process.cwd();
+  customPackageName: string,
```
josephperrott, line 57: If this argument is now required, does not that mean that `typePackageNames` and `packageJsonPath` both are not able to be used as default initializers?

josephperrott, line 57: If this argument is now required, does not that mean that `typePackageNames` and `packageJsonPath` both are not able to be used as default initializers?

bazel/api-golden/test_api_report.ts
```diff
@@ -145,21 +137,30 @@ export async function testApiGolden(
     return info;
   };
 
-  return Extractor.invoke(extractorConfig, {
+  const reportTmpOutPath = path.join(tempDir, `${configObject.apiReport!.reportFileName!}.api.md`);
+
+  const result = Extractor.invoke(extractorConfig, {
     // If the golden should be approved, then `localBuild: true` instructs
```
josephperrott, line 143: This comment is a bit out of date now right?

josephperrott (approved):
LGTM

bazel/api-golden/test_api_report.ts
devversion: Fixed. good catch

PR #2599, opened 2025-02-18
feat(github-actions): support RBE in windows action VMs

devversion:
We may execute inside WSL, and the configure remote action will run on the host, and should be able to forcibly enable RBE.


josephperrott (approved):
LGTM

PR #2607, opened 2025-02-19
fix(github-actions): escaping of double backslash is breaking

devversion:
Apparently the double backslach ends up being escaped, while e.g. C:\wsl doesn't need the escaping :)


josephperrott (approved):
LGTM

PR #2609, opened 2025-02-19
fix(github-actions): do not auto convert to CRLF when checking out

devversion:
This is a useful addition to the shared checkout action that will help integrating with Windows support.


josephperrott (approved):
LGTM

PR #2611, opened 2025-02-19
refactor(ng-dev/ts-circular-dependencies): reorder processing of warnings and golden existence check

josephperrott:
Move the warnings processing before the initial processing of golden file and approval to always include its output.  Additionally, provide alternate text output for approval attempts when no golden exists.

### DIFF
--- a/ng-dev/ts-circular-dependencies/index.ts
+++ b/ng-dev/ts-circular-dependencies/index.ts
@@ -119,13 +119,13 @@ export function main(
       return 1;
     }
   } else {
+    // No golden file exists
     if (approve) {
       Log.error(
         `x  Cannot approve circular depdencies within this repository as no golden file exists.`,
       );
       return 1;
     }
-    // No golden file exists
     if (cycles.length > 0) {
       Log.error(`x  No circular dependencies are allow within this repository.`);
       return 1;
--- a/ng-dev/ts-circular-dependencies/index.ts
+++ b/ng-dev/ts-circular-dependencies/index.ts
@@ -107,19 +107,7 @@ export function main(
     Log.warn(`   Please rerun with "--warnings" to inspect unresolved imports.`);
   }
 
-  if (goldenFile) {
-    // Golden file exists
-    if (approve) {
-      writeFileSync(goldenFile, JSON.stringify(actual, null, 2));
-      Log.info(green('✔  Updated golden file.'));
-      return 0;
-    }
-    if (!existsSync(goldenFile)) {
-      Log.error(`x  Could not find golden file: ${goldenFile}`);
-      return 1;
-    }
-  } else {
-    // No golden file exists
+  if (goldenFile === undefined) {
     if (approve) {
       Log.error(
         `x  Cannot approve circular depdencies within this repository as no golden file exists.`,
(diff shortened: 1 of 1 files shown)


ng-dev/ts-circular-dependencies/index.ts
```diff
@@ -119,6 +107,34 @@ export function main(
...
+      return 1;
+    }
+  } else {
+    if (approve) {
+      Log.error(
+        `x  Cannot approve circular depdencies within this repository as no golden file exists.`,
+      );
+      return 1;
+    }
+    // No golden file exists
```
devversion, line 128: NIt: comment seems off?

ng-dev/ts-circular-dependencies/index.ts
```diff
@@ -119,6 +107,34 @@ export function main(
     Log.warn(`   Please rerun with "--warnings" to inspect unresolved imports.`);
   }
 
+  if (goldenFile) {
```
devversion, line 110: NIt: I'd prefer if we guard by `approve` at the first condition, as that would make it easier to follow IMO.

(maybe that's just me though)

devversion (approved):

ng-dev/ts-circular-dependencies/index.ts
```diff
@@ -119,6 +107,33 @@ export function main(
...
   }
 
+  if (goldenFile === undefined) {
+    if (approve) {
+      Log.error(
+        `x  Cannot approve circular depdencies within this repository as no golden file exists.`,
+      );
+      return 1;
+    }
+    if (cycles.length > 0) {
```
devversion, line 117: Shouldn't this be outside? What if there is a `goldenFile` and cycles?

ng-dev/ts-circular-dependencies/index.ts
```diff
@@ -119,6 +107,33 @@ export function main(
...
+    return 0;
+  }
+
+  if (approve) {
+    writeFileSync(goldenFile, JSON.stringify(actual, null, 2));
+    Log.info(green('✔  Updated golden file.'));
+    return 0;
+  }
+
+  if (!existsSync(goldenFile)) {
```
devversion, line 132: Wouldn't this always exist after writing before?

ng-dev/ts-circular-dependencies/index.ts
josephperrott: This is within the `goldenFile` being `undefined` so there is no golden file.

ng-dev/ts-circular-dependencies/index.ts
josephperrott: The file would only be guaranteed to be written to, if an approval happened.  If its a regular check, then it wouldn't be written in the block above, and could technically not exist.  We verify it exists because just below here we read from it to do the actual comparison.

ng-dev/ts-circular-dependencies/index.ts
devversion: But shouldn't this fail? if there are cycles and `goldenFile` being `!== undefined`?

devversion: I guess it will happen automatically below with the comparison. That's the thing I missed. Very confusing :D

devversion (approved):

ng-dev/ts-circular-dependencies/index.ts
josephperrott: For the case where there is no golden file, and there are cycles here. This fails as it returns a `1` as the exit code.

For cases where there is a golden file to do the comparison to later, we check the specifics.

josephperrott:
This PR was merged into the repository by commit 0ad6a370f70638e785d6ef1f90dc6ede34684a47.

PR #2612, opened 2025-02-19
fix(bazel): forward provided module names for interop compatible `ts_project`s

josephperrott:
Forward the provided module name as a valid import module name for downstream usages in strict deps checks for interop compatible ts_project usage.


devversion (approved):
Aspect is so good!

PR #2615, opened 2025-02-20
feat(github-actions): allow for WSL firewall interface to be configurable

devversion:
This will be useful for e.g. 2025 Windows images that seemingly have a different interface name.


josephperrott (approved):
If we can have options defined so that we know whatever option is provided is valid, we should.

devversion:
Agreed, but not yet sure what the names will be/are, so not way to lock it down :(

PR #2623, opened 2025-02-25
build: allow for `@devinfra//` to be used in downstream repos

devversion:
Currently when trying to import `@devinfra//` in a workspace that doesn't use `rules_nodejs` anymore, we will surface errors.

This is because the top-level BUILD target/package is loaded for the shared tsconfig defaults, while this Bzl package imports from `defaults.bzl` which inherently triggers various load statements for `rules_nodejs`.

This shouldn't be the case, and this commit fixes this.


josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit 07dba2a1b463431b7c1144e1734aa95c033dffa1.

PR #2633, opened 2025-03-04
feat(bazel): support spec bundling with `rules_js`

devversion:
This is an initial, and still sufficient version of spec bundling for `rules_js`. Notably it's significantly shorter and very clean. Very convincing to see this short code snippet for a previously rather complicated piece.

We will continue to need spec bundling for ESM Karma tests. This is expected and even speeds up tests significantly, and improves stability.


josephperrott (approved):
LGTM

PR #2637, opened 2025-03-05
feat(ng-dev/release): support pnpm as primary package manager

devversion:
This commit adds support for working with `pnpm` as primary package manager. The Angular CLI repository will soon drop Yarn and this will be necessary then.

There is no need for testing of anything as we don't generally test any external commands, but in the future, when we drop Yarn completely here, we can update some other assertions to "recognize pnpm".


ng-dev/release/publish/pnpm-versioning.ts
```diff
@@ -0,0 +1,37 @@
...
+ * tool should automatically be able to satisfy those.
+ */
+export class PnpmVersioning {
+  async isUsingPnpm(repoPath: string) {
+    // If there is only a pnpm lock file at the workspace root, we assume pnpm
+    // is the primary package manager. We can remove such checks in the future.
+    return existsSync(join(repoPath, 'pnpm-lock.yaml')) && !existsSync(join(repoPath, 'yarn.lock'));
+  }
+
+  async getPackageSpec(repoPath: string) {
```
josephperrott, line 30: `getPackageSpec` as a name was not obviously to me what was being returned. Primarily because I asssociated spec with testing.

I don't know if there is a better naming option, but it was a bit confusing to me on first read.

josephperrott (approved):
LGTM overall, just the lint errors need to be corrected.

ng-dev/release/publish/pnpm-versioning.ts
devversion: Thanks for doing this!

devversion:
This PR was merged into the repository by commit 17e5969226628e7e92ac95cb6856b9509190c6ef.

PR #2653, opened 2025-03-10
fix(github-actions): pnpm detection accidentally matching `npm`

devversion:
The regex was accidentally matching `npm` out of `pnpm`.


josephperrott (approved):
LGTM

PR #2656, opened 2025-03-11
test(bazel): ensure shared chunks work properly in API golden testing

devversion:
This verifies that shared `.d.ts` chunks work properly in API golden testing. It's fine (and keeps the code simple) to just expect the shared definition to be repeated in the package entry-point goldens (it's still a single command to update)


josephperrott (approved):
LGTM

PR #2659, opened 2025-03-12
fix(ng-dev): check pnpm lock for latest `ng-dev` version when running release tool

devversion:
Currently we only check `yarn.lock`, but in e.g. CLI we already removed the yarn lock file.


josephperrott (approved):
LGTM

PR #2675, opened 2025-03-21
refactor(shared-scripts): Use `plugin-transform-async-generator-functions`

JeanMeche:
`plugin-proposal-async-generator-functions` is unmaintained and was replaced by `plugin-transform-async-generator-functions`


josephperrott (approved):
LGTM

PR #2688, opened 2025-03-27
feat(github-actions): allow a credential to be provided for configuring remote bazel setup

josephperrott:
Allow an input for a credential to provide authentication for RBE setup.


github-actions/bazel/configure-remote/action.yml
```diff
@@ -12,6 +12,14 @@ inputs:
     description: |
       Whether to allow remote execution to be configured for Windows. By default,
       Windows is only configured to leverage remote caching.
+  google_credential:
+    description: |
+      A Google credential to be used as authentication for RBE API usages.
+  upload_local_results:
```
devversion, line 18: It would probably also make sense to have a bazel configuration like `:trusted-builds` so that we can control more options like these in trusted builds?

devversion (approved):

github-actions/bazel/configure-remote/action.yml
josephperrott: Updated this to be `trusted_build` instead and set a config flag.

devversion (approved):

josephperrott:
This PR was merged into the repository by commit e3c0efecadda0e0fbb616abcdf447c788959ca64.

PR #2696, opened 2025-03-31
build: remove fix commit message label as it breaks with autosquash strategy

devversion:
This label and its merge script mode don't work properly with auto-squash. The label was initially built in the components repository; which doesn't use autosquashing.

We should not re-add the label automatically.


josephperrott (changes requested):
We need to modify the usages in `ng-dev/pr/common/validation/assert-allowed-target-label.ts` and `ng-dev/pr/merge/pull-request.ts`

devversion:
Wow, it does look like the build passes?! `createTypedObject` doesn't seem to be very typed :)

devversion:
Changed it to just not sync into FW.

josephperrott (approved):
LGTM

PR #2697, opened 2025-03-31
feat(bazel): support Karma spec bundling with `rules_js`

devversion:
We need AMD spec bundling for Karma, until we have our own IIFE-capable ruleset. This is a small layer on top of our much simpler spec bundling rule.

Note that we switch to `rules_esbuild` as otherwise Esbuild can escape sandbox due to its Go-based resolution.


bazel/spec-bundling/index_rjs.bzl
```diff
@@ -1,5 +1,5 @@
+load("@aspect_rules_esbuild//esbuild:defs.bzl", "esbuild")
 load("@devinfra//bazel/spec-bundling:spec-entrypoint.bzl", "spec_entrypoint")
-load("@devinfra_npm//bazel:esbuild/package_json.bzl", esbuild = "bin")
```
josephperrott, line 2: If we are removing esbuild usage here, can we removed it from the deps in `//bazel:package.json`? It seems to be now controlled via the workspace import of `aspect_rules_esbuild`

josephperrott (approved):

PR #2706, opened 2025-04-02
fix(github-actions): properly update mergeability status in early failures

josephperrott:
Previously during early failure cases, i.e. when the API request for the pull request validation tooling failed, the action would fail, but never update the mergeability status. This refactor attempts to ensure that we properly set a failing status in those situations.


devversion (approved):

josephperrott:
This PR was merged into the repository by commit a750a783454c0fff6c57fba0f4fefb642a10033e.

PR #2710, opened 2025-04-03
feat(bazel): support more options in spec-bundle rule

devversion:
Support `config` in `spec-bundle` rule. This is useful for supporting e.g. path mappings.


josephperrott (approved):
LGTM

PR #2724, opened 2025-04-09
feat(bazel): support `PORT` environment variable in http_server

devversion:
This will be useful for a more generic `server_test` rule.

### DIFF
--- a/bazel/http-server/main.ts
+++ b/bazel/http-server/main.ts
@@ -15,7 +15,11 @@ const {rootPaths, historyApiFallback, enableDevUi, environmentVariables, port, r
   process.argv.slice(2),
 )
   .strict()
-  .option('port', {type: 'number', default: 4200})
+  .option('port', {
+    type: 'number',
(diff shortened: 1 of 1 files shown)


bazel/http-server/main.ts
```diff
@@ -15,7 +15,11 @@ const {rootPaths, historyApiFallback, enableDevUi, environmentVariables, port, r
   process.argv.slice(2),
 )
   .strict()
-  .option('port', {type: 'number', default: 4200})
+  .option('port', {
+    type: 'number',
+    default: process.env['PORT'] !== undefined ? Number(process.env['PORT']) : 4200,
```
josephperrott, line 20: Do we want to do a `NaN` check as well?  Probably not strictly necessary, but figured I would ask.

josephperrott (approved):
LGTM

bazel/http-server/main.ts
devversion: I did consider it, but a runtime error is okay if this is off IMO; so will merge for nkw

PR #2727, opened 2025-04-10
refactor(ng-dev): add option for interactive rebase

thePunderWoman:
There are a lot of PRs that just need a commit message fix. This adds a -i flag so people can just interactively rebase to fix that for people quickly rather than dealing with a bunch of back and forth.

### DIFF
--- a/ng-dev/pr/rebase/cli.ts
+++ b/ng-dev/pr/rebase/cli.ts
@@ -15,16 +15,24 @@ import {rebasePr} from './index.js';
 /** The options available to the rebase command via CLI. */
 export interface RebaseOptions {
   pr: number;
+  i?: boolean;
 }
 
 /** Builds the rebase pull request command. */
 function builder(argv: Argv): Argv<RebaseOptions> {
-  return addGithubTokenOption(argv).positional('pr', {type: 'number', demandOption: true});
+  return addGithubTokenOption(argv)
+    .positional('pr', {type: 'number', demandOption: true})
+    .option('interactive', {
+      type: 'boolean',
+      alias: ['i'],
+      demandOption: false,
+      describe: 'Do the rebase interactively so that things can be squashed and amended',
+    });
 }
 
 /** Handles the rebase pull request command. */
-async function handler({pr}: Arguments<RebaseOptions>) {
-  process.exitCode = await rebasePr(pr);
+async function handler({pr, i}: Arguments<RebaseOptions>) {
+  process.exitCode = await rebasePr(pr, i);
 }
 
 /** yargs command module for rebasing a PR  */
--- a/ng-dev/pr/rebase/index.ts
+++ b/ng-dev/pr/rebase/index.ts
@@ -19,7 +19,7 @@ import {fetchPullRequestFromGithub} from '../common/fetch-pull-request.js';
  *
  * @returns a status code indicating whether the rebase was successful.
  */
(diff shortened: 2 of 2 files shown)


ng-dev/pr/rebase/cli.ts
```diff
@@ -15,16 +15,23 @@ import {rebasePr} from './index.js';
...
 /** Builds the rebase pull request command. */
 function builder(argv: Argv): Argv<RebaseOptions> {
-  return addGithubTokenOption(argv).positional('pr', {type: 'number', demandOption: true});
+  return addGithubTokenOption(argv)
+    .positional('pr', {type: 'number', demandOption: true})
+    .option('i', {
+      type: 'boolean',
+      demandOption: false,
+      describe: 'Do the rebase interactively so that things can be squashed and amended',
+    });
```
josephperrott, line 29: Can we have the option be the full word `interactive`, we can keep the `-i` as an alias flag.

```suggestion
    .option('interactive', {
      type: 'boolean',
      alias: ['i'],
      demandOption: false,
      describe: 'Do the rebase interactively so that things can be squashed and amended',
    });
```

josephperrott (approved):
LGTM

ng-dev/pr/rebase/cli.ts
thePunderWoman: Ooo yes, definitely. I didn't know you could do this in the tooling. :)

thePunderWoman: Done!

ng-dev/pr/rebase/index.ts
```diff
@@ -99,9 +99,18 @@ export async function rebasePr(prNumber: number): Promise<number> {
...
-    const [flags, env] = squashFixups
-      ? [['--interactive', '--autosquash'], {...process.env, GIT_SEQUENCE_EDITOR: 'true'}]
-      : [[], undefined];
+
+    let flags: string[] = [];
+    let env = undefined;
+
+    if (squashFixups || interactive) {
+      env = {...process.env, GIT_SEQUENCE_EDITOR: 'true'};
+      flags.push('--interactive');
```
devversion, line 108: do we always want to be interactive when squashing?

devversion (approved):

ng-dev/pr/rebase/index.ts
thePunderWoman: It was set to that before for autosquash and I just didn't want to break existing functionality.

josephperrott: Weirdly thats what we had previously

devversion: I think that's because the Git sequence editor was never opened. It was a "silent" interactive rebase to activate autosquashing (it doesn't work in a non-interactive rebase; or didn't back then at least)

thePunderWoman: That's a good question. I assumed it was needed, but I will try it locally and make sure.

thePunderWoman: Yeah, it is not needed. I'll put up a quick fix.

thePunderWoman: The fix is in #2728. I should have tested this before, but...I couldn't remember the command to run it locally. LOL

PR #2732, opened 2025-04-11
refactor(bazel): update integration test rule to use rules_js

devversion:
See individual commits


bazel/integration/test_runner/runner.mts
```diff
@@ -6,32 +6,31 @@
...
  */
 
-import * as fs from 'fs';
-import * as path from 'path';
-import * as tmp from 'tmp';
-import * as os from 'os';
+import fs from 'node:fs/promises';
+import path from 'path';
+import os from 'os';
+import {mkdtemp} from 'fs/promises';
```
josephperrott, line 12: Shouldn't we just be using `fs.mkdtemp` at the callsite?

bazel/integration/test_runner/runner.mts
```diff
@@ -6,32 +6,31 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import * as fs from 'fs';
-import * as path from 'path';
-import * as tmp from 'tmp';
-import * as os from 'os';
+import fs from 'node:fs/promises';
+import path from 'path';
+import os from 'os';
```
josephperrott, line 11:
```suggestion
import path from 'node:path';
import os from 'node:os';
```

josephperrott (approved):
LGTM

PR #2739, opened 2025-04-15
refactor(ci): add `--noEmit` to lint build

dgp1130:
Without this, the lint job actually generates a bunch of in-tree `*.js` files. This creates a bunch of noise for developers who blindly copy-past the command to debug an issue and also risks future commands in this job running against those `*.js` files, which likely wouldn't be intended.


josephperrott (approved):
LGTM

PR #2750, opened 2025-04-23
fix(ng-dev): allow for a retry during the validation that the commit sha being released is the expected commit

josephperrott:


ng-dev/release/publish/actions.ts
```diff
@@ -759,13 +730,45 @@ export abstract class ReleaseAction {
...
-      ...this.git.remoteParams,
-      ref: commitSha,
-    });
-    return data.commit.message.startsWith(getCommitMessageForRelease(version));
+  /**
+   * Retreive the latest commit from the provided branch, and verify that it is the expected
+   * release commit and is the direct child of the previous sha provided.
+   *
+   * The method will make one recursive attempt to check again before throwing an error if
+   * any error occurs during this validation.
```
devversion, line 738: Can we add context on when this can happen?

devversion (approved):

PR #2754, opened 2025-04-29
fix(apps): set up redirect from material.angular.io (and all subdomains) to material.angular.dev

josephperrott:


apps/functions/dns-redirecting/index.ts
```diff
@@ -46,6 +46,15 @@ export const dnsRedirecting = functions.https.onRequest(
     if (hostname === 'blog.angular.io') {
       response.redirect(redirectType, `https://blog.angular.dev${request.originalUrl}`);
     }
+    if (hostname === 'material.angular.io') {
+      response.redirect(redirectType, `https://material.angular.dev${request.originalUrl}`);
+    }
+    if (hostname.endsWith('material.angular.io')) {
```
devversion, line 52: what would be some of these sub-sub-domains?

devversion (approved):

apps/functions/dns-redirecting/index.ts
devversion: Ah, didn't even think about that. SGTM (nit: might want to try `endsWith('.material...')`

josephperrott:
This PR was merged into the repository by commit 14c3d6bd2fa5c3231be7bd4b3c0bba68c9d79e94.

PR #2756, opened 2025-04-30
a couple fixes for dns redirecting

josephperrott:
Previously the response redirect would complete the http response, but then we would attempt to still provide a status response and message, which caused errors throughout our logging.  The correct action is to no longer perform response actions after a redirect.


clydin (approved):

josephperrott:
This PR was merged into the repository by commit a4538b2474d3c551f0217c3d1f5f3a99cf4f8eb7.

PR #2772, opened 2025-05-08
feat(ng-dev): update `renovate.json` baseBranches when creating a new branch

alan-agius4:
Adds support for automatically modifying `renovate.json` to include newly created branches in the `baseBranches` array, when `updateRenovateConfig` is enabled.

### DIFF
--- a/ng-dev/release/publish/actions/renovate-config-updates.ts
+++ b/ng-dev/release/publish/actions/renovate-config-updates.ts
@@ -1,0 +1,41 @@
+import {existsSync} from 'node:fs';
+import {green, Log} from '../../../utils/logging.js';
+import {join} from 'node:path';
+import {writeFile, readFile} from 'node:fs/promises';
+
+/**
+ * Updates the `renovate.json` configuration file to include a new base branch.
+ *
+ * @param projectDir - The project directory path.
+ * @param newBranchName - The name of the new branch to add to the base branches list.
+ * @returns A promise that resolves to an string containing the path to the modified `renovate.json` file,
+ *          or null if config updating is disabled.
+ */
+export async function updateRenovateConfig(
+  projectDir: string,
+  newBranchName: string,
+): Promise<string | null> {
+  const renovateConfigPath = join(projectDir, 'renovate.json');
+  if (!existsSync(renovateConfigPath)) {
+    Log.warn(`  ✘   Skipped updating Renovate config as it was not found.`);
+
+    return null;
+  }
+
+  const config = await readFile(renovateConfigPath, 'utf-8');
+  const configJson = JSON.parse(config) as Record<string, unknown>;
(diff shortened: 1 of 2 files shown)


ng-dev/release/publish/actions.ts
```diff
@@ -140,6 +140,43 @@ export abstract class ReleaseAction {
...
+      spinner.success(
+        yellow('Skipped updating Renovate config: "baseBranches" must contain exactly 2 branches.'),
+      );
+
+      return null;
+    }
+
+    configJson.baseBranches = ['main', newBranchName];
+    await fs.writeFile(renovateConfigPath, JSON.stringify(configJson, undefined, 2));
+    spinner.success(green(' Updated Renovate config'));
```
devversion, line 175: Super optional nit: not sure a spinner is useful here if it's just file updating. A normal log message would probably integrate better into the existing output.

ng-dev/release/publish/actions.ts
```diff
@@ -140,6 +140,43 @@ export abstract class ReleaseAction {
...
+   * Updates the `renovate.json` configuration file to include a new base branch.
+   *
+   * This is used when `rulesJsInteropMode` is enabled to ensure the Renovate
+   * tool considers both the main branch and a new feature or update branch.
+   *
+   * @param newBranchName - The name of the new branch to add to the base branches list.
+   * @returns A promise that resolves to an string containing the path to the modified `renovate.json` file,
+   *          or null if config updating is disabled.
+   */
+  protected async updateRenovateConfig(newBranchName: string): Promise<string | null> {
```
devversion, line 153: Do you mind moving that to a file like `renovate-config-updates.ts`?

ng-dev/release/publish/actions.ts
```diff
@@ -140,6 +140,43 @@ export abstract class ReleaseAction {
...
+
+      return null;
+    }
+
+    const config = await fs.readFile(renovateConfigPath, 'utf-8');
+    const configJson = JSON.parse(config) as Record<string, unknown>;
+    const baseBranches = configJson.baseBranches;
+    if (!Array.isArray(baseBranches) || baseBranches.length !== 2) {
+      spinner.success(
+        yellow('Skipped updating Renovate config: "baseBranches" must contain exactly 2 branches.'),
```
devversion, line 167: Aren't there cases where we also want to update `patch`, `rc` and `main`?

alan-agius4: Good question. I think once the feature freeze begins, changes to the patch branch should be kept minimal to maintain stability, especially since it's heading to LTS soon.

Also, running Renovate on three branches would significantly increase runtime not to mention PRs will take longer to open due to rate limits.

devversion: Fair point

devversion (approved):

PR #2773, opened 2025-05-09
fix(github-actions): delete obsolete branches and close superseded PRs

alan-agius4:
When using the `create-pr-for-changes` workflow, old branches are not being deleted, and the content of superseded PRs remains open. This change ensures that:
- Obsolete branches are deleted after PR creation.
- Previously created PRs with outdated content are automatically closed.

Example: https://github.com/angular/angular/pull/61203

### DIFF
--- a/github-actions/create-pr-for-changes/lib/main.ts
+++ b/github-actions/create-pr-for-changes/lib/main.ts
@@ -129,6 +129,7 @@ async function main(): Promise<void> {
     const {
       data: {items: supersededPrs},
     } = await git.github.search.issuesAndPullRequests({
+      advanced_search: 'true',
       q: toGithubSearchQuery({
         repo: `${repo.owner}/${repo.name}`,
         type: 'pull-request',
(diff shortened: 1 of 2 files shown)


github-actions/create-pr-for-changes/lib/main.ts
```diff
@@ -129,6 +129,7 @@ async function main(): Promise<void> {
     const {
```
alan-agius4, line 129: Note: We could simplify this logic, if we keep re-using the same branch (like renovate)

github-actions/create-pr-for-changes/lib/main.ts
```diff
@@ -129,6 +129,7 @@ async function main(): Promise<void> {
...
       data: {items: supersededPrs},
     } = await git.github.search.issuesAndPullRequests({
+      advanced_search: 'true',
```
devversion, line 132: Yeah, when I read the commit description, I remembered that we already had logic for this in place. Is this just a fix because it didn't work?

devversion (approved):

github-actions/create-pr-for-changes/lib/main.ts
alan-agius4: Yes

PR #2774, opened 2025-05-09
ci: use corepack in ng-renovate workflow

alan-agius4:
This allow the repos to have different versions of pnpm. Potentially also in different branches

### DIFF
--- a/.github/workflows/ng-renovate.yml
+++ b/.github/workflows/ng-renovate.yml
@@ -28,8 +28,10 @@ jobs:
       # this step.
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: ./github-actions/npm/checkout-and-setup-node
-      - run: npm install --global pnpm@9.15.6
-        shell: bash
+      - name: Corepack install and enable
+        run: |
+          npm install corepack -g
+          corepack enable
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ng-renovate.yml
```diff
@@ -28,8 +28,10 @@ jobs:
       # this step.
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: ./github-actions/npm/checkout-and-setup-node
-      - run: npm install --global pnpm@9.15.6
-        shell: bash
+      - name: Corepack install and enable
```
josephperrott, line 31: Corepack is being dropped from node which is why we didn't use corepack for this initially.  Is this still a concern for us?

alan-agius4: Corepack is being dropped from core, but it is not deprecated and will still be available as a standalone package.

alan-agius4: PS: this change is needed as components and cli and FW uses different patch versions of PNPM 9 and causes the Renovate to fail https://github.com/angular/components/pull/31062#issuecomment-2866678190

josephperrott: 👍

PR #2784, opened 2025-05-12
ci: ignore `yarn` & `bazel` updates

alan-agius4:
`yarn` is copied locally

### DIFF
--- a/renovate-presets/default.json5
+++ b/renovate-presets/default.json5
@@ -28,7 +28,8 @@
   // Ignored dependencies in all repositories
   "ignoreDeps": [
     "build_bazel_rules_nodejs",
-    "rules_pkg"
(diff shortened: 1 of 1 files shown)


renovate-presets/default.json5
```diff
@@ -128,6 +129,7 @@
       "enabled": false,
       "matchPackageNames": [
         "@types/node",
+        "bazel",
```
alan-agius4, line 132: I am assuming this is fine to update minor and patch

devversion: Yeah, also we won't have many `@bazel` deps in the future anymore

devversion (approved):

alan-agius4:
Do not merge this yet.

PR #2790, opened 2025-05-12
fix(github-actions): move `angular/**` to `matchPackageNames`

alan-agius4:
This change ensures that `angular/dev-infra` update in WORKSPACE files is grouped with `cross-repo angular dependencies` instead of `bazel` to reduce the amount of PRs opened.

### DIFF
--- a/renovate-presets/default.json5
+++ b/renovate-presets/default.json5
@@ -3,6 +3,7 @@
   extends: ['group:monorepos'],
   dependencyDashboard: true,
   rangeStrategy: 'replace',
+  automerge: false,
 
   // Schedule Renovate to run during off-peak hours
   schedule: ['after 10:00pm every weekday', 'before 5:00am every weekday', 'every weekend'],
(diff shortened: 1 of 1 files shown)


devversion (approved):

renovate-presets/default.json5
```diff
@@ -60,13 +66,8 @@
...
       schedule: ['at any time'],
-      matchDepNames: [
-        '@angular-devkit/**',
-        '@angular/**',
-        '@schematics/**',
-        'angular/**',
-        'ng-packagr',
-      ],
+      matchDepNames: ['@angular-devkit/**', '@angular/**', '@schematics/**', 'ng-packagr'],
+      matchPackageNames: ['angular/**'],
```
devversion, line 70: What's the difference between both options? is this because there is no display/dep name for the actions?

alan-agius4: The `matchDepNames` uses the dependency name, in case of a WORKSPACE, this is the `name` property of the toolchain. `matchPackageName` uses the GitHub org/repo.

devversion: Gotcha, thx for the insight

PR #2796, opened 2025-05-14
fix(apps): respond to request for m.aio/assets/versions.json directly

josephperrott:
Rather than respond with a redirect for the versions information on m.aio, respond with the actual JSON string that would be responded at the destination url to prevent CORS issues.

### DIFF
--- a/apps/functions/dns-redirecting/index.ts
+++ b/apps/functions/dns-redirecting/index.ts
@@ -2,6 +2,7 @@ import * as functions from 'firebase-functions';
 
 export const dnsRedirecting = functions.https.onRequest(
   {
+    cors: true,
     invoker: 'public',
     timeoutSeconds: 5,
     minInstances: 1,
@@ -42,6 +43,13 @@ export const dnsRedirecting = functions.https.onRequest(
     } else if (hostname === 'blog.angular.io') {
       response.redirect(redirectType, `https://blog.angular.dev${request.originalUrl}`);
     } else if (hostname === 'material.angular.io') {
+      if (request.originalUrl === '/assets/versions.json') {
+        const versionsData = await fetch('https://material.angular.dev/assets/versions.json').then(
+          (r) => r.json(),
+        );
+        response.set('Content-Type', 'application/json').send(JSON.stringify(versionsData));
+        return;
+      }
       response.redirect(redirectType, `https://material.angular.dev${request.originalUrl}`);
     } else if (hostname.endsWith('.material.angular.io')) {
       response.redirect(
--- a/yarn.lock
+++ b/yarn.lock
@@ -2589,30 +2589,7 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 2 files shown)


apps/functions/dns-redirecting/index.ts
```diff
@@ -42,6 +43,17 @@ export const dnsRedirecting = functions.https.onRequest(
     } else if (hostname === 'blog.angular.io') {
       response.redirect(redirectType, `https://blog.angular.dev${request.originalUrl}`);
     } else if (hostname === 'material.angular.io') {
+      if (request.originalUrl === '/assets/versions.json') {
+        const versionsData = await fetch('https://material.angular.dev/assets/versions.json').then(
+          (r) => r.json(),
+        );
+        const cacheDuration = 60 * 60 * 12;
```
devversion, line 50: In `firebase.json` we have:

```
            "key": "Cache-Control",
            // The versions file should not be cached as it may change
            // at any time and otherwise old versions would be rendered.
            "value": "no-cache
```

Should we do that here as well? I don't think we'd want to cache?

josephperrott: My thinking was that we did want to cache this since it is called on literally every first call of a m.aio page. Since this is supposed to be the redirect service I just didn't want it to cause blocking too often.

josephperrott: Okay, I'll do no cache for now and I will check in on if we see queuing or overly long requests for the dns redirection service and we can revisit if that comes up.

apps/functions/dns-redirecting/index.ts
```diff
@@ -42,6 +43,13 @@ export const dnsRedirecting = functions.https.onRequest(
     } else if (hostname === 'blog.angular.io') {
       response.redirect(redirectType, `https://blog.angular.dev${request.originalUrl}`);
     } else if (hostname === 'material.angular.io') {
+      if (request.originalUrl === '/assets/versions.json') {
+        const versionsData = await fetch('https://material.angular.dev/assets/versions.json').then(
+          (r) => r.json(),
+        );
+        response.set('Content-Type', 'application/json').send(JSON.stringify(versionsData));
```
devversion, line 50: Nit: Might be worth setting the explicit response no-cache header as well. Not sure what the default is. LGTM though.

devversion (approved):

apps/functions/dns-redirecting/index.ts
josephperrott: According to the [documentation](https://firebase.google.com/docs/hosting/manage-cache), dynamic content (all function responses) are not cached by default.

josephperrott:
This PR was merged into the repository by commit ac6d1631beb9e605769b739a5349e178c1f46ec0.

apps/functions/dns-redirecting/index.ts
devversion: This would be more about the browser caching it incorrectly, given no cache control header

PR #2800, opened 2025-05-15
fix(github-action): replace `action: review` with `action: merge` in renovate preset

alan-agius4:
One less thing to do for users.


devversion (approved):
Ty!

PR #2808, opened 2025-05-16
refactor: remove docker sys_admin capability

devversion:
Separate branch for framework, as we otherwise can't update build-tooling during migration.


devversion:
actually nvm. not sure why GitHub shows this as valid

PR #2827, opened 2025-05-20
fix(bazel): the "paths[0]" argument must be of type string. Received an instance of Buffer

alan-agius4:
Workaround for
```
TypeError [ERR_INVALID_ARG_TYPE]: The "paths[0]" argument must be of type string. Received an instance of Buffer
    at Object.resolve (node:path:1101:7)
    at fs.readdir (internal/node/node_patches.cjs:257:44)
```


devversion:
Closing as we shouldn't need this.

PR #2850, opened 2025-06-05
feat(ng-dev): update target labels in renovate config

alan-agius4:
This commit enhances the release tool to dynamically adjust Renovate labels based on the current version.

Example:
https://github.com/angular/angular/pull/61894


ng-dev/release/publish/actions/renovate-config-updates.ts
```diff
@@ -34,8 +36,92 @@ export async function updateRenovateConfig(
...
 }
+
+/**
+ * Updates a specific target label in the `renovate.json` configuration file.
+ * This function specifically targets and replaces one label with another within the `packageRules`.
+ *
+ * @param projectDir - The path to the project directory.
+ * @param fromLabel - The label name to be replaced.
+ * @param toLabel - The new label name to replace `fromLabel` with.
+ * @returns A promise that resolves to `true` if a label was updated and the file was written `false` otherwise.
```
josephperrott, line 58: It looks like this doesn't resolve to `true` or `false` but instead to either `string` or `null`

alan-agius4: Good catch, I forgot to update the comment. Will do it tomorrow

PR #2869, opened 2025-06-23
fix(ng-dev): avoid exposing the API key in help text

crisbeto:
The API key from the environment was set as the default in `yargs` which means that it was being logged when the user doesn't type out the command properly.


josephperrott (approved):
LGTM

josephperrott:
This PR was merged into the repository by commit db18d7175d759fdfb3e40d363a1579621d1fcffd.

PR #2870, opened 2025-06-24
fix(bazel): add support for defining the platform used for bundling `spec_bundle` targets

josephperrott:
Previously the platform was hardcoded to node which worked by chance, but in some cases the platform should be explicitly set to browser for web test which will run in a browser


devversion (approved):

josephperrott:
This PR was merged into the repository by commit 92edde46390984cf9182d3d1e5c4c7651f75171f.

PR #2879, opened 2025-06-27
refactor(bazel): use a ts_project instead of ts_library for providing the driver-utilities

josephperrott:


bazel/benchmark/driver-utilities/BUILD.bazel
```diff
@@ -1,19 +1,26 @@
-load("//bazel:defaults.bzl", "ts_library")
+load("@aspect_rules_ts//ts:defs.bzl", "ts_project")
 
 package(default_visibility = ["//visibility:public"])
 
-ts_library(
+ts_project(
```
devversion, line 5: I'd suggest we move the driver utilities into FW `modules/`

josephperrott: I can do that instead.

PR #2908, opened 2025-07-24
fix(bazel/spec-bundling): allow for non init suffixed files to be used as bootstrap files

josephperrott:


bazel/spec-bundling/spec-entrypoint.bzl
```diff
@@ -71,7 +71,10 @@ def _spec_entrypoint_impl(ctx):
     # Note: `to_list()` is an expensive operation but we need to do this for every
     # dependency here in order to be able to filter out spec files from depsets.
     direct_bootstrap_files = depset(transitive = bootstrap_direct_deps).to_list()
-    bootstrap_files = _filter_files(direct_bootstrap_files, ["init"])
+
+    # Filter to only js and mjs files, we can't always define what the name of the files
+    # needed for bootstrapping are so we rely on the user to provide the exact list.
+    bootstrap_files = _filter_files(direct_bootstrap_files, [""])
```
devversion, line 77: IIRC this was a risk for some unintended files to be picked up, but not sure anymore.

Isn't this logic the same and should continue working even with RJS?

LGTM to unblock as it generally seems reasonable

devversion (approved):

PR #2912, opened 2025-07-29
fix(bazel): extract types from ts_project targets instead of ts_library

josephperrott:
Look for types and transitive types in JsInfo instead of DeclarationInfo providers.


devversion (approved):

josephperrott:
This PR was merged into the repository by commit 39c70c3c8c63165bc2df0efc7dcb3b3287b69685.

PR #2918, opened 2025-08-01
ci: enable pnpm updates

alan-agius4:
With pnpm version 10, pnpm will download the appropriate version and thus no longer require to pin the version of pnpm in all repositories to that of the renovate GitHub action.


josephperrott (approved):
LGTM

PR #2922, opened 2025-08-04
refactor(ng-dev): use pnpm instead of yarn

JeanMeche:
This is for the build-and-link command


josephperrott (approved):
LGTM

JeanMeche:
The problem when you don't test your changes is that sometimes they don't work. 😄

PR #2950, opened 2025-08-11
fix(ng-dev): downgrade `@google-cloud/spanner` to `8.0.0`

alan-agius4:
See: https://github.com/googleapis/nodejs-spanner/issues/2375


josephperrott (approved):
LGTM

PR #2969, opened 2025-08-14
fix(bazel): relax node.js check in integration checks

alan-agius4:
This error currently breaks components.


devversion (approved):
I don't know where this comes from, but LGTM to unblock

PR #2975, opened 2025-08-19
feat(ng-dev): handle updating the `MODULE.bazel.lock` file during a release

alan-agius4:
This lock has contains hashes of the package.json file. Thus it needs to be updated once the version is updating during the release.

### DIFF
--- a/ng-dev/release/publish/external-commands.ts
+++ b/ng-dev/release/publish/external-commands.ts
@@ -7,7 +7,6 @@
  */
 
 import semver from 'semver';
-
 import {ChildProcess, SpawnResult, SpawnOptions} from '../../utils/child-process.js';
 import {Spinner} from '../../utils/spinner.js';
 import {NpmDistTag} from '../versioning/index.js';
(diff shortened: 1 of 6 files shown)


ng-dev/release/publish/external-commands.ts
```diff
@@ -263,6 +262,25 @@ export abstract class ExternalCommands {
...
   }
 
+  /**
+   * Invokes the `bazel mod deps --lockfile_mode=update` command in order
+   * to refresh `MODULE.bazel.lock` file.
+   */
+  static async invokeBazelModDepsUpdate(projectDir: string): Promise<void> {
+    const spinner = new Spinner('Updating "MODULE.bazel.lock"');
+    try {
+      await ChildProcess.spawn(getBazelBin(), ['mod', 'dep', '--lockfile_mode=update'], {
```
alan-agius4, line 272: I wonder if we can force a partial update of the lock file, instead of a full refresh. 

IE just `@://package.json`, I’ll try a bit more tomorrow.

PR #2978, opened 2025-08-20
ci: update `MODULE.bazel.lock` and refine update strategy

alan-agius4:
Adds a Renovate postUpdateTask to automatically update `MODULE.bazel.lock` and resolves an outstanding issue with Bazel lockfile support.

Reduces dependency updates on non-main branches to only Bazel, GitHub Actions, and cross-repository updates. This change aims to reduce friction and improve the caretaker workflow.


josephperrott (approved):
LGTM

PR #2989, opened 2025-08-22
feat(github-actions): cache nvm installations between runs

alan-agius4:
This commit introduces a caching mechanism for NVM installations within the GitHub Actions workflow.


josephperrott (approved):
LGTM

PR #2990, opened 2025-08-22
ci: add bazel lock file update command

alan-agius4:
Adds the bazel lock file update command to post update task


josephperrott (approved):
LGTM

PR #2999, opened 2025-08-25
fix(bazel): properly handle highlighter.js in strict deps tests

josephperrott:


bazel/ts_project/strict_deps/index.mts
```diff
@@ -52,16 +52,18 @@ function checkPathsForMatch(moduleSpecifier: string, paths?: ts.MapLike<string[]
   return false;
 }
 
+/** The list of known package names that end with extensions and need to be special cased. */
+const knownModuleSpecifiersWithExtensions = ['highlight.js', 'zone.js'];
```
alan-agius4, line 56: NIT: consider a Set for constant time.

alan-agius4 (approved):
LGTM.

PR #3002, opened 2025-08-26
ci: disable updating of `@google-cloud/spanner`

alan-agius4:
This dependency had a missing dependency for sometime, which now has been addressed however  it's transtive dependency `@opentelemetry/api` now has missing depedencies.


josephperrott (approved):
LGTM

PR #3011, opened 2025-08-27
fix(github-actions): add an additional retry on failed attempts to set the merge status

josephperrott:
Add an additional retry if the branch manager fails to set the mergeability status on a pull request, after a short timeout.


thePunderWoman (approved):
Huge LGTM for this. Thanks, @josephperrott!

PR #3013, opened 2025-08-28
ci: change when `postUpgradeTasks` are invoked

alan-agius4:
Since the `package.json` file is no longer included in the Bazel lock file, the `postUpgradeTasks` are only necessary when the `matchManagers` are 'bazel', 'bazel-module', or 'bazelisk' for the majority of our repos.

For the dev-infra and Angular framework repositories, the `Angular framework repositories, the postUpgradeTasks` should be executed when the managers are 'bazel', 'bazel-module', 'bazelisk', or 'npm' due to them having generated checked-in files.

### DIFF
--- a/renovate-presets/default.json5
+++ b/renovate-presets/default.json5
@@ -36,26 +36,25 @@
     'yarn', // Yarn is copied locally in all repositories where needed.
   ],
 
-  // Renovate does not update Bazel lockfile for the time being.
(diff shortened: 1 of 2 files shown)


josephperrott:
This PR was merged into the repository by commit c0d97dbb3d3a1d20be56d3db5cdf3ad82547c0ac.

PR #3023, opened 2025-08-29
ng-renovate workflow changes

alan-agius4:
**feat(github-actions): add cache-dependency-path to checkout-and-setup-node**
This allows specifying a dependency file path for more accurate caching, aligning with the official `setup-node` action's capabilities. The `ng-renovate` workflow is updated to use this new input.

See: https://github.com/actions/setup-node?tab=readme-ov-file#caching-global-packages-data

---

**ci: remove `pnpm/action-setup` set from `ng-renovate` workflow**
This step is already included in `./github-actions/npm/checkout-and-setup-node` which makes this duplicate.


josephperrott (approved):
LGTM

PR #3024, opened 2025-08-29
feat(actions): use package-manager-cache input in checkout-and-setup-node

alan-agius4:
The `cache` input in `actions/setup-node` is deprecated and will be removed in a future version. This commit updates the `checkout-and-setup-node` action to use the new `package-manager-cache` input instead.

The default value of `disable-package-manager-cache` has also been changed to `true` to reflect the new default behavior of `actions/setup-node`.


josephperrott (approved):
LGTM

alan-agius4:
Actually this has not yet been released upstream. So I'll block it for now.

PR #3042, opened 2025-09-10
fix(ng-dev): run `git clean` after `git checkout` during the release

alan-agius4:
This is needed as in some cases, `node_modules` can be left around and in old bazel versions can cause an infinite symlink expansion


josephperrott (changes requested):

josephperrott (approved):
LGTM

alan-agius4:

- main: ed31a415fcf0ffe88185a6480c7f5b8e6d42ec0b

PR #3127, opened 2025-10-03
test: update all integration tests to use pnpm instead of yarn

alan-agius4:
Remove `yarn` and replace it with pnpm


devversion (approved):
LGTM

Reviewed-for: global-dev-infra-approvers

PR #3169, opened 2025-10-22
feat(ng-dev): add conditional autosquash merge strategy

alan-agius4:
Introduces a new conditional autosquash merge strategy. This strategy uses the autosquash merge strategy if the pull request contains fixup or squash commits, and the GitHub API merge strategy otherwise.

This allows pull requests that do not need to be squashed to be closed as "merged" on GitHub, rather than "closed".

### DIFF
--- a/ng-dev/pr/config/index.ts
+++ b/ng-dev/pr/config/index.ts
@@ -79,8 +79,7 @@ export function assertValidPullRequestConfig<T extends NgDevConfig>(
     );
   }
 
-  const {githubApiMerge} = config.pullRequest;
-  if (githubApiMerge === undefined) {
+  if (config.pullRequest.githubApiMerge === undefined) {
     errors.push('No explicit choice of merge strategy. Please set `githubApiMerge`.');
   }
 


ng-dev/pr/config/index.ts
```diff
@@ -36,6 +36,13 @@ export interface PullRequestConfig {
   /** List of statuses that are required before a pull request can be merged. */
   requiredStatuses?: {type: 'check' | 'status'; name: string}[];
 
+  /**
+   * Whether pull requests should be merged using a conditional autosquash strategy.
+   * If a pull request contains fixup or squash commits, the autosquash strategy
+   * will be used. Otherwise, the Github API merge strategy will be used.
+   */
+  conditionalAutosquashMerge?: boolean;
```
josephperrott: Would it make more sense to actually just add something like `rebase-with-fixups`  to the github API strategy? 

Or just change the `rebase` merge method to mean conditional autosquash and only have one "strategy"?

ng-dev/pr/config/index.ts
```diff
@@ -76,7 +79,8 @@ export function assertValidPullRequestConfig<T extends NgDevConfig>(
     );
   }
 
-  if (config.pullRequest.githubApiMerge === undefined) {
+  const {githubApiMerge} = config.pullRequest;
+  if (githubApiMerge === undefined) {
```
josephperrott:
```suggestion
  if (config.pullRequest.githubApiMerge === undefined) {
```

I think we can just go back to the check we had before right?

PR #3193, opened 2025-10-27
feat(github-actions): add auto labeling by path to github labeler action

thePunderWoman:
This adds the ability to add additional labels automatically if a specific path is matched in the pull request commits.

### DIFF
--- a/github-actions/pull-request-labeling/action.yml
+++ b/github-actions/pull-request-labeling/action.yml
@@ -5,6 +5,12 @@ inputs:
   angular-robot-key:
     description: 'The private key for the Angular Robot Github app.'
     required: true
+
 runs:
-  using: 'node20'
-  main: 'main.js'
+  using: composite
+  steps:
+    - name: Initialize environment
+      uses: ./github-actions/npm/checkout-and-setup-node
+    - name: Apply labels to pull requests
+      run: node $GITHUB_ACTION_PATH/main.js
+      shell: bash


ng-dev/pr/config/index.ts
```diff
@@ -58,6 +58,11 @@ export interface PullRequestConfig {
    */
   validators?: PullRequestValidationConfig;
 
+  /**
+   * Optional map of paths that have extra labels to apply when PRs are created.
+   */
+  managedLabelByPath?: {[path: string]: string[]};
```
josephperrott: I think the key and value should be flipped here so that we can more efficiently find matches?

I think we should have
```
{label: string]: string[]}
```

Its much more common that we would like to apply the same label to multiple different matching paths, than wanting multiple labels on a specific path.

github-actions/pull-request-labeling/lib/main.ts
```diff
@@ -17,7 +20,9 @@ class PullRequestLabeling {
     const token = await getAuthTokenFor(ANGULAR_ROBOT);
     const git = new Octokit({auth: token});
     try {
-      const inst = new this(git);
+      const config = await getConfig();
+      assertValidPullRequestConfig(config);
```
josephperrott:
```suggestion
      const config = await getConfig([assertValidPullRequestConfig]);
```

PR #3230, opened 2025-11-11
fix(ng-dev): properly remove reverted commits during release note generation

josephperrott:
When we are compiling the list of the commits that should be included for release notes, we now remove the commits which have been identified as reverted based on later commits in the release commmits range.


josephperrott:

- main: b14d0b01760280c36f6fcfdd31f120a465a56501

PR #3248, opened 2025-11-14
Create tooling around `merge-mode` usage

josephperrott:
See individual commits.

Overall, these changes create the tooling necessary for us to change the `merge-mode` custom property setting to control the merging permissions for our repositories.


devversion (approved):

josephperrott:
@devversion PTAL, had to do a bit of an overhaul

PR #3502, opened 2026-03-11
fix(ng-dev): ensure node version check uses project directory after nvm install

alan-agius4:
Sets the working directory to the project directory and enables silent mode when retrieving the Node.js version. This ensures that the logged version correctly reflects project-specific configurations like those in an `.nvmrc` file.


josephperrott (approved):
LGTM

PR #3504, opened 2026-03-11
ci: skip non-affected snapshots in the publish workflow

alan-agius4:
Add usage of `--skip-non-affected-snapshots`


josephperrott (approved):
LGTM

PR #3514, opened 2026-03-12
ci: publish bazel rules snapshots

alan-agius4:
This commit integrates rules_angular, rules_browsers, and rules_sass into the repository's build and snapshot publishing workflow.

### DIFF
--- a/.ng-dev/release.mjs
+++ b/.ng-dev/release.mjs
@@ -5,6 +5,18 @@ export const release = {
       name: '@angular/ng-dev',
       snapshotRepo: 'dev-infra-private-ng-dev-builds',
     },
+    {
+      name: 'rules_angular',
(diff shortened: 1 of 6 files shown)


.ng-dev/release.mjs
```diff
@@ -5,6 +5,18 @@ export const release = {
       name: '@angular/ng-dev',
       snapshotRepo: 'dev-infra-private-ng-dev-builds',
     },
+    {
+      name: 'rules_angular',
+      snapshotRepo: 'rules_angular',
```
alan-agius4: Can you change the settings in these repos please to the same as the other snapshot repos?

IE: disallow PRs, disable issues, branch protection settings etc..

josephperrott (approved):
LGTM

PR #3520, opened 2026-03-12
ci: add conditional setup steps for UV and Dart to the ng-renovate workflow

alan-agius4:
- UV is only needed for web-codegen-scorer as it uses python.
- Dart is only needed for rules_sass which is in dev-infra.


josephperrott (approved):
LGTM

PR #3529, opened 2026-03-13
fix(bazel/rules_angular): fix 'NgtscCompilerHost' type incompatibility

alan-agius4:
This change fixes an issue where the 'NgtscCompilerHost' was not assignable to 'CompilerHost'
due to incompatible 'SourceFile' types from different TypeScript versions.

The fix involves ensuring that the Angular compiler worker uses the user-configured
TypeScript dependency instead of a potentially different version from the root.
As part of this, the 'typescript-local' dependency alias is renamed to 'typescript'
and the symlinking logic is moved to the worker package.

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -547,7 +547,7 @@
     "@@rules_angular+//setup:extensions.bzl%rules_angular": {
       "general": {
         "bzlTransitiveDigest": "fkaH7HMicL3g7/NDaFzlq39kcLopMyQ3KdbDn+5CRzA=",
-        "usagesDigest": "7iXpy97hRtKmmBzd+nMFAwTGsF2in10kXImaLMA/Muk=",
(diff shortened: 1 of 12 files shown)


josephperrott (approved):
LGTM

We should keep an eye out to make sure this behaves as expected once its used elsewhere, it looks correct, but I know we have also run into alot of weirdness/edges with having the user provided versions of the packages coupled with multiple instances of the `rules_angular` module being set up within repositories.

alan-agius4:
> LGTM
> 
> We should keep an eye out to make sure this behaves as expected once its used elsewhere, it looks correct, but I know we have also run into alot of weirdness/edges with having the user provided versions of the packages coupled with multiple instances of the `rules_angular` module being set up within repositories.

For context, the issue was that angular/angular was broken. I did link it and tested it in both patch and main prior to opening the PR and it does seem to behave as expected.

https://github.com/angular/angular/pull/67677
https://github.com/angular/angular/pull/67681

PR #3534, opened 2026-03-13
fix(ng-dev/release): allow empty commits when publishing snapshots

alan-agius4:
When publishing snapshots, it's possible that a package is built but 
contains no functional changes (e.g. only version placeholder updates 
which are ignored by the diff). 
Previously, the git commit step would fail in these cases because the 
working tree was clean. This change adding the --allow-empty flag to 
ensure the commit succeeds regardless of changes.


josephperrott (changes requested):
Rather than selecting running the `git commit` command we should `--allow-empty` for cases where we do want it to publish regardless of whether its changed.

For instance in angular/angular, we always want it to publish since we want to have a snapshot of all of the packages, not just the changed ones for a given commit.

josephperrott (approved):
LGTM

PR #3535, opened 2026-03-13
fix(github-actions): use stable TMPDIR for rules_sass compiler compilation

alan-agius4:
Forcing a stable temporary directory prevents the Dart compiler on macOS from embedding random absolute paths into the compiled binaries.
    
Previously, the random temp paths (e.g., /var/folders/.../random_hash/snapshot.aot) caused the binaries to differ on every build run even when the source code was identical. Using a workspace-relative TMPDIR ensures deterministic binary output.


josephperrott (approved):
LGTM

PR #3547, opened 2026-03-16
refactor(ng-dev): remove rules_* repositories from ManagedRepositories

alan-agius4:
These are now considered as snapshot builds repos


josephperrott (approved):
LGTM

PR #3549, opened 2026-03-16
fix(ng-dev/release): ignore lock files when detecting snapshot changes

alan-agius4:
Update the `diff-index` command used during snapshot publishing to ignore `MODULE.bazel.lock` and `package-lock.json` files.

This prevents  triggering snapshot publishes when the only changes are in dependency lock files.


josephperrott (approved):
LGTM

PR #3553, opened 2026-03-17
build: remove redundant Bazel lockfile update workarounds

alan-agius4:
Renovate now natively supports updating Bazel lockfiles.

This change removes the `postUpgradeTasks` workaround in Renovate presets and a redundant synchronization step in the `sync-all-modules.sh` script that were previously used to manually update Bazel lockfiles.


josephperrott (approved):
LGTM

PR #3554, opened 2026-03-17
fix(renovate-presets): move PR changelog to the bottom

alan-agius4:
Moves the changelog section of the PR body template to the bottom of the body to ensure that the PR controls are not hidden or truncated when the changelog is excessively long.


josephperrott (approved):
LGTM

PR #3555, opened 2026-03-17
build: suppress pnpm output in local-dev.sh

alan-agius4:
Adds the silent flag (`-s`) to the `pnpm` command in the `local-dev.sh` script to reduce noise when building the `ng-dev` package.


josephperrott (approved):
LGTM

PR #3557, opened 2026-03-17
refactor(ng-dev/release): centralize and expand ignored paths for snapshot publishing

alan-agius4:
Centralizes the paths to exclude when detecting changes for snapshot publishing into a single constant. Additionally, adds `pubspec.lock` to the list of ignored files to avoid unnecessary snapshot releases when only lock files change.


josephperrott (approved):
LGTM

PR #3568, opened 2026-03-24
fix(ng-dev/release): improve lockfile exclusion logic in snapshot publishing

alan-agius4:
Update the snapshot publishing logic to more robustly exclude lockfiles from determining if a snapshot is affected by changes.

Specifically:
- Added pnpm-lock.yaml to the list of excluded paths.
- Added both root-level and glob variants for MODULE.bazel.lock, pnpm-lock.yaml, and pubspec.lock.


josephperrott (approved):
LGTM

PR #3571, opened 2026-03-25
ci: include bazel and bazelisk managers in renovate post-upgrade tasks

alan-agius4:
Ensure that post-upgrade tasks, such as module synchronization and generated file updates, are executed when Renovate updates Bazel dependencies or Bazelisk versions.


josephperrott (approved):
LGTM

PR #3582, opened 2026-04-01
refactor(ng-dev): Add github api retries for increased reliability

thePunderWoman:
Github has been increasingly flakey lately resulting in partial merges and all sorts of issues with failures in the tooling. This adds a retry system to ensure github API calls retry up to 3 times if they fail.


josephperrott (approved):
LGTM

I would love to see the comment about not applying retry recursively to be addressed as well

PR #3592, opened 2026-04-07
ci: allow bazelModDeps unsafe execution in renovate runner config

alan-agius4:
Renovate requires permission to run `bazel mod deps --lockfile_mode=update` when updating Bazel module dependencies. This commit adds `bazelModDeps` to `allowedUnsafeExecutions` in the Renovate runner configuration to prevent updates from failing with permission errors.


josephperrott (approved):
LGTM

PR #3597, opened 2026-04-09
fix(bazel/rules_angular): update worker to use user-configured TypeScript instead of npm version

alan-agius4:
Ensure that the users typescript version is used.

### DIFF
--- a/bazel/rules/rules_angular/src/worker/BUILD.bazel
+++ b/bazel/rules/rules_angular/src/worker/BUILD.bazel
@@ -52,7 +52,7 @@ ts_project(
         # NOTE: The Angular compiler in this target is not affecting any compilation
         # output, but it's still necessary for some foundational utils like virtual FS.
         "//:node_modules/@angular/compiler-cli",  # compiler from npm.
-        "//:node_modules/typescript",  # typescript from npm.
+        ":node_modules/typescript",  # user-configured typescript. Ensure that the users TypeScript version is always used for the compilation.
     ],
 )
 


bazel/rules/rules_angular/src/worker/BUILD.bazel
```diff
@@ -52,7 +52,7 @@ ts_project(
         # NOTE: The Angular compiler in this target is not affecting any compilation
         # output, but it's still necessary for some foundational utils like virtual FS.
         "//:node_modules/@angular/compiler-cli",  # compiler from npm.
-        "//:node_modules/typescript",  # typescript from npm.
+        ":node_modules/typescript",  # user-configured typescript. Ensure that the users TypeScript version is always used for the compilation.
```
josephperrott, line 55: I think this was previously using the version from `npm` because we can cause mismatches between the expected typescript types for the local envinronment (the users TS version) and the version that `@angular/compiler-cli` expects from npm.

Is that not a concern? Or I am misunderstanding?

alan-agius4: The problem with that approach is that any usage of `ts_project` will use the built-in version of the TypeScript instead of what is supplied by the users.

In the case of having mismatches, I think we need to come up with a better solution or use an `@angular/compiler-cli` version that supports the TypeScript versions that are used in both main and patch.

One other alternative would be not using the the `@angular/compiler-cli` FS for plain TS compilations but that is a larger task.

josephperrott: Okay, If we are okay with moving the "error point" to the angular compiler-cli version like this, thats fine with me.

josephperrott (approved):
LGTM

alan-agius4:

- main: 320dbd19c7846fafaafa47e0f1fff22a3eab00b1

PR #3598, opened 2026-04-10
fix(bazel/rules_angular): ensure correct typescript version is utilized in persistent worker runfiles

alan-agius4:
Updates the symlink_package rule to correctly propagate transitive source files, including standard library definitions, into the generated Bzlmod runfiles layout. Additionally, integrates the BazelSafeFilesystem instance within the persistent compiler worker to reliably resolve and load the workspace-provided TS toolchain at execution time.

### DIFF
--- a/bazel/rules/rules_angular/setup/extensions.bzl
+++ b/bazel/rules/rules_angular/setup/extensions.bzl
@@ -1,21 +1,37 @@
 load("//setup:repositories.bzl", "configurable_deps_repo")
 
 def _extension(ctx):
+    root_setup = None
+    generated = {}
     for mod in ctx.modules:
+        if mod.is_root and mod.tags.setup:
+            root_setup = mod.tags.setup[0]
+        elif not root_setup and mod.tags.setup:
+            root_setup = mod.tags.setup[0]
+
         for attr in mod.tags.setup:
-            configurable_deps_repo(
-                name = attr.name,
-                angular_compiler_cli = attr.angular_compiler_cli,
-                typescript = attr.typescript,
-            )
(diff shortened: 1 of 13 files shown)


bazel/rules/rules_angular/setup/extensions.bzl
```diff
@@ -1,21 +1,37 @@
...
+        elif not root_setup and mod.tags.setup:
+            root_setup = mod.tags.setup[0]
+
         for attr in mod.tags.setup:
-            configurable_deps_repo(
-                name = attr.name,
-                angular_compiler_cli = attr.angular_compiler_cli,
-                typescript = attr.typescript,
-            )
+            if attr.name not in generated:
```
josephperrott: I understand we have control here so we can manage it, but does this mean that we we will just silently "fail" if someone tries to use the same name twice?

If dev-infra creates `"dependency_rules_angular_configurable_deps"` and then in framework we also create `"rules_angular_configurable_deps"`, the dev-infra one would be encountered first and the framework one would get skipped? Would it make more sense to ensure that the same name is not generated twice?

josephperrott (approved):
LGTM

Just one thing I want to make sure we think through, but happy with whatever decision you make on it.

bazel/rules/rules_angular/setup/extensions.bzl
alan-agius4: Yes let me add an error message.

alan-agius4:

- main: 0844373e79b13faf25dcf4de316a1e4269e4fd8f

PR #3604, opened 2026-04-14
ci: update repos in org-wide-actions.yml

alan-agius4:
Remove repos that have been archived or are not active.


josephperrott (approved):
LGTM

Though I think in practice it would be fine for the `lock_closed` action to continue since it is a noop

alan-agius4:

- main: 15e8ac39e7165c428f91668fd80548fefe93c3e2

PR #3621, opened 2026-04-22
fix(ng-dev): remove unnecessary quotes from npm deprecate arguments

alan-agius4:
With the recent changes to disable the shell in the ChildProcess wrappers, quotes are passed literally to the command. This removes the unnecessary quotes around the version to ensure the deprecate command functions correctly.

### DIFF
--- a/ng-dev/release/versioning/npm-command.ts
+++ b/ng-dev/release/versioning/npm-command.ts
@@ -40,7 +40,7 @@ export abstract class NpmCommand {
     message: string,
     registryUrl: string | undefined,
   ) {
-    const args = ['deprecate', `${packageName}@"${version}"`, `"${message}"`];
+    const args = ['deprecate', `${packageName}@${version}`, message];
 
     // If a custom registry URL has been specified, add the `--registry` flag.
     if (registryUrl !== undefined) {


ng-dev/release/versioning/npm-command.ts
```diff
@@ -40,7 +40,7 @@ export abstract class NpmCommand {
     message: string,
     registryUrl: string | undefined,
   ) {
-    const args = ['deprecate', `${packageName}@"${version}"`, `"${message}"`];
+    const args = ['deprecate', `${packageName}@${version}`, message];
```
josephperrott, line 43: Do we still want the quotes around the `message` to ensure it doesn't get split into multiple args if there are quotes in the `message` value?

josephperrott (approved):
LGTM

ng-dev/release/versioning/npm-command.ts
alan-agius4: Afaik I don’t think that is needed since shell is disabled as node.js.

josephperrott:

- main: 9fc958feba3cea291aca698839e794db5a55d051

PR #3630, opened 2026-04-27
fix(ng-dev): remove quotes from bazel query in update-generated-files

alan-agius4:
The quotes around the query expression caused Bazel to treat it as a target name instead of a query, leading to an error.


josephperrott (approved):
LGTM

josephperrott:

- main: 4c9f91f47b0e5bb1134d6bfaa26cda0ba5ec629a

PR #3654, opened 2026-05-09
security: reject symlinks in PR-supplied Firebase preview artifact

wecbaiyk-blip:
## Summary

`upload-artifacts-to-firebase` extracts a PR-supplied tarball (`unsafe-artifact.zip` -> `deploy-artifact.tar.gz`) into the Firebase public directory and then deploys it to a Firebase Hosting preview channel. The artifact is explicitly annotated as `RISK` in the action's own comments, and the `description` points at the GitHub Security Lab guidance on preventing pwn requests.

This PR adds a defense-in-depth check after extraction: if `firebase-public-dir` ends up containing any symlinks, the deploy step fails fast with an explanatory error.

## Why

Currently `tar -xvzf` materialises any symlinks the PR author put in their artifact onto the runner's `firebase-public-dir`. Downstream tooling that walks `firebase-public-dir` to enumerate the files to upload (e.g. `firebase-tools` `listFiles`) calls `fs.readFile*` on each entry, which follows symlinks at the OS layer. A tar entry like

- `public/leak -> /proc/self/environ`
- `public/leak -> ~/.config/gcloud/application_default_credentials.json`
- `public/leak -> ~/.ssh/id_rsa`
- `public/leak -> /run/secrets/<anything>`

would therefore have the **target's bytes** uploaded to the publicly-readable preview CDN under a path the attacker chose.

There is no legitimate reason for a static-hosting `public/` artifact to contain symlinks (the Hosting CDN serves files, not symlinks), so refusing them outright is safer than enumerating "safe" symlink targets.

## Changes

In the `Extracting workflow artifact into Firebase public directory.` step:

- Add `set -euo pipefail` so a shell error during extraction does not silently fall through to the deploy step.
- After `tar -xvzf`, run `find "$publicDir" -type l` and exit non-zero with a `::error::` annotation if any entry is found. `-type l` catches symlink-to-file and symlink-to-directory regardless of whether their target is reachable.
- No-op when the artifact is well-formed (no symlinks).

The check intentionally fails loudly rather than silently deleting symlinks, so the contributor sees the security policy and the maintainer can audit any legitimate edge case.

## Backward compatibility

Workflows whose preview artifacts contain only regular files and directories are unaffected. Workflows that intentionally include symlinks in their preview artifact would start failing - but the maintainers' own framing (`RISK` annotation, pwn-request reference) suggests symlinks were never an intended use case.

## Why not rely on the upstream `firebase-tools` fix alone

There are in-flight `firebase-tools` PRs on the same threat model:

- [firebase/firebase-tools#10477](https://github.com/firebase/firebase-tools/pull/10477) - default-flip on `fsAsync.readdirRecursive` ignoreSymlinks.
- [firebase/firebase-tools#10478](https://github.com/firebase/firebase-tools/pull/10478) - drop symlinks from Hosting deploy upload list (`listFiles`).

Those fix the same bug class at the deployment-tool layer. This PR is the action-layer belt-and-suspenders: even if those land later, get partially reverted, or are bypassed by a downstream tool that re-introduces symlink-following, the artifact never reaches the deploy stage with symlinks present in the first place.

## Testing

Manual validation locally:

```bash
mkdir -p /tmp/pub
echo "<!doctype html>" > /tmp/pub/index.html
ln -s /etc/passwd /tmp/pub/leak
# new check rejects:
find /tmp/pub -type l   # prints /tmp/pub/leak

mkdir -p /tmp/pub2
echo "<!doctype html>" > /tmp/pub2/index.html
# new check passes:
find /tmp/pub2 -type l  # prints nothing
```

CI for this action runs in the Angular dev-infra integration suite; no test changes are needed because the action is shell-based and the new step is exercised end-to-end whenever a preview deploy runs.

### DIFF
--- a/github-actions/previews/upload-artifacts-to-firebase/action.yml
+++ b/github-actions/previews/upload-artifacts-to-firebase/action.yml
@@ -65,14 +65,25 @@ runs:
     # RISK: The downloaded `unsafe-artifact` is of input category `RISK`.
     - name: Extracting workflow artifact into Firebase public directory.
       shell: bash
+      env:
+        FIREBASE_PUBLIC_DIR: ${{inputs.firebase-public-dir}}
       run: |
-        extractDir="$RUNNER_TEMP/artifact-unpack"
+        set -euo pipefail
 
-        mkdir -p '${{inputs.firebase-public-dir}}'
+        extractDir="$RUNNER_TEMP/artifact-unpack"
+        publicDir="$FIREBASE_PUBLIC_DIR"
+
+        mkdir -p "$publicDir"
         mkdir -p "$extractDir"
 
         unzip unsafe-artifact.zip -d "$extractDir"
-        tar -xvzf "$extractDir/deploy-artifact.tar.gz" -C '${{inputs.firebase-public-dir}}'
+        tar -xzf "$extractDir/deploy-artifact.tar.gz" -C "$publicDir"
+
+        symlinks=$(find "$publicDir" -type l)
+        if [ -n "$symlinks" ]; then
+          printf '%s\n' "$symlinks" | sed 's/^/  /'
+          exit 1
+        fi
 
     - name: Extracting artifact metadata
       id: artifact-info


github-actions/previews/upload-artifacts-to-firebase/action.yml
```diff
@@ -66,13 +66,34 @@ runs:
...
         extractDir="$RUNNER_TEMP/artifact-unpack"
+        publicDir='${{inputs.firebase-public-dir}}'
 
-        mkdir -p '${{inputs.firebase-public-dir}}'
+        mkdir -p "$publicDir"
         mkdir -p "$extractDir"
 
         unzip unsafe-artifact.zip -d "$extractDir"
-        tar -xvzf "$extractDir/deploy-artifact.tar.gz" -C '${{inputs.firebase-public-dir}}'
+        tar -xvzf "$extractDir/deploy-artifact.tar.gz" -C "$publicDir"
```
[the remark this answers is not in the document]
josephperrott: Agreed on this one, please make this change.

github-actions/previews/upload-artifacts-to-firebase/action.yml
```diff
@@ -66,13 +66,34 @@ runs:
     - name: Extracting workflow artifact into Firebase public directory.
       shell: bash
       run: |
+        set -euo pipefail
+
         extractDir="$RUNNER_TEMP/artifact-unpack"
+        publicDir='${{inputs.firebase-public-dir}}'
```
josephperrott: AGENT: Instead of using single-quoted shell interpolation, it is safer to pass this input via a step-level environment variable to avoid potential bash syntax or injection issues if the input contains special characters.

Example structure:
```yaml
    - name: Extracting workflow artifact into Firebase public directory.
      shell: bash
      env:
        FIREBASE_PUBLIC_DIR: ${{inputs.firebase-public-dir}}
      run: |
        set -euo pipefail

        extractDir="$RUNNER_TEMP/artifact-unpack"
        publicDir="$FIREBASE_PUBLIC_DIR"
```

github-actions/previews/upload-artifacts-to-firebase/action.yml
```diff
@@ -66,13 +66,34 @@ runs:
...
+        # symlinks. The artifact is attacker-influenced (PR-supplied build
+        # output); a symlink such as `public/leak -> /proc/self/environ` or
+        # `public/leak -> ~/.config/gcloud/application_default_credentials.json`
+        # would otherwise be followed by downstream tooling that walks
+        # `firebase-public-dir` and reads each entry, ending up on the
+        # public Firebase Hosting CDN. There is no legitimate reason for a
+        # `public` artifact destined for static hosting to contain
+        # symlinks, so refusing them outright is the safest default.
+        # See https://securitylab.github.com/research/github-actions-preventing-pwn-requests/.
+        symlinks=$(find "$publicDir" -type l 2>/dev/null || true)
```
josephperrott: AGENT: Consider removing `|| true` and `2>/dev/null` from the `find` command.

If `find` fails for any reason, we want the step to fail (which it will due to `set -e` if `|| true` is removed). Keeping `|| true` might silently bypass the security check if `find` fails to execute properly. Removing `2>/dev/null` ensures that any errors are visible in the workflow logs for debugging.

```suggestion
        symlinks=$(find "$publicDir" -type l)
```

josephperrott (commented):
AGENT: I have reviewed the PR and left some inline suggestions.

Additionally, please check the commit message:
The commit message header uses type 'security', which is not a standard type in Angular commit guidelines. Consider changing it to:
`fix(github-actions): reject symlinks in PR-supplied Firebase preview artifact`
to match the guidelines (since 'fix' is for bug/vulnerability fixes and 'github-actions' is a valid scope).

github-actions/previews/upload-artifacts-to-firebase/action.yml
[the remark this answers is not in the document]
wecbaiyk-blip: Fixed in 4cb600f — `find` is now `symlinks=$(find "$publicDir" -type l)` with no `|| true` / `2>/dev/null`. Under `set -euo pipefail` the step now fails closed if the scan cannot complete, instead of silently treating the artifact as symlink-free.

github-actions/previews/upload-artifacts-to-firebase/action.yml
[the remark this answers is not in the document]
wecbaiyk-blip: Fixed in 4cb600f — `firebase-public-dir` is now passed via a step-level `env: FIREBASE_PUBLIC_DIR` and read as `publicDir="$FIREBASE_PUBLIC_DIR"`; it is no longer interpolated into the script body, so its value can't be parsed as part of the shell script.

github-actions/previews/upload-artifacts-to-firebase/action.yml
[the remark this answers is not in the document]
wecbaiyk-blip: Fixed in 4cb600f — dropped `-v` from tar (`tar -xzf ...`). A crafted entry name can no longer be echoed to stdout during extraction and interpreted as a workflow command.

github-actions/previews/upload-artifacts-to-firebase/action.yml
```diff
@@ -66,13 +66,34 @@ runs:
...
+        if [ -n "$symlinks" ]; then
+          echo "::error title=Symlinks rejected in artifact::The deploy artifact contains symlinks, which are not permitted in a Hosting public directory for security reasons (a symlink can leak the contents of files outside the source tree onto the public Firebase Hosting CDN)."
+          echo "Symlinks found:"
+          echo "$symlinks"
```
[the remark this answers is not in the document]
wecbaiyk-blip: Fixed in 4cb600f — the symlink list is now printed with `printf '%s\n' "$symlinks" | sed 's/^/  /'`, indenting each line so a filename containing a newline cannot start a line with `::` and be interpreted as a workflow command.

github-actions/previews/upload-artifacts-to-firebase/action.yml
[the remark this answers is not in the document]
wecbaiyk-blip: Done in 4cb600f — `tar` no longer uses `-v`.

github-actions/previews/upload-artifacts-to-firebase/action.yml
wecbaiyk-blip: Done in 4cb600f — moved to a step-level `env: FIREBASE_PUBLIC_DIR: ${{inputs.firebase-public-dir}}` and read it as `publicDir="$FIREBASE_PUBLIC_DIR"`, exactly as suggested.

github-actions/previews/upload-artifacts-to-firebase/action.yml
wecbaiyk-blip: Done in 4cb600f — removed `|| true` and `2>/dev/null`; the check now fails closed under `set -e`.

wecbaiyk-blip:
Thanks for the review @josephperrott. Pushed `4cb600f` addressing every point:

- **Commit message** — retyped header as `fix(github-actions): ...` per the Angular guidelines (was `security:`, which isn't a valid type — this was failing the commit-message lint).
- **`firebase-public-dir` via env** — now passed as a step-level `env: FIREBASE_PUBLIC_DIR` and read as `publicDir="$FIREBASE_PUBLIC_DIR"`, not interpolated into the script body.
- **`find` fail-open** — removed `|| true` and `2>/dev/null`; under `set -euo pipefail` the check now fails closed.
- **`tar -v`** — removed; extraction no longer echoes attacker-controlled entry names to stdout.
- **Symlink list output** — now `printf '%s\n' "$symlinks" | sed 's/^/  /'` so a newline in a filename can't be parsed as a workflow command.

Kept it as a single squashed commit. Happy to adjust further.

github-actions/previews/upload-artifacts-to-firebase/action.yml
```diff
@@ -65,14 +65,49 @@ runs:
...
         extractDir="$RUNNER_TEMP/artifact-unpack"
+        publicDir="$FIREBASE_PUBLIC_DIR"
 
-        mkdir -p '${{inputs.firebase-public-dir}}'
+        mkdir -p "$publicDir"
         mkdir -p "$extractDir"
 
         unzip unsafe-artifact.zip -d "$extractDir"
-        tar -xvzf "$extractDir/deploy-artifact.tar.gz" -C '${{inputs.firebase-public-dir}}'
+        # No `-v`: the archive is attacker-influenced (PR-supplied) and a crafted
```
josephperrott: Can you remove all of the comments explaining the choices, this information should be included in the commit message if you would like to preseve this thought process.

github-actions/previews/upload-artifacts-to-firebase/action.yml
```diff
@@ -65,14 +65,49 @@ runs:
...
+        # `public` artifact destined for static hosting to contain
+        # symlinks, so refusing them outright is the safest default.
+        # See https://securitylab.github.com/research/github-actions-preventing-pwn-requests/.
+        #
+        # `find` is not suffixed with `|| true` / `2>/dev/null`: under `set -e`
+        # any failure to complete the scan must fail the step (fail closed)
+        # rather than silently treating the artifact as symlink-free.
+        symlinks=$(find "$publicDir" -type l)
+        if [ -n "$symlinks" ]; then
+          echo "::error title=Symlinks rejected in artifact::The deploy artifact contains symlinks, which are not permitted in a Hosting public directory for security reasons (a symlink can leak the contents of files outside the source tree onto the public Firebase Hosting CDN)."
```
josephperrott: We can drop this preamble message, and just `exit 1` with the list of symlinks found.

josephperrott (changes requested):
LGTM

github-actions/previews/upload-artifacts-to-firebase/action.yml
wecbaiyk-blip: Done in 443bcaf — removed all the explanatory inline comments from the run script. The rationale is preserved in the commit message body.

github-actions/previews/upload-artifacts-to-firebase/action.yml
wecbaiyk-blip: Done in 443bcaf — dropped the `::error::` preamble and the `Symlinks found:` label. The step now just prints the (indented) list of symlinks found and `exit 1`.

wecbaiyk-blip:
Pushed 443bcaf addressing both remaining points:

- Removed all explanatory inline comments from the run script (rationale preserved in the commit message body).
- Dropped the `::error::` preamble and the `Symlinks found:` label — the step now just prints the list of symlinks and `exit 1`.

I kept the `printf | sed 's/^/  /'` indent on that list, since it is the previously-agreed guard against a symlink filename containing a newline + a `::workflow-command::` being interpreted by the runner; happy to drop it to a bare `printf` if you'd prefer. All earlier items (env-var input, no `tar -v`, fail-closed `find`) remain in place. Ready for re-review.

josephperrott (approved):
LGTM

josephperrott:

- main: b785efd68e376786dba49010622667a434dcaaee

PR #3659, opened 2026-05-12
fix(ng-dev/release): update PATH after nvm install to affect subsequent commands

josephperrott:
Updates the PATH environment variable in the parent process after running nvm install, so that subsequent commands spawned by the script use the correct Node.js version.


alan-agius4 (approved):

josephperrott:

- main: 96bda20c27167a0e655d292f85a1daf611a9db62

PR #3660, opened 2026-05-12
fix(ng-dev/release): make package deprecation errors non-fatal

alan-agius4:
Catches and logs errors that occur during `npm deprecate` rather than allowing them to bubble up and cause the release process to fail. This serves as a temporary workaround for issues where package deprecation requests might fail unexpectedly.

### DIFF
--- a/ng-dev/release/versioning/npm-command.ts
+++ b/ng-dev/release/versioning/npm-command.ts
@@ -11,6 +11,7 @@ import semver from 'semver';
 import {ChildProcess} from '../../utils/child-process.js';
 
 import {NpmDistTag} from './npm-registry.js';
+import {Log} from '../../utils/logging.js';
 
 export abstract class NpmCommand {
   /**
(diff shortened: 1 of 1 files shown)


josephperrott (approved):
LGTM

Your call on the comment I left, mostly just want to make sure we notice that a package failed to deprecate since it seems intermittent currently.

ng-dev/release/versioning/npm-command.ts
```diff
@@ -47,7 +48,13 @@ export abstract class NpmCommand {
       args.push('--registry', registryUrl);
     }
 
-    await ChildProcess.spawn('npm', args, {mode: 'silent'});
+    try {
+      await ChildProcess.spawn('npm', args, {mode: 'silent'});
+    } catch (e) {
+      // TODO(alanagius): remove try/catch block once https://buganizer.corp.google.com/u/1/issues/512428441 is fixed.
+      Log.error(e);
+      Log.error(`  ✘   An error occurred while deprecating "${packageName}".`);
```
josephperrott, line 56: We should consider making the failure log large maybe prefixing with 

```suggestion
      Log.error(Array(80).join('#'))
      Log.error(`  ✘   An error occurred while deprecating "${packageName}".`);
      Log.error(e);
      Log.error(Array(80).join('#'))
```

josephperrott:

- main: 602e90c2a387da841df7a6b530932fe5527bbc50

PR #3706, opened 2026-05-27
feat(github-actions): add daily Gemini model update workflow and local action

josephperrott:
Automatically updates Gemini model versions based on the official deprecation schedules. Rebuilds any dependent checked-in action bundles and verifies all tests pass daily.


.github/local-actions/update-models/action.yml
```diff
@@ -0,0 +1,5 @@
+name: 'Update Gemini Models'
+description: 'Check Gemini deprecations page and auto-update references to recommended replacements'
+runs:
+  using: 'node22'
```
alan-agius4:
```suggestion
  using: 'node24'
```

.github/local-actions/update-models/BUILD.bazel
```diff
@@ -0,0 +1,28 @@
...
+
+esbuild_checked_in(
+    name = "main",
+    srcs = [
+        ":lib",
+    ],
+    entry_point = "lib/main.ts",
+    format = "esm",
+    platform = "node",
+    target = "node22",
```
alan-agius4:
```suggestion
    target = "node24",
```

.github/local-actions/update-models/lib/main.ts
```diff
@@ -0,0 +1,121 @@
...
+        const newContent = updatedContent.replace(regex, newModel);
+        if (newContent !== updatedContent) {
+          updatedContent = newContent;
+          fileChanged = true;
+          core.info(`Found reference to deprecated model "${oldModel}" in ${file}.`);
+        }
+      }
+
+      if (fileChanged) {
+        await writeFile(file, updatedContent, 'utf-8');
```
alan-agius4, line 103: Is it safe not to re-format the file? I am guessing in most of the case this should be fine.

alan-agius4 (approved):
LGTM, just a couple of NITs

.github/local-actions/update-models/lib/main.ts
```diff
@@ -0,0 +1,121 @@
+/**
+ * @license
+ * Copyright Google LLC
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as core from '@actions/core';
+import {readFile, writeFile} from 'fs/promises';
```
alan-agius4: NIT
```suggestion
import {readFile, writeFile} from 'node:fs/promises';
```

.github/local-actions/update-models/lib/main.ts
josephperrott: I think in most cases it will be fine since its just replacing the string in place.  Its possible it could make a line too long, but when the commit is created it should autorun the formatter anyway. If we see its not working as expected we can come back and correct it here.

alan-agius4: SGTM

josephperrott:

- main: a9fcd3b48e50c7198cf257dea0db07b230afac13

PR #3714, opened 2026-05-29
fix(github-actions): resolve pending mergeability checks on LTS branches

alan-agius4:
The branch-manager local GitHub Action previously relied on a mocked minimal configuration which lacked the "release" configuration block. As a result, validations checking the `TARGET_LTS` target label failed because they could not determine active LTS branches (which requires release config settings). This caused LTS PRs to throw an `InvalidTargetLabelError` and remain in a "pending" status permanently.

This change fixes the issue by dynamically importing the actual repository's `.ng-dev/config.mjs` configuration after the repository is cloned inside the local action workspace, merging it with action input overrides and re-caching it.

We also update `setCachedConfig` to support clearing/overwriting the cache by allowing `null` values.


.github/local-actions/branch-manager/lib/main.ts
```diff
@@ -89,6 +94,16 @@ async function main() {
     // Create a tmp directory to perform checks in and change working to directory to it.
     await cloneRepoIntoTmpLocation({owner, repo});
 
+    // Remove cached config and use the repo config.
+    // TODO(alanagius): this is needed because `setupConfigAndGitClient` uses a dummy config.
```
josephperrott, line 98: Have we verified this will work even if the config is defined using a typescript files?  I know we have tried to do this in the past and found it didn't work in all locations with the changes we tried to make.

josephperrott (approved):
LGTM

Just one comment I want to make sure we have checked about loading the config from the environment locally.

.github/local-actions/branch-manager/lib/main.ts
alan-agius4: I thought we migrate those, but looks like components are still using typescript configs.

We can either migrate that, or keep the dummy config if config is in typescript format.

josephperrott: I'm fine with either, I think we just need to make sure that we don't get into a failing state with the change

alan-agius4: I went with not failing of the config is not found.

josephperrott (approved):
LGTM

josephperrott:

- main: df2b39e1010db60cea7cfbc11988897ed424a7a3

PR #3721, opened 2026-06-02
feat(bazel/rules_angular): add manual_chunks attribute to angular_package_format

alan-agius4:
Introduce a new `manual_chunks` attribute to the `angular_package_format` rule. This attribute allows configuring Rollup's `manualChunks` option to dynamically split specific source files into separate chunks during packaging.

To optimize bundle compilation performance, manual chunk patterns are normalized (extensions like `.ts`/`.mts`/`.cts` are stripped) directly within the Starlark rule context helper (`_strip_ts_extension`) before writing the template-based Rollup configuration. The Rollup configuration can then perform simple suffix checks using an optimized nested loop, avoiding anonymous closure allocations and runtime call overhead.

### Context
In PR #68810 (https://github.com/angular/angular/pull/68810#issuecomment-4594452853), introducing a new dedicated entry point to solve tree-shaking and size concerns was rejected due to negative DX overhead and bundle complexity. Instead, it was suggested (https://github.com/angular/angular/pull/68810#pullrequestreview-4389933068) that we look into packaging-level alternatives, letting the bundler group specific parts of the package into separate chunks.

This change implements this packaging-level solution directly within the `angular_package_format` Starlark rule and Rollup configuration, also ensuring the generated `.d.ts` typings align with the split chunks.

### Usage Note
To package specific files into a separate chunk, define the `manual_chunks` dictionary attribute under the `angular_package_format` rule, mapping the desired chunk name to a list of source file patterns (extensions like `.ts` can be omitted or included, and will match both `.js`/`.mjs` and `.d.ts` generation):

```starlark
angular_package_format(
    name = "pkg",
    manual_chunks = {
        "other_chunk": [
            "other_file.ts",
        ],
    },
    package = "my-package",
    deps = [":my_lib"],
)
```


josephperrott (approved):
LGTM

alan-agius4:

- main: 62bbf51f0308ddcd004c9dceec137ba83fdcd70a

JeanMeche:
(was reverted by #3722)

PR #3724, opened 2026-06-03
fix(ng-dev): call correct subcommand for setting npm tags

josephperrott:
After renaming the npm tagging sub command we failed to update the location where the script is called by the main release process


alan-agius4 (approved):

josephperrott:

- main: 6a613f2a8529185fda503da34fc4ddcab6967e7d

PR #3725, opened 2026-06-03
fix(ng-dev): prevent arbitrary file write via .ng-dev.log symlink

josephperrott:
This PR fixes a security vulnerability where an attacker could create a symbolic link named `.ng-dev.log` pointing to a sensitive file, causing it to be overwritten when `ng-dev` runs. The fix ensures that if a symbolic link is detected, execution is aborted.


alan-agius4 (approved):

josephperrott:

- main: 9dfe0e4c760b8e28fa0ce3ad7c073c805caa7877

PR #3726, opened 2026-06-03
fix(ng-dev): prevent arbitrary code execution via option injection and symlinks

josephperrott:
This PR fixes an Arbitrary Code Execution (RCE) vulnerability in the `ng-dev format` command.

### Description
The formatter utility previously failed to separate command-line options from positional file paths using the standard `--` end-of-options delimiter. This allowed an attacker to execute arbitrary code by supplying a file named `--plugin=exploit.cjs`.

### Fix
- Added the `--` delimiter when spawning the formatter subprocess to correctly pass file paths as positional arguments.
- Added an explicit check using `fs.lstatSync` to prevent symlinks from being processed, throwing a hard synchronous error instead of attempting to sanitize them.

### DIFF
--- a/ng-dev/format/run-commands-parallel.ts
+++ b/ng-dev/format/run-commands-parallel.ts
@@ -94,12 +94,8 @@ export function runFormatterInParallel(allFiles: string[], action: FormatterActi
       // Get the file and formatter for the next command.
       const {file, formatter} = nextCommand;
 
-      try {
-        if (lstatSync(file).isSymbolicLink()) {
-          throw new Error(`Security violation: symlink detected for file ${file}`);
-        }
-      } catch (error) {
-        throw new Error(`Security check failed for file ${file}: ${error}`);
+      if (lstatSync(file).isSymbolicLink()) {
+        throw new Error(`Security violation: symlink detected for file ${file}`);
       }
 
       const [spawnCmd, ...spawnArgs] = [...formatter.commandFor(action).split(' '), '--', file];
--- a/ng-dev/format/run-commands-parallel.ts
+++ b/ng-dev/format/run-commands-parallel.ts
@@ -9,6 +9,7 @@
 import {Bar} from 'cli-progress';
 import multimatch from 'multimatch';
 import {cpus} from 'os';
+import {lstatSync} from 'fs';
 
 import {ChildProcess, SpawnResult} from '../utils/child-process.js';
 import {Log} from '../utils/logging.js';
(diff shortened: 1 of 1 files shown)


ng-dev/format/run-commands-parallel.ts
```diff
@@ -93,7 +94,11 @@ export function runFormatterInParallel(allFiles: string[], action: FormatterActi
       // Get the file and formatter for the next command.
       const {file, formatter} = nextCommand;
 
-      const [spawnCmd, ...spawnArgs] = [...formatter.commandFor(action).split(' '), file];
+      if (lstatSync(file).isSymbolicLink()) {
+        throw new Error(`Security violation: symlink detected for file ${file}`);
+      }
```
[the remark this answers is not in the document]
josephperrott: Good catch! I've updated the code to use `Log.error()` and cleanly exit with `process.exit(1)` to avoid unhandled promise rejections. I have just pushed the fix.

ng-dev/format/run-commands-parallel.ts
```diff
@@ -93,7 +94,17 @@ export function runFormatterInParallel(allFiles: string[], action: FormatterActi
       // Get the file and formatter for the next command.
       const {file, formatter} = nextCommand;
 
-      const [spawnCmd, ...spawnArgs] = [...formatter.commandFor(action).split(' '), file];
+      try {
+        if (lstatSync(file).isSymbolicLink()) {
+          Log.error(`Security violation: symlink detected for file ${file}`);
+          process.exit(1);
```
josephperrott: We actually don't want to have process exits here, we would prefer to throw an `Error` instead

josephperrott: Understood! I've reverted the `process.exit(1)` and replaced it with `throw new Error(...)` as requested. The commit has been pushed.

ng-dev/format/run-commands-parallel.ts
```diff
@@ -93,7 +94,15 @@ export function runFormatterInParallel(allFiles: string[], action: FormatterActi
       // Get the file and formatter for the next command.
       const {file, formatter} = nextCommand;
 
-      const [spawnCmd, ...spawnArgs] = [...formatter.commandFor(action).split(' '), file];
+      try {
```
josephperrott: Lets not include this try/catch wrapper.  If the file isn't there we can just have a regular Error thrown by the lstatsync function

josephperrott: Make sense. I've removed the `try/catch` wrapper so `lstatSync` will throw naturally if the file is missing. The commit has been pushed!

alan-agius4 (approved):

josephperrott:

- main: 53689052908ef1f582a0bef36af509d05e3a8634

PR #3728, opened 2026-06-03
fix(ng-dev): prevent arbitrary code execution via PullApprove VM sandbox escape

josephperrott:
This PR fixes a Remote Code Execution (RCE) vulnerability in the PullApprove condition evaluator.
The previous implementation evaluated python expressions by translating them to JS and running them via Node's `vm.runInNewContext`. Because the `vm` module is not a secure sandbox, an attacker could traverse the prototype chain (e.g. using `this.constructor.constructor`) to access the host's `process` object and execute arbitrary shell commands.

This fix replaces `vm.runInNewContext` with a custom AST-based parser and evaluator (using `typescript` compiler API). The evaluator only processes a safe, whitelisted set of nodes and property accesses, preventing prototype traversal and arbitrary code execution entirely.

### DIFF
--- a/ng-dev/pullapprove/BUILD.bazel
+++ b/ng-dev/pullapprove/BUILD.bazel
@@ -18,6 +18,7 @@ ts_project(
         "//ng-dev:node_modules/@types/node",
         "//ng-dev:node_modules/@types/yargs",
         "//ng-dev:node_modules/minimatch",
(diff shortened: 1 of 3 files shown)


alan-agius4 (approved):

josephperrott:

- main: 9e6be13f9bc87370fdef74396c5a0ead463eadf7

PR #3732, opened 2026-06-04
fix(ng-dev): prevent RCE via argument injection in rollup template

josephperrott:
This PR fixes a critical security vulnerability that allows Remote Code Execution via argument injection in the `TMPL_external` substitution in the `angular_package_format` bazel rule.

It safely injects the escaped JSON object directly as a JavaScript array without wrapping it in a parsed string, eliminating the injection vector.


josephperrott:
I have implemented the suggested fixes to address the review comments. `TMPL_external` is now referenced directly without being wrapped in `JSON.parse('...')`, safely dropping in the escaped JSON array. The manual character validation loop in the `.bzl` rule has also been removed since it is no longer needed. Thank you for the review!

josephperrott:
I have moved the malicious test to `bazel/rules/rules_angular/test/ng_package_malicious` so it sits properly alongside the rule it is testing rather than at the root of the repo.

josephperrott:
I have pushed a fix for the CI failures: I updated the dummy test to be a valid target so it passes the execution phase, and I ran the code formatter across all files to resolve the lint errors. The local test suite is completely green now.

alan-agius4 (approved):

alan-agius4:

- main: 37980a075d2b9ac266a2ca1055e84c604907ec44

PR #3734, opened 2026-06-04
fix(ng-dev): prevent arbitrary file write via symlink

josephperrott:
This PR addresses a vulnerability where arbitrary file writes/information leaks could occur via symlink dereferencing when copying user-controlled files into the optimization build output.

**Changes:**
- Removed the `-L` (dereference) flag from `cp -Rf` in `optimize.sh` to prevent copying symlink targets.
- Added a pre-build `genrule` in `index.bzl` that acts as a strict security gatekeeper. It validates the original workspace `srcs` and securely aborts the build if any of the underlying source files is a symlink, fulfilling strict architectural security requirements while maintaining compatibility with Bazel's internal sandbox symlinking.

### DIFF
--- a/bazel/rules/rules_angular/src/optimization/index.bzl
+++ b/bazel/rules/rules_angular/src/optimization/index.bzl
@@ -19,11 +19,30 @@ def optimize_angular_app(
         root_paths = [""],
     )
 
+    native.genrule(
+        name = "_%s_check_symlinks" % name,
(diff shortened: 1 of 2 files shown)


alan-agius4 (approved):

alan-agius4:

- main: 252cd1a5c6eed7a5715a77c407e610304e1644c3

PR #3739, opened 2026-06-05
fix(ng-dev): prevent arbitrary code execution via HTML injection

josephperrott:
This PR mitigates a Cross-Site Scripting (XSS) vulnerability where an attacker could inject arbitrary HTML/JavaScript into the release notes via a malicious commit message.

We implemented a robust HTML escaping mechanism for commit fields in `ng-dev/release/notes/context.ts` using named HTML entities to ensure they are safely rendered by the EJS templates.


alan-agius4 (approved):

alan-agius4:

- main: 7ba841c3c224954fc17eac72150483284751e074

PR #3740, opened 2026-06-05
fix(apps): prevent open redirect via backslash manipulation in Host header

josephperrott:
This PR fixes an Open Redirect vulnerability in the `dns-redirecting` Cloud Function. The fix implements strict hostname validation (`isValidDomain`) to ensure only alphanumeric characters, dots, and hyphens are allowed, securely preventing attackers from bypassing subdomain verification via backslashes.\n\nValidated that tests pass and the bypass via `evil.com\\.material.angular.io` is appropriately rejected with a 400 Bad Request.

### DIFF
--- a/apps/functions/dns-redirecting/index.ts
+++ b/apps/functions/dns-redirecting/index.ts
@@ -1,5 +1,10 @@
 import * as functions from 'firebase-functions';
 
+const isValidDomain = (domain: string): boolean => {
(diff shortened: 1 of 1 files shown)


alan-agius4 (approved):

alan-agius4:

- main: 00ec31b42db276577188de9c0ca34d830c3f83dd

PR #3745, opened 2026-06-05
fix(rules_browsers): prevent wildcard expansion in dmg conversion

josephperrott:
This PR mitigates an option injection vulnerability in `convert_dmg_to_zip.sh`.
The script was previously expanding `*` directly in `zip -r "../${OUTFILE}" *` which allowed files starting with `-` to be interpreted as options by `zip`.
This is fixed by adding `--` before the wildcard to signal the end of options.


alan-agius4 (approved):

josephperrott:

- main: a1fd9a11b5da41eac77fb4742206bc618ac9d850

PR #3774, opened 2026-06-09
feat(ng-dev/release): implement publish-ci command and tests

josephperrott:
This PR implements the publish-ci command for CI-driven publishing, including monorepo-style tagging support and full unit tests. Stacked on top of the staging CLI changes.

### DIFF
--- a/ng-dev/release/publish/index-ci.ts
+++ b/ng-dev/release/publish/index-ci.ts
@@ -322,6 +322,7 @@ export class PublishCiTool {
     } else {
       const tempDir = mkdtempSync(join(tmpdir(), 'angular-publish-ci-'));
       const tempNpmrcPath = join(tempDir, '.npmrc');
+      const originalUserconfig = process.env['NPM_CONFIG_USERCONFIG'];
 
       try {
         const wombatNpmrcContent =
@@ -332,10 +333,17 @@ export class PublishCiTool {
         writeFileSync(tempNpmrcPath, wombatNpmrcContent);
         Log.info(green(`  ✓   Created temporary .npmrc for Wombat registry.`));
 
+        // Set the environment variable to point to the temporary config
+        process.env['NPM_CONFIG_USERCONFIG'] = tempNpmrcPath;
+
         // Publish packages
         for (const pkg of builtPackages) {
           Log.info(`Publishing "${pkg.name}"...`);
-          await NpmCommand.publish(pkg.outputPath, npmDistTag, undefined, tempNpmrcPath);
+          await NpmCommand.publish(
+            pkg.outputPath,
+            npmDistTag,
+            'https://wombat-dressing-room.appspot.com/',
+          );
           Log.info(green(`  ✓   Successfully published "${pkg.name}".`));
         }
 
@@ -346,14 +354,26 @@ export class PublishCiTool {
           }
           Log.info(`Deprecating "${pkg.name}"...`);
           const {version, message} = pkg.deprecated;
-          await NpmCommand.deprecate(pkg.name, version, message, undefined, tempNpmrcPath);
+          await NpmCommand.deprecate(
+            pkg.name,
+            version,
+            message,
+            'https://wombat-dressing-room.appspot.com/',
+          );
           Log.info(green(`  ✓   Successfully deprecated "${pkg.name}@${version}".`));
         }
       } finally {
+        // Guaranteed cleanup of files and environment
         try {
           rmSync(tempDir, {recursive: true, force: true});
         } catch (e) {
           Log.warn(`Warning: Failed to clean up temporary directory ${tempDir}: ${e}`);
+        } finally {
+          if (originalUserconfig !== undefined) {
+            process.env['NPM_CONFIG_USERCONFIG'] = originalUserconfig;
+          } else {
+            delete process.env['NPM_CONFIG_USERCONFIG'];
+          }
         }
       }
     }
--- a/ng-dev/release/publish/index-ci.ts
+++ b/ng-dev/release/publish/index-ci.ts
@@ -7,9 +7,19 @@
  */
 
 import {join} from 'path';
-import {readdirSync, statSync, readFileSync, existsSync, writeFileSync, rmSync} from 'fs';
+import {
+  readdirSync,
+  readFileSync,
+  existsSync,
+  writeFileSync,
+  rmSync,
+  mkdtempSync,
+  Dirent,
+} from 'fs';
+import {tmpdir} from 'os';
 import semver from 'semver';
-import {ReleaseConfig, BuiltPackage} from '../config/index.js';
+import {ReleaseConfig, BuiltPackage, BuiltPackageWithInfo} from '../config/index.js';
+import {analyzeAndExtendBuiltPackagesWithInfo} from './built-package-info.js';
 import {GithubConfig, NgDevConfig} from '../../utils/config.js';
 import {AuthenticatedGitClient} from '../../utils/git/authenticated-git-client.js';
 import {ReleaseNotes, workspaceRelativeChangelogPath} from '../notes/release-notes.js';
@@ -70,6 +80,11 @@ export class PublishCiTool {
       throw new Error(`No built packages found under directory ${this.options.builtPackagesDir}`);
     }
 
+    const builtPackagesWithInfo = await analyzeAndExtendBuiltPackagesWithInfo(
+      builtPackages,
+      this.config.release.npmPackages,
+    );
+
     const beforeStagingSha = this.getBeforeStagingSha();
 
     const newVersion = readPackageJsonAtRef(this.git, 'HEAD').version;
@@ -97,7 +112,7 @@ export class PublishCiTool {
 
     await this.createGithubReleaseAndTags(newVersion, newSemver, releaseNotes, npmDistTag);
 
-    await this.publishPackagesToNpm(builtPackages, npmDistTag);
+    await this.publishAndDeprecatePackages(builtPackagesWithInfo, npmDistTag);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


ng-dev/release/publish/index-ci.ts
```diff
@@ -0,0 +1,334 @@
...
+    const versionAtBeforeStagingSemver = semver.parse(versionAtBeforeStaging);
+    if (!versionAtBeforeStagingSemver) {
+      throw new Error(`Failed to parse version ${versionAtBeforeStaging} as semver.`);
+    }
+
+    // Resolve previousVersionTag
+    let previousVersionTag: string;
+    if (newSemver.prerelease.length === 0 && versionAtBeforeStagingSemver.prerelease.length > 0) {
+      // Stable release compared to prerelease. We must find the previous stable version.
+      const tagsOutput = this.git.run(['tag', '-l', 'v*']).stdout.trim();
```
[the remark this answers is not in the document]
alan-agius4: @josephperrott I think this change makes sense even though we don't always use `fetch-depth: 1`

ng-dev/release/publish/index-ci.ts
```diff
@@ -0,0 +1,447 @@
...
+          Log.warn(
+            `Warning: GitHub release for ${globalTagName} already exists, skipping release creation.`,
+          );
+        } else {
+          throw e;
+        }
+      }
+    }
+
+    if (this.config.release.npmPackages.length > 1) {
```
alan-agius4, line 276: OOC, why are we creating this tag?

ng-dev/release/publish/index-ci.ts
```diff
@@ -0,0 +1,447 @@
...
+   * @param builtPackages List of built packages to be published.
+   * @param npmDistTag The NPM distribution tag (e.g. 'latest', 'next') to publish under.
+   */
+  private async publishPackagesToNpm(builtPackages: BuiltPackage[], npmDistTag: NpmDistTag) {
+    if (this.options.dryRun) {
+      for (const pkg of builtPackages) {
+        Log.info(`[Dry-Run] Would write .npmrc and publish package: ${pkg.name} to Wombat`);
+      }
+    } else {
+      const npmrcPath = join(this.projectDir, '.npmrc');
```
alan-agius4: Instead of overwriting the current `.npmrc` file and reverting the changes afterward, why not use the `--userconfig` flag or the `NPM_CONFIG_USERCONFIG` environment variable in the child process to supply the custom Wombat configuration?

ng-dev/release/publish/index-ci.ts
```diff
@@ -0,0 +1,447 @@
...
+          [
+            `registry=https://wombat-dressing-room.appspot.com/`,
+            `//wombat-dressing-room.appspot.com/:_authToken=\${WOMBOT_TOKEN}`,
+          ].join('\n') + '\n';
+        writeFileSync(npmrcPath, wombatNpmrcContent);
+        Log.info(green(`  ✓   Configured .npmrc to use Wombat registry.`));
+
+        for (const pkg of builtPackages) {
+          Log.info(`Publishing "${pkg.name}"...`);
+          await NpmCommand.publish(
```
alan-agius4: Can be done in a followup, but we also need to handle the `npm deprecate` part.

alan-agius4 (approved):
Overall, looks good just a couple of minor things.

ng-dev/release/publish/index-ci.ts
josephperrott: This is a requirement for Wombat Dressing Room in monorepo mode. It requires package-specific tags (e.g. `@angular/core@version`) to verify publications before publishing to NPM. We still create the global tag as well.

ng-dev/release/publish/index-ci.ts
josephperrott: That is a great suggestion. It would indeed be cleaner than writing/restoring `.npmrc`. I will implement this refactoring in a followup PR to keep this PR's scope as is, since it's already approved.

ng-dev/release/publish/index-ci.ts
josephperrott: Acknowledged. I will track the deprecation support and address it in a followup PR.

ng-dev/release/publish/index-ci.ts
```diff
@@ -0,0 +1,447 @@
...
+            pkg.outputPath,
+            npmDistTag,
+            'https://wombat-dressing-room.appspot.com/',
```
alan-agius4: Shouldn't this be supplied via the `--registry` flag? So, do we really need this since we are already setting the registry in the npmrc?

ng-dev/release/publish/index-ci.ts
```diff
@@ -0,0 +1,447 @@
...
+    let files: string[];
+    try {
+      files = readdirSync(currentDir);
+    } catch (e) {
+      return;
+    }
+    if (files.includes('package.json')) {
+      try {
+        const pkgJson = JSON.parse(readFileSync(join(currentDir, 'package.json'), 'utf8'));
+        if (pkgJson.name) {
```
alan-agius4, line 412: Should we filter private?

ng-dev/release/publish/index-ci.ts
```diff
@@ -0,0 +1,447 @@
...
+          packages.push({
+            name: pkgJson.name,
+            outputPath: currentDir,
+          });
+          return;
```
alan-agius4, line 421: Can you add a comment here, where we do not traverse any deeper? (ie: we don't want something like `@angular/core/something/package.json` to be processed as a package.

ng-dev/release/publish/index-ci.ts
josephperrott: Yes, since we write the registry to the `.npmrc` file, passing it to `NpmCommand.publish` is redundant. I will clean this up in a followup PR along with the `--user-config` refactoring.

ng-dev/release/publish/index-ci.ts
josephperrott: Yes, we should definitely filter out private packages to prevent accidental publishing. I will add this safety check in a followup PR.

ng-dev/release/publish/index-ci.ts
```diff
@@ -0,0 +1,447 @@
...
+ */
+function findBuiltPackages(dir: string): BuiltPackage[] {
+  if (!existsSync(dir)) {
+    throw new Error(`The built packages directory does not exist: ${dir}`);
+  }
+  const packages: BuiltPackage[] = [];
+  const walk = (currentDir: string) => {
+    let files: string[];
+    try {
+      files = readdirSync(currentDir);
```
alan-agius4: You can use `withFileTypes` to avoid doing additional IO operations to determine if the the entry is a directory in the for loop.

```suggestion
      entries = readdirSync(currentDirm, { withFileTypes: true });
```

ng-dev/release/publish/index-ci.ts
```diff
@@ -0,0 +1,447 @@
...
+    } catch (e) {
+      return;
+    }
+    if (files.includes('package.json')) {
```
alan-agius4:
```suggestion
    const hasPackageJson = entries.some(e => e.isFile() && e.name === 'package.json');
    if (hasPackageJson)
```

ng-dev/release/publish/index-ci.ts
```diff
@@ -0,0 +1,447 @@
...
+    for (const file of files) {
+      const fullPath = join(currentDir, file);
+      try {
+        if (statSync(fullPath).isDirectory()) {
+          walk(fullPath);
+        }
+      } catch (e) {
+        // Ignore broken symlinks or unreadable files
+      }
+    }
```
alan-agius4:
```suggestion
for (const entry of entries) {
    if (entry.isDirectory()) {
      walk(join(currentDir, entry.name));
    }
}
```

ng-dev/release/publish/index-ci.ts
josephperrott: Good point, I will add a comment explaining that we return early to avoid traversing into nested node_modules or sub-packages once we identify a built package.

ng-dev/release/publish/index-ci.ts
josephperrott: Yes, using `withFileTypes` is much better to avoid extra stat calls. I will include this in the cleanup followup.

ng-dev/release/publish/index-ci.ts
josephperrott: Ack, will clean this up with the `withFileTypes` changes.

ng-dev/release/publish/index-ci.ts
josephperrott: Ack, will clean this up as well.

ng-dev/release/publish/index-ci.ts
```diff
@@ -0,0 +1,493 @@
...
+        // Set the environment variable to point to the temporary config
+        process.env['NPM_CONFIG_USERCONFIG'] = tempNpmrcPath;
+
+        // Publish packages
+        for (const pkg of builtPackages) {
+          Log.info(`Publishing "${pkg.name}"...`);
+          await NpmCommand.publish(
+            pkg.outputPath,
+            npmDistTag,
+            'https://wombat-dressing-room.appspot.com/',
```
alan-agius4: Shouldn't this be removed?

ng-dev/release/publish/index-ci.ts
```diff
@@ -0,0 +1,493 @@
...
+          if (!pkg.deprecated) {
+            continue;
+          }
+          Log.info(`Deprecating "${pkg.name}"...`);
+          const {version, message} = pkg.deprecated;
+          await NpmCommand.deprecate(
+            pkg.name,
+            version,
+            message,
+            'https://wombat-dressing-room.appspot.com/',
```
alan-agius4: Shouldn't this be removed?

alan-agius4 (approved):
LGTM, just to NITs

ng-dev/release/publish/index-ci.ts
josephperrott: JOEYBOT: Done, removed in the latest commit.

ng-dev/release/publish/index-ci.ts
josephperrott: JOEYBOT: Done, removed in the latest commit.

josephperrott:

- main: b9d15e3de22a6b8b161046d1ed5cea574a2537ca

PR #3776, opened 2026-06-10
feat(ng-dev/release): create reusable GHA workflow for publishing

josephperrott:
This PR creates a reusable GitHub Actions workflow for publishing releases using the new `publish-ci` CLI command.

### DIFF
--- a/.github/workflows/reusable-release.yml
+++ b/.github/workflows/reusable-release.yml
@@ -25,13 +25,33 @@ jobs:
     outputs:
       version: ${{ steps.get-version.outputs.version }}
       is-release-commit: ${{ steps.check-commit.outputs.is-release }}
+      dev-infra-ref: ${{ steps.get-ref.outputs.ref }}
     steps:
+      # Step 1: Extract the ref of the running reusable workflow to use for checking out dev-infra actions.
+      - name: Extract dev-infra ref
+        id: get-ref
+        run: |
+          # github.workflow_ref is of the form: "owner/repo/path/to/workflow.yml@ref"
+          # Extract the ref (part after @)
+          REF=$(echo "${{ github.workflow_ref }}" | cut -d'@' -f2)
+          echo "ref=$REF" >> $GITHUB_OUTPUT
+
+      # Step 2: Checkout dev-infra dynamically at the same ref to get the custom actions
+      - name: Checkout dev-infra
+        uses: actions/checkout@v4
+        with:
+          repository: angular/dev-infra
+          ref: ${{ steps.get-ref.outputs.ref }}
+          path: .dev-infra
+
+      # Step 3: Initialize environment (using local checkout-and-setup-node)
+      # We pass clean: false so it doesn't delete the .dev-infra directory we just checked out!
       - name: Initialize environment
-        uses: angular/dev-infra/github-actions/npm/checkout-and-setup-node@main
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-release.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Initialize environment
+        uses: angular/dev-infra/github-actions/npm/checkout-and-setup-node@main
+
+      - name: Install node modules
+        run: pnpm install --frozen-lockfile
+
+      - name: Check if commit is a release commit
+        id: check-commit
+        run: |
+          COMMIT_MSG=$(git log -1 --pretty=%B)
```
alan-agius4, line 58: Any reason why not use the commit message from github action? Maybe as an input since this is a shared action, and the consumer would pass in `github.event.head_commit.message`?

.github/workflows/reusable-release.yml
```diff
@@ -0,0 +1,97 @@
...
+    name: Build & Verify
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get-version.outputs.version }}
+      is-release-commit: ${{ steps.check-commit.outputs.is-release }}
+    steps:
+      - name: Initialize environment
+        uses: angular/dev-infra/github-actions/npm/checkout-and-setup-node@main
+
+      - name: Install node modules
```
alan-agius4: Shouldn't this be done when    ` if: steps.check-commit.outputs.is-release == 'true'`?

.github/workflows/reusable-release.yml
```diff
@@ -0,0 +1,97 @@
...
+  # Job 1: Build and Verify the release assets.
+  build:
+    name: Build & Verify
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get-version.outputs.version }}
+      is-release-commit: ${{ steps.check-commit.outputs.is-release }}
+    steps:
+      - name: Initialize environment
+        uses: angular/dev-infra/github-actions/npm/checkout-and-setup-node@main
```
alan-agius4: NIT: instead of `@main` likely this should be a SHA. Also in the case we check the commit message using the input as suggested below, we could also skip this skep if release is false.

alan-agius4:
Note: these are just drive by comments.

.github/workflows/reusable-release.yml
josephperrott: I think it actually makes sense to actually verify the commit being used is the release commit.  We currently control both sides of the process, but passing that in to the workflow call and allowing it to be provided instead of derived from the source of truth (the actual git tree) is "less secure" I believe

.github/workflows/reusable-release.yml
josephperrott: JOEYBOT: Done, pinned the internal actions to a stable SHA (442c2fcbf06a321b5196b4c5fc70e78a49242958) matching the pinning used in other workflows.

.github/workflows/reusable-release.yml
josephperrott: JOEYBOT: Done. Reordered the steps so that we check if it is a release commit immediately after initialization, and then conditionally run the pnpm install step only if it is a release commit.

alan-agius4 (approved):

josephperrott:

- main: 11ee1f5400aa3220ff42546c8e1fe75ea0041e6b

PR #3788, opened 2026-06-16
fix(ng-dev/pr): reset uncommitted changes after merge attempt

alan-agius4:
Ensure that the repository is cleaned up by resetting any uncommitted changes in the finally block of the merge tool.

This can happen because pnpm may install the dependencies of the target branch to run pre-commit hooks, which can cause the lockfile to be updated and leave the repository in a dirty state. Resetting the changes ensures a clean state before switching back to the previous branch.


josephperrott (approved):
LGTM

alan-agius4:

- main: e9faacd5b4df391f59989b6fb448b2c24115d592

PR #3850, opened 2026-07-17
fix: remove bundled bazel remote credential

### LINKED ISSUE #3849: Stop shipping a bundled GCP credential in configure-remote
The `github-actions/bazel/configure-remote` action currently falls back to a checked-in Google credential when `google_credential` is omitted. Because the repository also ships the material needed to reverse that bundle, the credential can be recovered offline by anyone who clones the repo.

A fix is ready that removes the bundled credential entirely and makes the action fail closed:
- remove the committed `gcp_token.data` artifact and its decryption helpers
- configure Bazel remote execution only when `google_credential` is supplied explicitly
- update trusted workflows to pass `secrets.RBE_TRUSTED_BUILDS_USER`
- fall back to local execution for workflows that do not receive a credential, while keeping flaky test retries

Maintainers should also revoke or rotate the previously committed service-account key and review its recent usage, since removing the files from the repository does not revoke an already exposed key.

M0nd0R:
Closes #3849.

## What changed
- remove the bundled credential artifact and decryption helpers from `github-actions/bazel/configure-remote`
- only configure Bazel remote execution when `google_credential` is supplied explicitly
- update trusted workflows to pass `secrets.RBE_TRUSTED_BUILDS_USER`
- add a regression test that covers both the no-credential and explicit-credential paths

## Why
The action currently falls back to a checked-in GCP credential. Because the repository also includes the material needed to reverse that bundle, the action is effectively shipping a recoverable service-account key.

## Impact
- trusted workflows keep using RBE by passing the existing secret explicitly
- public or forked PR workflows without secrets fall back to local execution instead of using a bundled credential

## Validation
- `node --test github-actions/bazel/configure-remote/configure-remote.test.mjs`
- `bash proofs/angular-rbe-public-service-account/run-local-proof.sh` now fails because `gcp_token.data` is no longer present


josephperrott:
Closing as this is not something we plan to do.

PR #3851, opened 2026-07-17
fix: correct issue labeling action path

M0nd0R:
## What changed
- update `.github/workflows/dev-infra.yml` to use the checked-in issue labeling action path

## Why
The `issues` job in the `DevInfra` workflow points at `./github-actions/issue-labeling`, but the local action actually lives at `./github-actions/labeling/issue`. That mismatch causes issue-triggered runs to fail before the labeling action starts.

## Validation
- confirmed `github-actions/labeling/issue/action.yml` exists
- inspected the failing run log for workflow run `29552379887`, which reported that `./github-actions/issue-labeling` had no `action.yml`


M0nd0R:
Closing this stacked PR and replacing it with a clean one-commit fix based on current main.

PR #3856, opened 2026-07-20
fix(bazel): verify TypeScript version from MODULE.bazel instead of lockfile

alan-agius4:
In recent versions of `aspect_rules_ts`, the module extension is implemented in a way that doesn't record the version metadata in `MODULE.bazel.lock` under the expected path or format, causing `validate_ts_version_match` to fail to extract the version, resolving it to `unknown` and resulting in validation failures.

To resolve this issue, the `verify-typescript.mjs` validation script is refactored to read and extract `ts_version` directly from `MODULE.bazel` instead. The macro definition in `defs.bzl` and the invocation in `BUILD.bazel` are updated to pass `MODULE.bazel` instead of `MODULE.bazel.lock`.


josephperrott (approved):
LGTM

alan-agius4:

- main: 9d539d3fc5e3ab718830b2ab8ffa4d97d2f0afb2

PR #3905, opened 2026-08-06
fix(ng-dev): prevent current working directory executable injection on Windows

### LINKED ISSUE #3897: `Prompt.editor` spawns a bare `notepad` with no `cwd`, so on Windows it runs an executable from the repository being reviewed
## What happens

`utils/prompt.ts` aliases `Prompt.editor` to `@inquirer/prompts`' `editor`, which reaches
`@inquirer/external-editor`:

```js
this.editor = parseEditorCommand(process.env["VISUAL"] ?? process.env["EDITOR"]
              ?? (process.platform.startsWith("win") ? "notepad" : "vim"));
...
spawnSync(this.editor.bin, this.editorArgs(), { shell: false, stdio: "inherit" })
```

Three properties combine here:

1. the command is a **bare name** — `notepad` — when `VISUAL` and `EDITOR` are unset, which is the
   default on Windows;
2. `shell: false`, so this goes through `CreateProcess`, which searches the **current working
   directory before `PATH`**;
3. **no `cwd` option is passed**, so the child is looked for in the process working directory.

On Windows that means a file named `notepad.exe` sitting in the working directory is executed
instead of the real editor.

## Why the working directory is a contributor's tree

`pr/common/checkout-pr.ts` ends a review checkout with:

```ts
git.run(['fetch', '-q', headRefUrl, '--', headRefName]);
git.run(['checkout', '--detach', 'FETCH_HEAD']);
```

and `pr/checkout/checkout.ts` does not call `resetGitState()` on the normal path — leaving the
contributor's tree checked out is the whole point of the command. So after
`ng-dev pr checkout <n>` the maintainer is standing inside a tree whose contents came from the pull
request.

The commit-message editor is reached from `pr/merge/strategies/api-merge.ts:105`:

```ts
} else if (hasOnlySquashForOneCommit) {
  mergeOptions.merge_method = 'squash';
  await this._promptCommitMessageEdit(pullRequest, mergeOptions);
}
```

`hasOnlySquashForOneCommit` is `normalCommitsCount === 1 && squashCount > 1` — one ordinary commit
plus two or more `squash!` commits. A contributor writes their own commit messages, so whether that
prompt opens is decided by the pull request, not by the maintainer.

## What was executed

Against `@angular/ng-dev` installed from the revision `angular/angular` pins, driven through its own
bundled `@inquirer/external-editor`, on Windows:

| arm | condition | result |
|---|---|---|
| 1 | working tree contains `notepad.exe`, `NoDefaultCurrentDirectoryInExePath` at its default | the planted binary ran — `argv=[…\Temp\inquirer-editor-LMPwx4\24bcd89b-…]` |
| 2 | same tree, `NoDefaultCurrentDirectoryInExePath=1` | the real notepad opened |
| 3 | clean directory, mitigation unset | the real notepad opened |

Arms 2 and 3 are the controls: the instrument returns the other answer when the mitigation is on and
when the plant is absent.

Separately, `ng-dev pr checkout <n>` was run against a private repository and a pull request I
created, and left the contributor's file in place:

```
tree contains : git.exe, notepad.exe, package.json, README.md
```

The spawn passes no `env`, so the child receives `process.env` whole. ng-dev's own help text says
where the token lives:

```
--github-token  Github token. If not set, token is retrieved from the environment variables.
```

and a probe reporting only presence and length confirmed it arrives:

```
PLANT notepad.exe argv=[…\Temp\inquirer-editor-K4a8yz\97d707d6-…]
  inherited environment:
    GITHUB_TOKEN = <PRESENT len=40>
```

## What was NOT executed, stated plainly

**The join.** I have not run `ng-dev pr merge <n>` end to end into the editor prompt from a checked
out contributor tree. `pr merge` validates release trains before it reaches the merge strategy, and
reproducing that on a lab repository needs a full Angular-shaped release setup (published versions,
version branches, release config) that I did not build. Each end of the chain is executed; the join
is established from source only.

I would rather say that than imply a complete chain.

## Suggested fix

Resolve the editor once and pass an absolute path, and pass an explicit `cwd` that is not the
repository under review. The same shape as PR #3753, which validated the resolved node path in
`invokeNvmInstall`.

Pinning `VISUAL`/`EDITOR` to an absolute path before calling `Prompt.editor` would also close it, and
is a smaller change.

## Disclosure

Reported to Google's OSS VRP as issue 542356914. They assessed it as below the threshold they use for
escalating to product teams and explicitly invited public disclosure, which is why this is filed here.
No Angular repository was used in any of the testing above — all of it ran against a private
repository and a pull request I created myself.

alan-agius4:
Sets `NoDefaultCurrentDirectoryInExePath=1` during CLI initialization on Windows to ensure `CreateProcess` does not include the current working directory in the search path when resolving bare executable names.

Fixes #3897

### DIFF
--- a/ng-dev/cli.ts
+++ b/ng-dev/cli.ts
@@ -27,6 +27,12 @@ import {Argv} from 'yargs';
 
 runParserWithCompletedFunctions((yargs: Argv) => {
   process.exitCode = 0;
+
+  // Ensure Windows CreateProcess does not search the current directory for bare executable names.
+  if (process.platform === 'win32') {
+    process.env['NoDefaultCurrentDirectoryInExePath'] = '1';
+  }
+
   return yargs
     .scriptName('ng-dev')
     .middleware([captureLogOutputForCommand, ngDevVersionMiddleware], true)


clydin (approved):

alan-agius4:

- main: 92c6b596e59e320c9edbc6d6490c1047264c4a0d

PR #3911, opened 2026-08-08
fix(ng-dev): verify validated head SHA before merging a pull request

herdiyana256:
`ng-dev pr merge` validates a pull request (approvals, CI status, target labels) against the head commit the GitHub API reports when the PR is loaded, and stores it as `pullRequest.headSha`. That value was never checked again, so both merge strategies operate on the live pull request head instead of the commit that was validated:

- `MergeStrategy.prepare` fetches the mutable `pull/<number>/head` ref into `merge_pr_head`, and the autosquash strategy rebases and cherry-picks from there.
- `GithubApiMergeStrategy` calls `pulls.merge` without a `sha`, so GitHub merges whatever the head currently is.

Between the validation read and the fetch (or the API call) the pull request author can push a new commit. The merge then lands that commit even though it was never reviewed or run through CI, a time-of-check/time-of-use gap.

`prepare` now resolves the fetched head and fails closed unless it equals the validated `headSha`, and the API merge strategy passes `headSha` to `pulls.merge` so the server rejects the merge with a 409 if the head has moved since validation. A unit test covers both the mismatch (rejected) and match (allowed) cases.


herdiyana256:
Gentle bump on this one. It has been open a week with no reviewer assigned.

This is a security fix in the same `ng-dev pr merge` area as the recent hardening work: the validated head SHA is captured but never enforced before the merge, so both merge strategies can operate on a PR head that changed after validation (CWE-367 TOCTOU), landing unreviewed code.

@josephperrott would you be the right person to look, given the recent `ng-dev` merge/security fixes? Happy to rebase or split if that helps.

herdiyana256:
Pushed 7c7c8dde addressing the feedback on the 409. Added a branch next to the existing 403/404 handling in `GithubApiMergeStrategy.merge` that catches the 409 and throws a `FatalMergeToolError` with a clear message (the head moved after validation, re-run the merge), instead of letting the raw Octokit error propagate.

`tsc --noEmit` and `prettier --check` are both clean on the touched file.

alan-agius4 (approved):
Thanks

alan-agius4:

- main: 6d918a6dd670e28afd4ba54d00b805d4aba43dca

PR #3951, opened 2026-09-02
security: remove fake credential encryption

Tednoob17:
Fixes recoverable production credentials (GCP, BrowserStack, SauceLabs) by removing fake AES-256-GCM encryption whose key was derivable from the public org name.

- Delete committed `.data` credential files and `encrypt.ts`
- Replace bundled JS with minimal ESM scripts that read credentials from env
- Update `action.yml` inputs to accept secrets from GitHub Actions
- Mark required inputs and pass them as env vars in composite action steps
- Restore `BUILD.bazel` targets without encrypted blob references
- Add test verifying no hardcoded keys or encrypted blobs remain


alan-agius4:
Forked repos cannot access secrets, which disables RBE for pull requests and breaks existing workflows.

PR #3959, opened 2026-09-04
fix(bazel/rules_angular): emit rollup config through a single template substitution

### LINKED ISSUE #3958: angular/dev-infra — `angular_package_format`: rollup config template injection (bypass of the #3732 fix)
**Project:** `angular/dev-infra` (`bazel/rules/rules_angular`), also mirrored to `angular/rules_angular` and consumed by `angular/angular` via `git_override`
**Commit tested:** `6a54b1f6` (`main`)
**Affected files:** `bazel/rules/rules_angular/src/ng_package/angular_package_format.bzl`, `bazel/rules/rules_angular/src/ng_package/rollup/rollup.config.js`
**Type:** Code injection into a generated build script, executed during the Bazel build
**Prior remediation:** PR #3732 / commits `9eff3b73`, `7fd1ad87` — titled "prevent RCE via argument injection in rollup template". That fix is ineffective; this report bypasses it.

---

## Description

`_write_rollup_config` passes nine substitutions to `ctx.actions.expand_template`. Bazel applies them **sequentially over the accumulating template** — `LocalTemplateExpansionStrategy.getExpandedTemplateUnsafe` loops `replaceAllLiteral` and reassigns the result — so text inserted by one substitution is re-scanned by every later one.

`json.encode` escapes quotes inside a value, but has no reason to escape the literal text `TMPL_side_effect_entrypoints`; it is plain `[A-Za-z_]`. An `externals` entry containing that text therefore survives encoding into the emitted string literal, and the next substitution splices a raw JSON array — with unescaped `"` — into the middle of it, terminating the literal and injecting JavaScript into the generated config.

Rollup executes that config when loading `--config`, so the injected code runs as part of the build.

This defeats the premise of the #3732 fix. The reviewing bot on that PR asserted that `json.encode` "already produces a perfectly valid and safe JavaScript array literal … completely safe from injection". That holds only for a single-pass substitution; `expand_template` is not single-pass.

Two further sinks introduced by the same code were never converted and remain directly injectable:

```js
const entrypointMetadata = JSON.parse(`TMPL_metadata`);              // template literal: ${…} evaluates
const sideEffectEntryPoints = JSON.parse('TMPL_side_effect_entrypoints'); // single-quoted: ' breaks out
```

## Reproduction

Append to `bazel/rules/rules_angular/test/ng_package_malicious/BUILD.bazel`:

```starlark
angular_package_format(
    name = "bypass_pkg",
    package = "+require(String.fromCharCode(99,104,105,108,100,95,112,114,111,99,101,115,115)).execSync(String.fromCharCode(116,111,117,99,104,32,47,116,109,112,47,114,117,108,101,115,95,97,110,103,117,108,97,114,95,114,99,101,95,112,111,99))+",
    side_effect_entry_points = ["+require(String.fromCharCode(99,104,105,108,100,95,112,114,111,99,101,115,115)).execSync(String.fromCharCode(116,111,117,99,104,32,47,116,109,112,47,114,117,108,101,115,95,97,110,103,117,108,97,114,95,114,99,101,95,112,111,99))+"],
    externals = ["aTMPL_side_effect_entrypointsb"],
    deps = [":dummy"],
)
```

The payload is quote- and backtick-free (`String.fromCharCode`) so it stays valid in all three templated contexts. It runs `touch /tmp/rules_angular_rce_poc`.

```
cd bazel/rules/rules_angular
rm -f /tmp/rules_angular_rce_poc
pnpm bazel build //test/ng_package_malicious:bypass_pkg
ls -la /tmp/rules_angular_rce_poc
```

## Actual result

The build **succeeds**:

```
INFO: Found 1 target...
INFO: Elapsed time: 56.915s, Critical Path: 6.67s
INFO: 589 processes: 363 internal, 225 darwin-sandbox, 1 worker.
INFO: Build completed successfully, 589 total actions
```

and the payload has run:

```
-rw-r--r--@ 1 user  wheel  0 /tmp/rules_angular_rce_poc
```

Line 70 of `bazel-out/.../test/ng_package_malicious/_bypass_pkg_fesm.rollup.conf.js`, with the `WELL_KNOWN_EXTERNALS` entries elided:

```js
const external = ["@angular/animations", ..., "tslib","a["+require(String.fromCharCode(99,104,105,108,100,95,112,114,111,99,101,115,115)).execSync(String.fromCharCode(116,111,117,99,104,32,47,116,109,112,47,114,117,108,101,115,95,97,110,103,117,108,97,114,95,114,99,101,95,112,111,99))+"]b"];
```

There is no error and no warning. Execution is silent and the build result is green.

## Expected result

The `externals` value stays inert: `"aTMPL_side_effect_entrypointsb"`.

## Why the existing regression target does not catch this

`test/ng_package_malicious` (added by #3732 as the regression target for this exact class) is a build-only smoke target with no assertion on the generated config, so it passes whether or not injection occurs.

Separately, a payload containing quotes produces a syntax error rather than execution — the naive form breaks the build loudly. Only a quote-free payload is silent, which is why this was not noticed incidentally.

## Root cause, isolated

A two-entry template with no Angular dependencies reproduces the primitive on Bazel 8.7.0:

```starlark
substitutions = {"KEY_A": "A_val<KEY_C>", "KEY_C": "C_val<KEY_A>"}
```

```
A=A_val<C_val<KEY_A>>
B=B_val
C=C_val<KEY_A>
```

The forward reference is spliced; the backward one is not. Starlark dict insertion order is the substitution order. Any Bazel rule that passes multiple substitutions to `expand_template` with values derived from rule inputs has this property — `json.encode` provides no protection against it.

## Suggested fix

Collapse to a single substitution carrying one JSON object, destructured in the template. With one substitution there is no later pass, so no inserted value can be re-entered as a placeholder:

```starlark
substitutions = {
    "TMPL_config": json.encode(struct(
        bannerFile = ctx.file.license_banner.path if ctx.file.license_banner else None,
        dtsMode = dts_mode,
        entrypointMetadata = metadata_arg,
        external = externals,
        moduleMappings = mappings,
        nodeModulesRoot = "node_modules",
        rootDir = root_dir,
        sideEffectEntryPoints = side_effect_entry_points,
        workspaceName = ctx.workspace_name,
    )),
},
```

```js
const {
  workspaceName, rootDir, bannerFile, moduleMappings, nodeModulesRoot,
  entrypointMetadata, sideEffectEntryPoints, external, dtsMode,
} = TMPL_config;
```

Alternatively, keep rule inputs out of the template entirely by writing them with `ctx.actions.write` to a sidecar JSON file read at runtime — a pattern already used in six other places in this repository.

Patch submitted as a pull request against this issue.

## Verified fix

Applying the change above to `angular/dev-infra@6a54b1f6` blocks the payload: the same target builds green and the injected value becomes an inert literal, `"aTMPL_side_effect_entrypointsb"`. The `rules_angular` test suite (`bazel test //...`, 9 test targets) still passes with it applied.

## Status

- `rollup.config.js` and `angular_package_format.bzl` on `angular/dev-infra@main` are still the pre-fix form; last touch is `8befcbf7` (2026-07-20), an unrelated dependency bump.
- The `angular/rules_angular` mirror is identical and equally unfixed.
- No matching open or closed issue, no PR, and no published security advisory.
- The review on #3732 (one automated reviewer plus one approval) does not mention the other substitutions.

## Environment

`angular/dev-infra@6a54b1f6`, Bazel 8.7.0 (pinned by `.bazelversion`), Node v24.20.0 toolchain / Node v26.7.0 host, macOS arm64. `pnpm` is not installed on the test machine, so Bazel was invoked as `npx @bazel/bazelisk@1.28.1 build …`; `pnpm bazel` is the repository's wrapper around the same binary, and the reproduction steps use that form because it is the documented workflow.

VenkatKwest:
`ctx.actions.expand_template` applies substitutions sequentially over the accumulating template, so text inserted by one substitution is re-scanned by every later one. An `externals` entry containing the literal placeholder name `TMPL_side_effect_entrypoints` therefore has a raw JSON array spliced into its string literal, injecting JavaScript into the generated rollup config. Full analysis and reproduction in #3958.

This collapses the nine substitutions into a single `TMPL_config` object, destructured in the template. With one substitution there is no later pass, so no inserted value can be re-entered as a placeholder.

Two incidental changes: `moduleMappings` moves from `str()` to `json.encode`, the correct encoding for a JavaScript literal; and `bannerFile` is now `null` rather than `undefined` when no banner is set, whose only use is the `if (bannerFile)` guard.

Verified: the reproduction in #3958 no longer executes, and `bazel test //...` in `bazel/rules/rules_angular` passes 9/9.

Fixes #3958.

### DIFF
--- a/bazel/rules/rules_angular/src/ng_package/angular_package_format.bzl
+++ b/bazel/rules/rules_angular/src/ng_package/angular_package_format.bzl
@@ -77,17 +77,22 @@ def _write_rollup_config(
     ctx.actions.expand_template(
         output = config,
         template = ctx.file._rollup_config_tmpl,
+        # Substitutions are applied sequentially over the accumulating template, so text inserted
+        # by one substitution is re-scanned by every later one. A single substitution carrying all
+        # values as one JSON object keeps rule inputs from being re-entered as placeholders.
         substitutions = {
-            "TMPL_banner_file": "\"%s\"" % ctx.file.license_banner.path if ctx.file.license_banner else "undefined",
-            "TMPL_module_mappings": str(mappings),
-            # TODO: Determine node_modules_root
-            "TMPL_node_modules_root": "node_modules",
-            "TMPL_metadata": json.encode(metadata_arg),
-            "TMPL_root_dir": root_dir,
-            "TMPL_workspace_name": ctx.workspace_name,
-            "TMPL_external": json.encode(externals),
-            "TMPL_side_effect_entrypoints": json.encode(side_effect_entry_points),
-            "TMPL_dts_mode": "true" if dts_mode else "false",
+            "TMPL_config": json.encode(struct(
+                bannerFile = ctx.file.license_banner.path if ctx.file.license_banner else None,
+                dtsMode = dts_mode,
+                entrypointMetadata = metadata_arg,
+                external = externals,
+                moduleMappings = mappings,
+                # TODO: Determine node_modules_root
+                nodeModulesRoot = "node_modules",
+                rootDir = root_dir,
+                sideEffectEntryPoints = side_effect_entry_points,
+                workspaceName = ctx.workspace_name,
+            )),
         },
     )
 
--- a/bazel/rules/rules_angular/src/ng_package/rollup/rollup.config.js
+++ b/bazel/rules/rules_angular/src/ng_package/rollup/rollup.config.js
@@ -60,15 +60,17 @@ function log_verbose(...m) {
   if (!!process.env['VERBOSE_LOGS']) console.error(`[${path.basename(__filename)}]`, ...m);
 }
 
-const workspaceName = 'TMPL_workspace_name';
-const rootDir = 'TMPL_root_dir';
-const bannerFile = TMPL_banner_file;
-const moduleMappings = TMPL_module_mappings;
-const nodeModulesRoot = 'TMPL_node_modules_root';
-const entrypointMetadata = JSON.parse(`TMPL_metadata`);
-const sideEffectEntryPoints = JSON.parse('TMPL_side_effect_entrypoints');
-const external = TMPL_external;
-const dtsMode = TMPL_dts_mode;
+const {
+  workspaceName,
+  rootDir,
+  bannerFile,
+  moduleMappings,
+  nodeModulesRoot,
+  entrypointMetadata,
+  sideEffectEntryPoints,
+  external,
+  dtsMode,
+} = TMPL_config;
 
 log_verbose(`running with
   cwd: ${process.cwd()}


alan-agius4 (approved):
LGTM, thanks

alan-agius4:

- main: 5fe4e669288e4d83880c5e880bf9b307f03f7471

VenkatKwest:
Ownership verification for Google IssueTracker report #555552987 ([Report #555552987](https://issuetracker.google.com/issues/555552987)).

PR #3984, opened 2026-09-17
fix(ng-dev): use pnpm list for ng-dev version check

alan-agius4:
Previously, `verifyNgDevToolIsUpToDate` attempted to fetch `pnpm-lock.yaml` from GitHub upstream via `git.github.repos.getContent`. This resulted in frequent failures and flakes (which we have seen a lot of in `angular-cli`), because:

1. **GitHub API 1MB Limit**: `repos.getContent` rejects files larger than 1 MB, and lockfiles in large repositories like `angular-cli` or `angular` easily exceed this limit (~3.5 MB+).
2. **Rate Limiting & Auth in Sandboxes**: In non-interactive contexts (such as Bazel workspace status actions or CI), missing GitHub credentials or rate limits cause the fetch to fail.
3. **Interactive Prompt Crash**: When the check failed to extract the expected version from the lockfile, it invoked `Prompt.confirm`. In non-interactive environments where stdin is closed or not a TTY, `@inquirer/prompts` throws an unhandled error (`ExitPromptError`), causing the Node process to exit with status 1 (failing `BazelWorkspaceStatusAction stable-status.txt`).

This PR updates the check to use:
```sh
pnpm list "@angular/ng-dev" --json --lockfile-only
```
directly in the workspace directory. This:
- Executes locally with no network or GitHub API requests.
- Uses pnpm's native lockfile parser.
- Eliminates the `@pnpm/dependency-path` dependency and custom YAML parsing.
- Guards prompts against non-interactive environments.


clydin (approved):

alan-agius4:

- main: 71e60c68d9d483b0d763b8e6ed4dd9deb8da3b1a

josephperrott (commented):
LGTM

But I will note that this does perform a different action.  This is now verifying that the version matches the expected value in the current repository state, rather than checking against the version at head upstream.

